PR #166, opened 2020-03-25
Payment API database model change

Wiezzel:
⚠️ **Please review, but do not merge!** ⚠️ 

* Introduced `pay_agreement` and `pay_activity` tables to store pre-computed amounts (due, accepted, paid).
* It is assumed that all payments within a single agreement will be sent from a single `payer_addr` to a single `payee_addr` (for now `payer_addr` = `requestor_id` and `payee_addr` = `provider_id`).
* It is assumed that all payments within a single agreement will be sent using the same `payment_platform` (for now this field is ignored).
* Debit note's `total_amount_due` represents amount due for **a single activity**. It's **not** the total amount due for an agreement.
* Invoice is the final payment document issued for an agreement. There can be only one invoice per agreement (unless cancelled or rejected). Invoice's `amount` is the total amount for an agreement.
* `last_debit_note_id` was removed from invoice because invoice is related to an agreement, while debit notes are related to particular activities. There could be multiple 'last' debit notes.


mfranciszkiewicz:
@Wiezzel Will do. But I believe that could also be addressed in a separate PR

.env-template
```diff
@@ -23,10 +22,6 @@ REQUIRED_CONFIRMATIONS=5
 # Descriptor file (JSON) for available ExeUnits.
 #EXE_UNIT_PATH=exeunits-descriptor.json
 
-# Credit address. Can be set same as default identity.
-# It will be removed in future release -- agreement will specify it.
-CREDIT_ADDRESS=0xa74476443119A942dE498590Fe1f2454d7D4aC0d
```
tworec, line 28: :+1:

PR #223, opened 2020-04-16
New GNT Driver

bidzyyys:
Refers: #151


core/payment-driver/src/gnt.rs
```diff
@@ -1,699 +1,560 @@
...
-        // Do not have any info about date in receipt
-        let date = None;
-
-        PaymentDetails {
-            recipient,
-            sender,
-            amount,
-            date,
+        invoice_id: &str,
+        _sign_tx: SignTx<'a>,
```
bidzyyys, line 386: Remove parameter if unused

PR #251, opened 2020-04-29
Constraints DSL

filipgolem:
Constraints creation (more examples: `cargo run --example constraints`):
```rust
let c = constraints![
    "golem.inf.mem.gib.macro" == 1,
    "only.key",
    "c.macro" < 3,
    constraints!["test.1" > 1, "test.2" != 2],
    "d.macro" == 20,
    "e.macro" < 10
]
.without("c.macro")
.or(constraints!["golem.macro.one" == 1, "golem.macro.two" < 2]);
println!("Created with macro:\n{}", c);
```
Output:
```c
Created with macro:
(|
  (&
    (golem.inf.mem.gib.macro=1)
    (only.key),
    (test.1>1)
    (test.2<>2)
    (d.macro=20)
    (e.macro<10)
  )
  (&
    (golem.macro.one=1)
    (golem.macro.two<2)
  )
)
```


Cargo.toml
```diff
@@ -29,6 +29,7 @@ members = [
     "utils/process",
     "utils/transfer",
     "utils/path",
+    "utils/constraints",
```
nieznanysprawiciel, line 32: Note that there's utils/agreement-utils crate. It would be nice to have all utilities for offers/demand building in one place. But it's not a must.

utils/constraints/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub fn new<T: Into<serde_json::Value>>(v: T) -> Self {
+        ConstraintKey(v.into())
+    }
+}
+
+impl From<&str> for ConstraintKey {
+    fn from(key: &str) -> Self {
+        ConstraintKey::new(serde_json::Value::String(key.to_string()))
+    }
+}
```
nieznanysprawiciel, line 152: Probably you would like to use AsRef https://doc.rust-lang.org/std/convert/trait.AsRef.html to catch everything what is stringable

utils/constraints/examples/constraints.rs
```diff
@@ -0,0 +1,58 @@
+use ya_utils_constraints::{
+    constraints_and,
+    ClauseOperator::{self, And, Or},
+    ConstraintExpr, ConstraintKey, ConstraintValue, Constraints,
+};
+
+fn main() {
+    let constraints_1 = Constraints::new_clause(
+        And,
```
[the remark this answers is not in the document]
filipgolem: This function can be also used to create OR clauses, eg. (|(abc=1)(abc=2)), but most users will be using constraints! macro and .and() and .or() methods.

E.g. constraints![...].add(constraints![].or(constraints![])).

Cargo.toml
filipgolem: Good idea.

utils/constraints/src/lib.rs
filipgolem: This method is for creating ConstraintKey from key: &str, so it's better if it owns the key instead of having a reference to it, but I'm not sure.

utils/constraints/examples/constraints.rs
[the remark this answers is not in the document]
nieznanysprawiciel: So you need 2 functions add_clause and or_clause internally they can be translated to new_clause

utils/constraints/src/lib.rs
nieznanysprawiciel: Thats not what I meant. If you will write signature:
```
impl From<T: AsRef<str>> for ConstraintKey
...
```
You can write key.as_ref().to_string() and create new object as you do now.
But
advantage is, that you can call ConstraintKey::from passing as parameter whatever implements AsRef for str.
But it's just a suggestion, I don't know if you need it

Cargo.toml
filipgolem: Done!

utils/constraints/src/lib.rs
filipgolem: You are right. This was added mostly for &str (as library users would be directly typing &strs), but I'm going to improve this to handle everything that can be treated as a reference to str now.

filipgolem: Done.

utils/constraints/examples/constraints.rs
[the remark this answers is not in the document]
filipgolem: I understand what you mean, but in my opinion it's better to put operator as a parameter than as a part of a method name. Anyway, it's not very important: most users would be using constraints! macro and .and()/.or() methods; these constructors don't even have to be public.

PR #259, opened 2020-05-13
[gsb] caller support

tworec:
~This will register all unlocked identities on the central network instance (as opposite to registering only default one).~

~All incoming traffic will be handled out of the box.~
Outgoing traffic is more tricky, as it needs to have GSB caller properly set. 

EDIT: added ability to specify caller on service bus endpoints

EDIT2: after @prekucki has introduced from/to/service net addressing in #262 I've stripped this PR to only address adding caller on for GSB (also for bcasts)

### DIFF
--- a/core/net/examples/test_net_mk1.rs
+++ b/core/net/examples/test_net_mk1.rs
@@ -38,6 +38,7 @@ struct Options {
 enum Side {
     Listener,
     Sender,
+    Other,
 }
 
 impl Options {
@@ -45,6 +46,7 @@ impl Options {
         match side {
             Side::Listener => "0xbabe000000000000000000000000000000000000",
             Side::Sender => "0xfeed000000000000000000000000000000000000",
+            Side::Other => "0xbeef000000000000000000000000000000000000",
         }
         .parse()
         .unwrap()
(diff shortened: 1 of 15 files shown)


Wiezzel:
This functionality is duplicated in #262. Please coordinate with @prekucki

core/net/examples/test_net_mk1.rs
```diff
@@ -91,12 +91,32 @@ async fn main() -> Result<()> {
...
+                .parse()
+                .unwrap();
             let r = listener_id
                 .try_service(net::PUBLIC_PREFIX)?
-                .send(Test("Test msg".into()))
+                .send_as(caller_id, Test("Test 1 msg".into()))
                 .map_err(Error::msg)
                 .await?;
-            log::info!("Sending Result: {:?}", r);
+            assert!(r.is_ok());
```
mfranciszkiewicz, line 102: Maybe we could unit test the root cause of the failure to keep the example cleaner?

service-bus/proto/protos/gsb_api.proto
```diff
@@ -115,8 +115,10 @@ message UnsubscribeReply {
 }
 
 message BroadcastRequest {
-  string topic = 1;
-  bytes data = 2;
+  string caller = 1;
+  string topic = 2;
+  string request_id = 3;
```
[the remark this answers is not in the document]
tworec: will remove and restore tags for back compat.

core/net/examples/test_net_mk1.rs
tworec: what do you mean?

mfranciszkiewicz: That we most probably could unit test why an unknown sender (`ya_net::from`) would fail the call, so we could  clean up the assertions. On a second look, the example is brief enough so please disregard this comment

tworec: this is not unit test, but rather an integration test
I'm adding more comments in form of assert custom messages

PR #262, opened 2020-05-14
Feature/net bcast

prekucki:
* bcast api for net
* from/to sending for net


core/net/src/service.rs
```diff
@@ -19,27 +23,44 @@ pub fn central_net_addr() -> std::io::Result<SocketAddr> {
         .expect("central net hub addr needed"))
 }
 
+#[inline]
+fn net_node_id(node_id: &NodeId) -> String {
+    format!("{}/{:?}", net::BUS_ID, node_id)
+}
+
+fn parse_from_addr(from_addr: &str) -> anyhow::Result<(NodeId, NodeId, &str)> {
```
Wiezzel, line 31: Could de done easier with a regex: `/from/(\w+)/to/(\w+)(/[\w/]+)`

core/net/src/api.rs
```diff
@@ -31,8 +31,52 @@ impl TryRemoteEndpoint for &str {
...
+    pub fn to(&self, dst: NodeId) -> NetDst {
+        NetDst {
+            identity: self.identity,
+            dst,
+        }
+    }
+}
+
+pub struct NetDst {
+    identity: NodeId,
```
Wiezzel, line 48: Both fields are identities. I suggest to rename `identity` to `src`.

core/net/src/service.rs
prekucki: it won't:

- regex compilation runs in runtime. so i need lazy static or something.

does not looks very simpler?. and it is impossible to return borrowed slice from an argument. (or i do not know how to do this). so it has additional allocations on heap.

```rust
fn parse_from_addr(from_addr: &str) -> anyhow::Result<(NodeId, NodeId, String)> {
    let re = regex::Regex::new("/from/(\\w+)/to/(\\w+)(/[\\w/]+)")?;
    if let Some(c) = re.captures(from_addr) {
        let (from_node_id, to_node_id, service_id) = (&c[1], &c[2], &c[3]);
        return Ok((from_node_id.parse()?, to_node_id.parse()?, service_id.to_owned()));
    }
    Err(anyhow!("invalid net-from destination: {}", from_addr))
}
```

core/net/src/api.rs
prekucki: i was thinking about `src` when i wrote this. I was thinking about identity selection for the caller.

core/net/src/service.rs
Wiezzel: It's clearer to see what the function does, but if it's too much trouble then never mind.

PR #263, opened 2020-05-14
fix ya-client rev dep

tworec:


tworec:
already fixed by #261

PR #264, opened 2020-05-14
Use NodeId type from client crate

Wiezzel:
Replace all usage of `ya_core_model::ethaddr::NodeId` with `ya_client_model::NodeId`.


core/market-forwarding/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2018"
 
 [dependencies]
 ya-client = "0.3"
+ya-client-model = "0.1"
```
tworec, line 10: not required, since it is visible via `ya_client::model`

core/market-forwarding/src/api.rs
```diff
@@ -3,7 +3,7 @@ use jsonwebtoken::{encode, Header};
 use serde::{Deserialize, Serialize};
 use url::Url;
 
-use ya_core_model::ethaddr::NodeId;
+use ya_client_model::NodeId;
```
tworec, line 6:
```suggestion
use ya_client::model::NodeId;
```

core/market-forwarding/Cargo.toml
Wiezzel: OK, I will update this for all crates.

core/payment/Cargo.toml
```diff
@@ -10,7 +10,7 @@ gnt-driver = []
 dummy-driver = []
 
 [dependencies]
-ya-core-model = { version = "0.1", features = ["payment", "market", "identity", "with-diesel"] }
+ya-core-model = { version = "0.1", features = ["payment", "market", "identity"] }
 ya-client-model = "0.1"
```
tworec, line 14: don't you need?
```suggestion
ya-client-model = { version = "0.1", features = [ "with-diesel" ] }
```

Wiezzel: Looks like it's needed, but the compiler doesn't complain... :thinking:

core/payment/src/service.rs
```diff
@@ -146,7 +146,7 @@ mod public {
         };
 
         let offeror_id = agreement.offer.provider_id.clone().unwrap(); // FIXME: provider_id shouldn't be an Option
-        let issuer_id = debit_note.issuer_id.clone();
+        let issuer_id = debit_note.issuer_id.to_string();
```
tworec, line 149: id rather parse sender_id here and compare node_ids below

core/payment/Cargo.toml
tworec: try to compile from within the crate directory... ;)

core/persistence/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 
 [dependencies]
-ya-core-model = { version = "0.1", features=["ethaddr"]}
+ya-core-model = "0.1"
 ya-client-model = "0.1"
```
tworec, line 10:
```suggestion
ya-client-model = { version = "0.1", features = [ "with-diesel" ] }
```

core/payment/src/service.rs
Wiezzel: But parsing requires error handling and this doesn't. Also, I would have to parse offeror_id as as well. So how is that better?

tworec: it was a suggestion / non-blocking comment :)

core/payment/Cargo.toml
Wiezzel: Oh, right. I think Cargo should detect this nevertheless. If building of any of the crates alone fails, so should building the whole workspace.

core/payment/src/service.rs
Wiezzel: OK :)

PR #268, opened 2020-05-19
Update readme's, dockerfiles, comments and scripts to reflect new workspace shape

tworec:


agent/provider/readme.md
```diff
@@ -3,12 +3,12 @@
 ## Central setup
 We have centrally deployed (ip: `34.244.4.185`) three independent standalone modules/apps:
  - [net Mk1](https://github.com/golemfactory/yagna/blob/master/docs/net-api/net-mk1-hub.md) @ 34.244.4.185:7464 \
-   (can be invoked locally with `cargo run --release --example ya_sb_router`)
+   (can be invoked locally with `cargo run --release -p ya-sb-router --example ya_sb_router`)
```
etam, line 6: What does it do? The paragraph starts with saying there are deployed modules/apps. What does this do? Is it a client?

tworec: this paragraph is to give context to the reader. You do not need to instantiate those. They are centrally deployed for all. Does this make sense?

etam: I think so. The term "invoked locally" is misleading, because that means, you're running this command or your computer, which is true, but misses the point. I think "can be run locally" would be better.

agent/provider/readme.md
```diff
@@ -38,36 +38,77 @@ This can be displayed using `cargo run --bin ya-provider run --help`
...
+This is the first ExeUnit we've prepared for you.
+You need to clone its repository and build.
+In following sections we assume you've cloned it to the same directory where `yagna` is cloned.
+```
+cd ../..  # assuming you are in ./yagna/ya-prov
+git clone git@github.com:golemfactory/ya-runtime-wasi.git
+cd ya-runtime-wasi
+cargo build --release
+cd ../yagna/ya-prov
+```
```
nieznanysprawiciel, line 62: We force users to build ExeUnits separately. It's very cumbersome.
We should think about publishing binaries on our servers and make download commands in provider.

agent/provider/readme.md
tworec: ok, I've moved this section to the bottom, end reworded as you suggested. take a look

agent/provider/readme.md
tworec: yes, we should... and dozen other user-facing hassles :]

agent/provider/readme.md
```diff
@@ -38,36 +27,77 @@ This can be displayed using `cargo run --bin ya-provider run --help`
...
 
+## ExeUnits
+
+## WASI (wasmtime)
+
+This is the first ExeUnit we've prepared for you.
+You need to clone its repository and build.
+In following sections we assume you've cloned it to the same directory where `yagna` is cloned.
+```
+cd ../..  # assuming you are in ./yagna/ya-prov
```
nieznanysprawiciel, line 57: You assume to be in provider directory, but local-debug-exeunits-descriptor.json points to _ya-runtime-wasi_ in yagna directory as I see

tworec: nope, check carefully, there is thriple times `..` going up in tree

agent/provider/readme.md
```diff
@@ -38,36 +27,77 @@ This can be displayed using `cargo run --bin ya-provider run --help`
...
+cd ya-runtime-wasi
+cargo build --release
+cd ../yagna/ya-prov
+```
+
+You can list available ExeUnits with command:
+
+```bash
+$ cargo run -p ya-provider exe-unit list \
+--exe-unit-path ../exe-unit/resources/local-debug-exeunits-descriptor.json
```
etam, line 68: Above the instruction says to build ya-runtime-wasi in release mode, but here and later you refer to it, as if it was built in debug mode.

agent/provider/readme.md
etam: You reworded 1 of 3, but it's going in the right direction ;)

agent/provider/readme.md
tworec: nice catch

agent/provider/readme.md
tworec: done

agent/provider/readme.md
tworec: fixed

PR #269, opened 2020-05-19
Market decentralized - database models

nieznanysprawiciel:


core/serv/Cargo.toml
```diff
@@ -0,0 +1,57 @@
+[package]
+name = "yagna"
```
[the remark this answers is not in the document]
nieznanysprawiciel: It's probably merge artifact (squashes are awful :( ).
Should I just delete it??

core/market/decentralized/src/db/models/subscription.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::str::FromStr;
+use thiserror::Error;
+use uuid::Uuid;
+
+use digest::generic_array::GenericArray;
+use ya_client::model::ErrorMessage;
+
+#[derive(Error, Debug)]
+pub enum SubscriptionParseError {
+    #[error("Subscription id [{}] has invalid format.", .0)]
```
tworec, line 18:
```suggestion
    #[error("Subscription id [{0}] has invalid format.")]
```

core/market/decentralized/src/market.rs
```diff
@@ -70,7 +90,64 @@ impl MarketService {
...
+    }
+
+    pub async fn subscribe_offer(&self, offer: Offer, id: Identity) -> Result<String, MarketError> {
+        let offer = ModelOffer::from_new(&offer, &id);
+        let subscription_id = offer.id.to_string();
+
+        self.provider_negotiation_engine
+            .subscribe_offer(&offer)
+            .await?;
+        self.matcher.subscribe_offer(&offer).await?;
```
tworec, line 107: swap

core/market/decentralized/src/market.rs
```diff
@@ -70,7 +90,64 @@ impl MarketService {
...
+        id: Identity,
+    ) -> Result<String, MarketError> {
+        let demand = ModelDemand::from_new(&demand, &id);
+        let subscription_id = demand.id.to_string();
+
+        self.requestor_negotiation_engine
+            .subscribe_demand(&demand)
+            .await?;
+
+        self.matcher.subscribe_demand(&demand).await?;
```
tworec, line 136: swap

core/market/decentralized/src/matcher/matcher.rs
```diff
@@ -69,23 +108,119 @@ impl Matcher {
...
+        self.db
+            .as_dao::<OfferDao>()
+            .create_offer(model_offer)
+            .await
+            .map_err(|error| OfferError::InsertOfferFailure(error))?;
+
+        // TODO: Run matching to find local matching demands. We shouldn't wait here.
+        // TODO: Handle broadcast errors. Maybe we should retry if it failed.
+        self.discovery
+            .broadcast_offer(model_offer.into_client_offer()?)
```
tworec, line 129: https://github.com/golemfactory/yagna/issues/299

diesel.toml
```diff
@@ -0,0 +1,5 @@
+# For documentation on how to configure this file,
```
tworec: remove this file pls

core/market/decentralized/src/api/response.rs
```diff
@@ -0,0 +1,47 @@
+use actix_web::HttpResponse;
+use serde::Serialize;
+use ya_client::model::ErrorMessage;
+
+pub fn ok<T: Serialize>(t: T) -> HttpResponse {
+    HttpResponse::Ok().json(t)
+}
+
+pub fn created<T: Serialize>(t: T) -> HttpResponse {
+    HttpResponse::Created().json(t)
```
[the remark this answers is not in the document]
nieznanysprawiciel: For later

core/market/decentralized/src/api/provider.rs
```diff
@@ -120,3 +118,21 @@ async fn reject_agreement(
 ) -> HttpResponse {
     response::not_implemented()
 }
+
+#[actix_web::delete("/agreements/{agreement_id}/terminate")]
```
tworec, line 122:
```suggestion
#[actix_web::post("/agreements/{agreement_id}/terminate")]
```

core/market/decentralized/src/db/models/demand.rs
```diff
@@ -0,0 +1,88 @@
...
+        let creation_time = Utc::now().naive_utc();
+
+        Ok(Demand {
+            id,
+            properties,
+            constraints,
+            node_id,
+            creation_time: creation_time.clone(),
+            addition_time: creation_time.clone(),
+            expiration_time: creation_time.clone(),
```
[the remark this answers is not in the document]
tworec: Can you pls declare some -- even hardcoded -- TTL and set `exp_ts = creation_time  + TTL`?

core/market/decentralized/src/matcher/matcher.rs
```diff
@@ -4,21 +4,60 @@ use thiserror::Error;
...
+use crate::protocol::{
+    Discovery, DiscoveryBuilder, DiscoveryError, DiscoveryFactory, DiscoveryInitError,
+};
 use crate::protocol::{OfferReceived, RetrieveOffers};
 
 #[derive(Error, Debug)]
-pub enum MatcherError {}
+pub enum DemandError {
+    #[error("Failed to insert Demand. Error: {}.", .0)]
+    InsertDemandFailure(#[from] DbError),
```
tworec, line 24: I feel like this error (insert/remove/not-found) belongs rather to DAO than here, isn't it?  

Secondly, having `#[from]` you can use plain `?` without need of mapping error (see [line 142](https://github.com/golemfactory/yagna/pull/269/files#diff-c373b03f716ab716b09699faa1c50599R142)), but you can consider adding here `subscription_id` (see my suggestion below)

And lastly, change other error msgs below to use short syntax `{0}` also

```suggestion
    #[error("Failed to insert Demand[{1}] : {0}")]
    InsertDemandFailure(DbError, String),
```

core/market/decentralized/src/matcher/matcher.rs
```diff
@@ -4,21 +4,60 @@ use thiserror::Error;
...
 pub enum MatcherInitError {
     #[error("Failed to initialize Discovery interface. Error: {}.", .0)]
     DiscoveryError(#[from] DiscoveryInitError),
     #[error("Failed to initialize database. Error: {}.", .0)]
     DatabaseError(#[from] DbError),
+    #[error("Failed to migrate market database. Error: {}.", .0)]
+    MigrationError(#[from] anyhow::Error),
 }
 
 /// Receivers for events, that can be emitted from Matcher.
```
tworec, line 63: missed that before, but I'd rather use generics instead of `dyn` in Matcher struct: [a few lines below](https://github.com/golemfactory/yagna/pull/269/files#diff-c373b03f716ab716b09699faa1c50599R70)

nieznanysprawiciel: Which one _dyn_? This link doesn't point to anything :(

core/market/decentralized/src/api/provider.rs
nieznanysprawiciel: Changed in both Provider and Requestor

core/market/decentralized/src/matcher/matcher.rs
nieznanysprawiciel: 1. Database errors are rather ugly. I want to add more descriptive message in matcher.
2. You need error mapping when you have multiple enum variants that can be created from one error type.
3. `Subscription_id` doesn't make sense in case of insert. It was generated while ago and doesn't give any value for provider agent, that never saw this id before and will never see it again since adding Offer failed.

I changed message syntax to `{0}` as you requested in whole matcher

core/market/decentralized/src/matcher/matcher.rs
nieznanysprawiciel: Ok, probably you meant Arc<dyn Discovery>.
It will disappear in https://github.com/golemfactory/yagna/pull/294, because thsi abstraction won't be necessary anymore.

But meanwhile, when we are speaking about this, I totally disagree, that it would be better.
If Matcher was generic, all structs that reference it, would have to be generic too. In general case it means generics explosion in top level struct (Market in this case). Maybe in current implementation it wouldn't happen, but you were talking here about general rules here, as I understand.
And what are actually disadvantages of `dyn` that are important here?

PR #270, opened 2020-05-19
Bug/driver db constraints

bidzyyys:


bidzyyys:
@Wiezzel  It would be better solution, but we do not have any info about sender while processing tx (e.g. while sending, updating status or updating payment staus).

PR #275, opened 2020-05-21
add `--one-agreement` argument to requestor

etam:


agent/requestor/src/main.rs
```diff
@@ -205,7 +207,10 @@ async fn main() -> anyhow::Result<()> {
                 exe_script.clone(),
                 commands_cnt,
                 activities.clone(),
-            ))
+            ));
+            if settings.one_agreement {
```
tworec, line 211: have you tested it? :)

agent/requestor/src/main.rs
```diff
@@ -48,6 +48,8 @@ struct AppSettings {
     /// - Provider was not as performant as he declared
     #[structopt(long, default_value = "15min")]
     pub task_expiration: Duration,
+    #[structopt(long)]
```
tworec, line 51: pls add a comment to appear in CLI help

agent/requestor/src/market.rs
```diff
@@ -69,6 +69,7 @@ pub(crate) async fn spawn_negotiations(
     allocation_size: i64,
     agreement_allocation: Arc<Mutex<HashMap<String, String>>>,
     agreement_tx: mpsc::Sender<String>,
+    one_agreement: bool,
```
tworec, line 72: I think it is not necessary to inform this part of code that we want to process only one agreement
It is better to break the while loop which is processing agreements

agent/requestor/src/payment.rs
```diff
@@ -105,6 +109,8 @@ async fn process_invoice(
     payment_api: PaymentRequestorApi,
     invoice_id: String,
     agreement_allocation: Arc<Mutex<HashMap<String, String>>>,
+    one_agreement: bool,
+    ctrl_c_abort: futures::future::AbortHandle,
```
tworec, line 113: you can pass Option here, and if it is `Some(ctrl_c_abort)` than call abort on it. It will remove one extra param

agent/requestor/src/main.rs
```diff
@@ -195,20 +192,38 @@ async fn main() -> anyhow::Result<()> {
...
+                    activities.clone(),
+                ));
+            }
+        });
+    }
+
+    match future_ctrl_c.await {
+        Ok(Ok(())) => log::info!("Caught ctrl-c"),
+        Ok(Err(error)) => log::error!("Failed listening to ctrl-c: {}", error),
+        Err(futures::future::Aborted) => log::info!("Listening to ctrl-c aborted"),
```
tworec, line 218: would replace this technical info into some more meaningful:
```suggestion
        Err(futures::future::Aborted) => log::info!("`Finished processing one Agreement. Payment is confirmed. Finishing`"),
```

agent/requestor/src/main.rs
```diff
@@ -156,19 +156,15 @@ async fn main() -> anyhow::Result<()> {
     let subscription_id = api.market.subscribe(&my_demand).await?;
     log::info!("\n\n DEMAND SUBSCRIBED: {}", subscription_id);
 
-    let shutdown = shutdown_handler(
-        activities.clone(),
-        agreement_allocation.clone(),
-        subscription_id.clone(),
-        api.clone(),
-    );
+    let (future_ctrl_c, ctrl_c_abort) = futures::future::abortable(signal::ctrl_c());
```
tworec, line 159: I'd rather hide the details of how we process ctrl_c. Can you pls move it back to `shutdown_handler` which in turn will return future to wait for at the end together with abort handler?

agent/requestor/src/main.rs
```diff
@@ -195,20 +192,38 @@ async fn main() -> anyhow::Result<()> {
...
+                ));
+            }
+        });
+    }
+
+    match future_ctrl_c.await {
+        Ok(Ok(())) => log::info!("Caught ctrl-c"),
+        Ok(Err(error)) => log::error!("Failed listening to ctrl-c: {}", error),
+        Err(futures::future::Aborted) => log::info!("`Finished processing one Agreement. Payment is confirmed. Finishing`"),
+    }
```
tworec, line 219: This also should be moved into shutdown handler I think

agent/requestor/src/main.rs
tworec: eg like this
`let (shutdown, app_abort_handle) = shutdown_handler(...)`
then at the end of `main` you can : 
`shutdown.await`
and in the payments:
`app_abort_handle.abort()`

agent/requestor/src/main.rs
```diff
@@ -125,6 +126,28 @@ async fn shutdown_handler(
     Arbiter::current().stop();
 }
 
+fn shutdown_handler(
```
tworec, line 129: 👍

agent/requestor/src/main.rs
```diff
@@ -156,7 +179,7 @@ async fn main() -> anyhow::Result<()> {
     let subscription_id = api.market.subscribe(&my_demand).await?;
     log::info!("\n\n DEMAND SUBSCRIBED: {}", subscription_id);
 
-    let shutdown = shutdown_handler(
+    let (future_shutdown, app_abort_handle) = shutdown_handler(
```
tworec, line 182: a common name for future
```suggestion
    let (shutdown_fut, app_abort_handle) = shutdown_handler(
```

agent/requestor/src/main.rs
```diff
@@ -195,20 +218,32 @@ async fn main() -> anyhow::Result<()> {
         payment_api.clone(),
         started_at,
         agreement_allocation.clone(),
+        if settings.one_agreement {
+            Some(app_abort_handle.clone())
+        } else {
+            None
+        },
```
tworec, line 225: shorter and more-readable -- just a case to know more rust-y way of doing this. You need not to follow me here 
```suggestion
        match settings.one_agreement {
            true => Some(app_abort_handle.clone())
            false => None
        },
```

agent/requestor/src/main.rs
```diff
@@ -195,20 +218,32 @@ async fn main() -> anyhow::Result<()> {
...
-                agreement_id,
-                exe_script.clone(),
-                commands_cnt,
-                activities.clone(),
-            ))
-        }
-    });
+    {
+        let activities = activities.clone();
+        let api = api.clone();
```
tworec, line 230: is not needed in current code shape
```suggestion
```

agent/requestor/src/payment.rs
```diff
@@ -158,4 +161,8 @@ async fn process_invoice(
             }
         }
     }
+
+    if let Some(handle) = app_abort_handle {
+        handle.abort();
+    }
```
tworec, line 167:
```suggestion
    app_abort_handle.map(|h| h.abort());
```

etam:
> etam dismissed tworec’s stale review via afa89ab 1 minute ago

that's something new

tworec:
Not so new
https://github.com/isaacs/github/issues/783#issuecomment-320211882

PR #278, opened 2020-05-26
db separation & service ctx passed also to rest activator 

tworec:
Market, Activity and Payment services has their own db files from now on.
This also enables access to DB from both GSB and rest activators.

As a side kick it also contains `ya-persistance` cleanup: migrations, models and schema is moved to `ya-activity`


core/serv/src/main.rs
```diff
@@ -136,35 +141,56 @@ impl TryFrom<&CliArgs> for CliCtx {
     }
 }
 
+#[derive(Clone)]
 struct ServiceContext {
-    db: DbExecutor,
-    data_dir: PathBuf,
```
nieznanysprawiciel, line 141: This version of decentralized market still has data_dir. It shouldn't compile

core/serv/src/main.rs
```diff
@@ -136,35 +141,56 @@ impl TryFrom<&CliArgs> for CliCtx {
...
-    fn component(&self) -> PathBuf {
-        self.data_dir.clone()
+impl ServiceContext {
+    fn make_entry<S: 'static>(path: &PathBuf, name: &str) -> Result<(TypeId, DbExecutor)> {
+        Ok((TypeId::of::<S>(), DbExecutor::from_data_dir(path, name)?))
+    }
+
+    fn from_data_dir(path: &PathBuf, name: &str) -> Result<Self> {
+        let default_db = DbExecutor::from_data_dir(path, name)?;
+        default_db.apply_migration(ya_persistence::migrations::run_with_output)?;
```
nieznanysprawiciel, line 168: This line will add activity migrations to all databases. In my code there were no migrations done here excluding default database. Each service runs migrations by itself.

nieznanysprawiciel: Maybe a little more explanation: activity database uses migrations defined in ya-persistence. But all other services have their own migrations sets.

tworec: I'll move `ya-persistence::migrations` to `ya-activity`

nieznanysprawiciel: It's probably good choice

core/serv/src/main.rs
tworec: done

tworec:
@prekucki @nieznanysprawiciel @mfranciszkiewicz pls approve

PR #285, opened 2020-05-28
[market-fwd] central market testbed url configurable

tworec:
To enable fully isolated integration tests we need configurable URL of central Market instance.

I'm also fixing and simplifying current default API URL by moving it definition to ya-client
related PR https://github.com/golemfactory/ya-client/pull/17

### DIFF
--- a/core/market/forwarding/src/api.rs
+++ b/core/market/forwarding/src/api.rs
@@ -8,8 +8,16 @@ use ya_client::model::NodeId;
 use ya_persistence::executor::DbExecutor;
 use ya_service_api_web::middleware::Identity;
 
+pub const CENTRAL_MARKET_URL_ENV_VAR: &str = "YAGNA_API_URL";
+pub const DEFAULT_CENTRAL_MARKET_URL: &str = "http://34.244.4.185:8080/market-api/v1/";
+
 lazy_static::lazy_static! {
-    pub(crate) static ref CENTRAL_MARKET_URL: Url = "http://34.244.4.185:8080/market-api/v1/".parse().unwrap();
+    pub(crate) static ref CENTRAL_MARKET_URL: Url =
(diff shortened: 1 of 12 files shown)


core/market/forwarding/src/api.rs
```diff
@@ -8,8 +8,16 @@ use ya_client::model::NodeId;
 use ya_persistence::executor::DbExecutor;
 use ya_service_api_web::middleware::Identity;
 
+pub const CENTRAL_MARKET_URL_ENV_VAR: &str = "YAGNA_API_URL";
```
maaktweluit, line 11: i think the env var should be `YAGNA_MARKET_URL`, since `YAGNA_API_URL` is missing the proper postfix

source: https://github.com/golemfactory/yagna/blob/master/.env-template#L21

tworec: of course
I had distractors writing it...

tworec:
@maaktweluit pls re-review

maaktweluit:
Thanks for the update!

What is the value of having `CENTRAL_MARKET_URL` besides `YAGNA_MARKET_URL`? i think the daemon should only use one market at the time.

if `CENTRAL_MARKET_URL` is added and valuable it would be good to also add it with description to `env.template`

tworec:
> What is the value of having `YAGNA_MARKET_URL` besides `YAGNA_MARKET_URL`? i think the daemon should only use one market at the time.

Demon is serving the market API -- this is why we need `YAGNA_MARKET_URL`.
And the default implementation of market is forwarding all requests to the central instance  -- that's why we need `CENTRAL_MARKET_URL` .

> if `CENTRAL_MARKET_URL` is added and valuable it would be good to also add it with description to `env.template`

I do not feel that average user should be aware of it. It is internal.

Issue #292, opened 2020-05-29
ExeUnit MVP features


mfranciszkiewicz:
- ~Execute multiple ExeScripts in parallel~ (merged #298)
  Limited to commands controlling the ExeScript execution flow

- New commands
  Commands to control the ExeScript execution flow
  - `SUSPEND` 
  - `RESUME`
  - ~`TERMINATE` - stops the execution of an ExeScript~ (merged #298)

- New commands for consideration
  - `FORWARD` - opens a TCP-like communication pipe between the payload of ExeUnit (eg. the hosted container) and the Requestor’s side App
  - `RAISEVENT` - raise an Event that can be observed by Provider, and potentially also other nodes, as per Observe right
  - `GRANT` / `REVOKE` - grant / revoke particular right on Activity ACL to specific golem Node 

- New Activity / ExeUnit States
  - `Suspended`
  - ~`Unresponsive` (issue #226)~ (merged #356)

- ~ExeUnit resource usage limiting~ (issue #334)

- ExeUnit supervisor CLI
  - ~Offer templates - templates include the capabilities of a particular ExeUnit (e.g. IPFS daemon availability)~ (https://github.com/golemfactory/yagna/pull/536)
  - Usage reporting interval option

- Misc
  - ~`TRANSFER` command retries~

PR #298, opened 2020-06-02
ExeUnit Supervisor: script termination

mfranciszkiewicz:
Termination scenario:

* a new ExeScript is received; the script starts with (or consists of) a TERMINATE command
* any other running ExeUnit::exec futures are interrupted via a oneshot channel (or fail due to Runtime shutdown)
* a Shutdown command is sent to the Runtime handler; the Runtime's state is expected to be reset
* pending commands are updated with an erroneous result (command interrupted)
* ExeUnit state is reverted to State::Initialized
* the remaining commands of the received script are executes as normally

Changes:

* new: handle the new TERMINATE command
* new: stop the currently executed script + runtime via the `Stop` message
* refactor: set ExeScriptCommandResult::is_batch_finished to true on command error
* refactor: do not terminate the Supervisor on command error
* refactor: disallow executing an ExeScript with an already known batchId
* refactor: move the body of ExeUnit::pre_exec to ExeUnit::exec
* refactor: example option to show a termination scenario flow

Part of #292

### DIFF
--- a/exe-unit/src/lib.rs
+++ b/exe-unit/src/lib.rs
@@ -1,4 +1,5 @@
 use actix::prelude::*;
+use futures::channel::oneshot;
 use futures::TryFutureExt;
 use std::path::PathBuf;
 use std::time::Duration;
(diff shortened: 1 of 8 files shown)


exe-unit/src/lib.rs
```diff
@@ -186,8 +178,21 @@ impl<R: Runtime> ExeUnit<R> {
         transfer_service: Addr<TransferService>,
         ctx: ExecCtx,
     ) -> Result<()> {
+        if let ExeScriptCommand::Terminate {} = &ctx.cmd {
+            log::warn!("Terminating running ExeScripts");
+
+            let exclude_batches = vec![ctx.batch_id];
+            let set_state = SetState::default()
+                .state(StatePair(State::Initialized, None))
```
tworec, line 186: is it intentional to set the `Initialized` state as a result of the `Terminate` command?

PR #315, opened 2020-06-03
Include last commit's SHA in .deb package version

azawlocki:
This PR modifies the workflow file for the `Build .deb` action and the dockerfiles that are used to build .deb packages, in order to include the SHA of the current commit in the `Version` property of the .deb packages. After this change, the version property of `yagna.deb` will be something like `0.3.0-8832501e` instead of just `0.3.0`.

This will allow us to identify which version of `yagna` code is actually being tested in integration tests.

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -16,14 +16,23 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Compute deb package version
+        ids: vars
+        run: |
+          VERSION="$(cargo pkgid | cut -d'#' -f2)"
+          ROUTER_VERSION=$(cargo pkgid --manifest-path service-bus/router/Cargo.toml | cut -d'#' -f2)"
+          SHA="$(echo ${{ github.sha }} | cut -c1-8)"
+          echo "::set-output name=deb_version::$VERSION-$SHA"
+          echo "::set-output name=router_deb_version::$ROUTER_VERSION-$SHA"
+
       - name: Build the Docker image
-        run: docker build . -f build-deb.Dockerfile -t build-deb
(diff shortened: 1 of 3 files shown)


.github/workflows/build-deb.yml
```diff
@@ -16,14 +16,23 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Compute deb package version
+        id: vars
+        run: |
```
tworec, line 21: there is nice syntax faciliating setting env var like this. Also, consider renaming `SHA` into sth more meaningful like
```suggestion
        env:
          COMMIT_SHA: ${{ github.sha }}
        run: |
```

tworec:
there is also nice crate : https://crates.io/crates/git-version

.github/workflows/build-deb.yml
azawlocki: Thanks for comments, I've included your suggestions in my latest comment (https://github.com/golemfactory/yagna/commit/3eccf0ae1a5c938fd1bb20d949b4bad8869038af)

azawlocki:
As for adding commit hash to version reported by `yagna` binary, I'll do this in a separate PR

build-deb-router.Dockerfile
```diff
@@ -1,3 +1,4 @@
 FROM build-deb
+ARG DEB_VERSION=unknown
```
tworec, line 2: 👍

tworec:
I forgot to add, that we'd rather not commit such "well-grained" commits individually, but rather as a squash... especially those for moving branch activator back-and-forward ;)

azawlocki:
> @azawlocki I forgot to add, that we'd rather not merge such "well-grained" commits individually to master, but rather as a squash commit... especially those moving branch activator back-and-forward ;)

Got it!

PR #344, opened 2020-06-15
Payment driver API

Wiezzel:
Rationale
----

New API introduced for the communication between payment service and payment driver(s). It is meant to enable the implementation of the following requirements:
1. Supporting multiple accounts and payment drivers on a single yagna daemon.
2. Supporting multiple payment platforms (e.g. GNT, NGNT, NGNT-zkSync).
3. Allowing third-party developers to implement their own payment drivers with ease and possibly much freedom.

Detailed list of changes
----

`BUS_ID` is replaced with `BUS_ID_PREFIX` because each particular driver will have its own ID.

Currency enum is abandoned. It is not desired to have a fixed set of available currencies. It is assumed that all amounts and balances are given in currency appropriate for the platform.

All gas-related parameters are abandoned. They were too driver-specific to be included in the general API. It is recommended that payment drivers which use gas allow to configure limits (e.g. by env variables).

`SchedulePayment` does not include `invoice_id` parameter anymore. Instead, it returns an ID of the payment order generated by the driver.

`GetPaymentStatus` is removed because the new payment driver API relies on push notifications instead. That is realized by the `NotifyPayment` endpoint added to the payment service. It shall be called by a driver after every successful transaction.

`Init` message is removed from payment API (GSB). The init command from CLI will accept `driver` parameter and route the call directly to the appropriate driver.

`RegisterAccount` / `UnregisterAccount` messages are added because the payment service needs to maintain a list of available accounts. Payment drivers should call `RegisterAccount` after every successful initialization of an account to make it 'visible' to the payment service. `UnregisterAccount` should be called when the driver is being shut down. The list of available accounts shall not be persisted by the payment service. If an account doesn't need to be re-initialized after restarting yagna daemon, then payment driver should store this information and call `RegisterAccount` immediately after startup.


Wiezzel:
Issue I failed to address in this API design: account permissions. Some kind of mechanism is necessary to control which identities are allowed to use which payment accounts. I couldn't figure out how and  where to store this information appropriately. This however doesn't necessarily have to be done right now (we might just assume for the moment that every identity is allowed to use every payment account).

Wiezzel:
Example flow:
1. Yagna daemon with two drivers starts
    Simple GNT driver connects to GSB at `local/driver/gnt`
    ZkSync driver connects to GSB at `local/driver/gnt-zksync`
2. User initializes their wallet `0xdeadbeef...` in simple GNT driver via CLI
    `yagna payment init gnt -r 0xdeadbeef...`
    `Init("0xdeadbeef...", R)` message is sent to GNT driver via GSB
    GSB driver sends `RegisterAccount("gnt", "gnt", "0xdeadbeef...", R)` to the payment service
3. User check their account balance via CLI
    `yagna payment status 0xdeadbeef...`
    Payment looks up in a hashmap that account `0xdeadbeef...` is registered as handled by the GNT driver operating on GNT platform so it sends `GetAccountBalance("gnt", "0xdeadbeef...")` to the GNT driver.
    Reported balance is displayed to the user.
4. User initializes their wallet `0xc0deface...` in GNT-zkSync driver via CLI
    `yagna payment init gnt-zksync -rs 0xc0deface...`
    `Init("0xc0deface...", R+S)` message is sent to GNT-zkSync driver via GSB
    GSB-zkSync driver sends `RegisterAccount("gnt-zksync", "gnt-zksync", "0xc0deface...", R+S)` to the payment service
5. Requestor agent acting as the user creates a demand for computation.
    During demand creation market service asks payment service for available payment methods (*we're probably going to need a new endpoint for that*).
    Payment service loops through all available accounts and concludes that only `0xc0deface...` is in send mode so it is returned as the only result.
6. After the demand is matched an agreement is created specifying `0xc0deface` as payer address and `"gnt-zksync"` as payment platform.
7. When the computation is done and the invoice gets accepted, payment service orders payment
    Information about platform and account is retrieved from the market service (invoice has a relationship to the agreement).
    Payment service looks up ("gnt-zksync", "0xc0deface...")` in the hasmap and establishes that it is handled by the gnt-zksync driver so it sends `SchedulePayment` message with all information to the gnt-zksync driver.
    Driver responds with payment order ID `"12345-abcdd-65421"`. That ID is stored by the payment service.
8. When the transaction is confirmed on blockchain, GNT-zkSync driver sends `NotifyPayment` message with all the details to the payment service.
9. Payment service forwards the payment details to the provider node.
10. On the provider node, another payment service verifies the transaction by looking up the appropriate driver for the payee account and sending `VerifyPayment` message.

Wiezzel:
Other interesting scenario might be an integration with Ledger Liver that:
* Uses Ledger hardware wallet for signing transactions instead of yagna identity.
* Is initialized without using yagna CLI at all, just Ledger Live app starting, connecting to GSB and calling RegisterAccount.
* Prompts the user to confirm the transaction when needed.

In general, I believe that integrating existing third-party wallets might be the most important use cases for the payment driver API (apart from our own driver, ofc). I tried to design the API in way that would allow this quite smoothly.

core/model/src/driver.rs
```diff
@@ -115,12 +66,9 @@ impl PaymentConfirmation {
...
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct GetAccountBalance(String);
-
-impl From<String> for GetAccountBalance {
-    fn from(address: String) -> Self {
-        GetAccountBalance(address)
-    }
+pub struct GetAccountBalance {
+    pub platform: String,
```
bidzyyys, line 70: IMHO we do not need to send platform directly in message, it should be only specified in gsb router endpoint.

core/model/src/driver.rs
```diff
@@ -211,23 +136,23 @@ impl RpcMessage for Init {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct SchedulePayment {
-    invoice_id: String,
-    amount: PaymentAmount,
+    platform: String,
```
bidzyyys, line 139: Same here

core/model/src/driver.rs
```diff
@@ -257,21 +182,22 @@ impl SchedulePayment {
...
     type Error = GenericError;
 }
 
 // ************************** GET TRANSACTION BALANCE **************************
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct GetTransactionBalance {
-    sender: String,
-    recipient: String,
+    pub platform: String,
```
bidzyyys, line 193: Same here.

core/model/src/payment.rs
```diff
@@ -69,37 +70,62 @@ pub mod local {
...
+        pub amount: BigDecimal,
+        pub sender: String,
+        pub recipient: String,
+        pub order_ids: Vec<String>,
+        pub confirmation: PaymentConfirmation,
+    }
+
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct GetStatus{
+        pub platform: String,
```
bidzyyys, line 127: Same here.

core/model/src/payment.rs
```diff
@@ -69,37 +70,62 @@ pub mod local {
...
     }
 
-    impl GetStatus {
-        pub fn identity(&self) -> NodeId {
-            self.0
-        }
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct NotifyPayment {
+        pub driver: String,
+        pub platform: String,
```
bidzyyys, line 117: Same here.

bidzyyys: I agree that in notification message we should specify the driver, but when we have info about the driver `platform` is redundant.

core/model/src/payment.rs
```diff
@@ -69,37 +70,62 @@ pub mod local {
...
+        pub amount: BigDecimal,
+        pub sender: String,
+        pub recipient: String,
+        pub order_ids: Vec<String>,
```
bidzyyys, line 121: We do not have `order_ids` in payment driver so far

core/model/src/payment.rs
```diff
@@ -69,37 +70,62 @@ pub mod local {
...
+        const ID: &'static str = "RegisterAccount";
         type Item = ();
-        type Error = GenericError;
+        type Error = RegisterAccountError;
     }
 
     #[derive(Clone, Debug, Serialize, Deserialize)]
-    pub struct GetStatus(NodeId);
+    pub struct UnregisterAccount {
+        pub platform: String,
```
bidzyyys, line 96: Same here

core/model/src/payment.rs
```diff
@@ -69,37 +70,62 @@ pub mod local {
     }
 
     #[derive(Clone, Debug, Serialize, Deserialize)]
-    pub struct Init {
-        pub identity: NodeId,
-        pub provider: bool,
-        pub requestor: bool,
+    pub struct RegisterAccount {
+        pub driver: String,
+        pub platform: String,
```
bidzyyys, line 75: Same here, both `driver` and `platform` are redundant.

bidzyyys:
> Issue I failed to address in this API design: account permissions. Some kind of mechanism is necessary to control which identities are allowed to use which payment accounts. I couldn't figure out how and where to store this information appropriately. This however doesn't necessarily have to be done right now (we might just assume for the moment that every identity is allowed to use every payment account).

@Wiezzel not exactly, think about the situation when somebody schedule payment as you and the you unlock the account -> driver sends transaction immediately. It is a possible attack

core/model/src/payment.rs
bidzyyys: IMHO only `driver` is necessary.

core/model/src/payment.rs
bidzyyys, line 96: IMHO `driver` should be better and consistent with `RegisterAccount`

core/model/src/driver.rs
Wiezzel: This depends on whether we allow a single driver to support more than one platform. If no, then we can drop 'platform' parameter from most messages.

core/model/src/payment.rs
Wiezzel: Well, this API proposal introduces them. They should work more or less like `invoice_id` in current API but will be generated by the driver, not given as parameter.

core/model/src/payment.rs
Wiezzel: I believe the `platform` parameter has to stay here. Payment service has no other way to know which driver handles which platform.

core/model/src/payment.rs
Wiezzel: The mapping stored by payment service is `(platform, address) -> (driver, mode)`. To remove an entry from a map we need to give the key as parameter. `driver` is not part of the key, while `platform` is.

Wiezzel:
> @Wiezzel not exactly, think about the situation when somebody schedule payment as you and then you unlock the account -> driver sends transaction immediately. It is a possible attack.

When I wrote "This however doesn't necessarily have to be done right now" I meant "We may ignore security for now". Obviously, not having permissions is insecure but we are on testnet now.

core/model/src/payment.rs
bidzyyys: Can single transaction  include multiple payments? If yes, sounds good to me :+1:

core/model/src/payment.rs
Wiezzel: Here it can't be done away with because we don't know the driver. `address` is not enough to establish that. There can be two different drivers handling e.g. GNT and ETH transactions but using the same account address.

PR #345, opened 2020-06-15
Include commit hash in version number reported by `yagna --version`

### LINKED ISSUE #394: Add commit SHA to the version reported by `yagna --version`
`yagna --version` now prints package version from `Cargo.toml`, which does not change very often. It would be useful if the version included also SHA of the current commit (or rather some initial fragment of it), like this:
```
$ yagna --version
yagna 0.3.0-b346d83
```

azawlocki:
Resolves #394.

This PR adds a package `ya-compile-time-utils` with a macro `define-version-string`.
The macro expands to a declaration of a static variable `VERSION: String` that contains the string
```
${CARGO_PKG_VERSION}-${GITHUB_SHA}
```
if the environment variable `GITHUB_SHA` is defined (in particular, when executing a github action), and
```
${CARGO_PKG_VERSION}
```
otherwise.


service-bus/router/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "GPL-3.0"
 edition = "2018"
 
 [dependencies]
+ya-compile-time-utils = "0.1"
```
tworec, line 12: ya_sb_router is an example so this might be moved into `[dev-dep..s]`

utils/compile-time-utils/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+#[macro_export]
+macro_rules! define_version_string {
+    () => {
+        #[macro_use]
+        extern crate lazy_static;
+
+        lazy_static! {
```
tworec, line 7: I think it would be more readable and usable to define it like [`clap::crate_version`](https://github.com/clap-rs/clap/blob/81457178fa7e055775867ca659b37798b5ae9584/src/macros.rs#L34-L55) which enables using this inplace.

I think lazy static is not adding much value here

azawlocki: With `lazy_static` I can make the definition of the version string conditional: it's `"${CARGO_PKG_VERSION}-${GITHUB_SHA}"` if `GITHUB_SHA` is defined and `"${CARGO_PKG_VERSION}"` otherwise. I don't know how to do this using only macros.

service-bus/router/Cargo.toml
azawlocki: Fixed!

tworec:
After my change version is reported as
```
$ yagna -V
yagna 0.3.0-5c5b6c36
```

or when git index is dirty

```
$ yagna -V
yagna 0.3.0-af0617a0-modified
```

tworec:
I've downloaded https://github.com/golemfactory/yagna/suites/869116831/artifacts/10086600

```
awokado:~/test-yagna-binaries-version-commit-sha2$ unzip Yagna\ Linux\ \(2\).zip 
Archive:  Yagna Linux (2).zip
  inflating: exe-unit                
  inflating: ya-provider             
  inflating: ya-requestor            
  inflating: yagna                   
awokado:~/test-yagna-binaries-version-commit-sha2$ chmod a+x exe-unit ya*
awokado:~/test-yagna-binaries-version-commit-sha2$ ll
total 69460
drwxrwxr-x  2 tworec tworec     4096 Jul  2 17:30  ./
drwxr-xr-x 56 tworec root       4096 Jul  2 17:30  ../
-rw-r--r--  1 tworec tworec 18532602 Jul  2 17:30 'Yagna Linux (2).zip'
-rwxrwxr-x  1 tworec tworec 10100736 Jul  2 15:15  exe-unit*
-rwxrwxr-x  1 tworec tworec  9082920 Jul  2 15:15  ya-provider*
-rwxrwxr-x  1 tworec tworec  7895520 Jul  2 15:15  ya-requestor*
-rwxrwxr-x  1 tworec tworec 25497664 Jul  2 15:15  yagna*
awokado:~/test-yagna-binaries-version-commit-sha2$ ./yagna -V
yagna 0.3.0-9c56c6a4
awokado:~/test-yagna-binaries-version-commit-sha2$ ./exe-unit -V
ya-exe-unit 0.1.0-9c56c6a4
awokado:~/test-yagna-binaries-version-commit-sha2$ ./ya-provider -V
ya-provider 0.1.0-9c56c6a4
awokado:~/test-yagna-binaries-version-commit-sha2$ ./ya-requestor -V
ya-requestor 0.1.0-9c56c6a4

```

azawlocki:
Thanks @tworec for helping with this!

utils/compile-time-utils/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+pub use git_version::git_version;
+
+#[macro_export]
+macro_rules! crate_version_commit {
+    () => {
+        Box::leak(
+            [
+                env!("CARGO_PKG_VERSION"),
+                "-",
+                ya_compile_time_utils::git_version!(),
```
mfranciszkiewicz, line 10: (not tested)

This should strip the `-` character if it's not possible to retrieve the git version (and thus does not require git since it would fail silently):
`git_version!(prefix = "-", fallback = "")`

mfranciszkiewicz:
Perhaps commit versions should be feature-enabled?

tworec:
I've also checked some scenarios when `git` is non functioning @ build host and version is just from `Cargo.toml`

PR #346, opened 2020-06-15
Market decentralized - query events

nieznanysprawiciel:
Changes couldn't be tested with real requestor, since we don't have resolver and proposals generation yet.
This PR works in tests.


core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -7,15 +7,15 @@ CREATE TABLE market_offer (
 	node_id VARCHAR(20) NOT NULL,
 
 	creation_ts DATETIME NOT NULL,
-	insertion_ts DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+	insertion_ts DATETIME NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
```
tworec, line 10: 👍

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -26,6 +26,85 @@ CREATE TABLE market_demand (
...
+
+CREATE TABLE market_requestor_event(
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    subscription_id INTEGER NOT NULL,
+    timestamp DATETIME NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+    event_type INTEGER NOT NULL,
+    artifact_id VARCHAR(100) NOT NULL,
+
+    FOREIGN KEY(event_type) REFERENCES market_event_type (id),
+    CHECK (event_type>=2000)
```
tworec, line 54: a bit hacky.. why do we need role if we do it like this?

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -26,6 +26,85 @@ CREATE TABLE market_demand (
...
+
+    properties TEXT NOT NULL,
+    constraints TEXT NOT NULL,
+
+    state INTEGER NOT NULL,
+	creation_ts DATETIME NOT NULL,
+	expiration_ts DATETIME NOT NULL,
+
+	FOREIGN KEY(state) REFERENCES market_proposal_state (id),
+	FOREIGN KEY(negotiation_id) REFERENCES market_negotiation (id)
```
tworec, line 107: identation

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -26,6 +26,85 @@ CREATE TABLE market_demand (
...
+    (1003, "PropertyQuery", "Provider"),
+    (2001, "Proposal", "Requestor"),
+    (2002, "PropertyQuery", "Requestor");
+
+CREATE TABLE market_requestor_event(
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    subscription_id INTEGER NOT NULL,
+    timestamp DATETIME NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+    event_type INTEGER NOT NULL,
+    artifact_id VARCHAR(100) NOT NULL,
```
tworec, line 51: this will be used to store `proposal_id` and `agreement_id`?

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -26,6 +26,85 @@ CREATE TABLE market_demand (
...
+    (1002, "Agreement", "Provider"),
+    (1003, "PropertyQuery", "Provider"),
+    (2001, "Proposal", "Requestor"),
+    (2002, "PropertyQuery", "Requestor");
+
+CREATE TABLE market_requestor_event(
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    subscription_id INTEGER NOT NULL,
+    timestamp DATETIME NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+    event_type INTEGER NOT NULL,
```
tworec, line 50: `event_type` has role, and there are two tables: one for each role. Is it really needed?

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -26,6 +26,85 @@ CREATE TABLE market_demand (
...
+    (4, "Expired");
+
+CREATE TABLE market_negotiation(
+    id VARCHAR(100) NOT NULL PRIMARY KEY,
+    subscription_id VARCHAR(97) NOT NULL,
+
+    offer_id VARCHAR(97) NOT NULL,
+    demand_id VARCHAR(97) NOT NULL,
+
+    identity_id VARCHAR(20) NOT NULL,
```
tworec, line 87: what is the relation of `identity_id` and below ids?

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
```diff
@@ -26,6 +26,85 @@ CREATE TABLE market_demand (
...
+    requestor_id VARCHAR(20) NOT NULL,
+    provider_id VARCHAR(20) NOT NULL,
+
+    agreement_id VARCHAR(100)
```
tworec, line 91: I can't see the Agreement table. I think I need live review...

core/market/decentralized/src/db/dao/demand.rs
```diff
@@ -19,6 +20,13 @@ impl<'c> AsDao<'c> for DemandDao<'c> {
     }
 }
 
+/// Returns state of Demand in database.
+pub enum DemandState {
+    Active(ModelDemand),
+    Expired(Option<ModelDemand>),
+    NotFound,
+}
```
tworec, line 28: 👍

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: Yes

nieznanysprawiciel: and property query

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: There's no Agreement table, because it is not necessary yet

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: Check structures in db::model first. Everything is described there in details. I didn't left comments in sql files

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: Would you remove on of tables than? Or Remove role from EventType?

core/market/decentralized/src/db/dao/events.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl<'c> AsDao<'c> for EventsDao<'c> {
+    fn as_dao(pool: &'c PoolType) -> Self {
+        Self { pool }
+    }
+}
+
+impl<'c> EventsDao<'c> {
+    pub async fn add_requestor_event(&self, proposal: Proposal) -> DbResult<()> {
```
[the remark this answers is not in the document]
tworec: You have already defined `OwnerType` enum, so it might be useful as a param for this fn

core/market/decentralized/src/db/models/events.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub subscription_id: SubscriptionId,
+    pub event_type: EventType,
+    pub artifact_id: String,
+}
+
+impl MarketEvent {
+    pub fn from_proposal(proposal: &Proposal, role: OwnerType) -> NewMarketEvent {
+        NewMarketEvent {
+            subscription_id: proposal.negotiation.subscription_id.clone(),
+            event_type: match role {
```
tworec, line 66: again, here is matching. So we did not avoid matching of roles even though tables are denormalized.

core/market/decentralized/src/db/models/events.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    async fn into_requestor_proposal(self, db: DbExecutor) -> Result<RequestorEvent, EventError> {
+        let prop = db
+            .as_dao::<ProposalDao>()
+            .get_proposal(&self.artifact_id)
+            .await
+            .map_err(|error| EventError::FailedGetProposal(error))?
+            .ok_or(EventError::ProposalNotFound(self.artifact_id.clone()))?;
```
tworec, line 93: I believe those mappings belong to `ProposalDao` with dedicated Error type

core/market/decentralized/src/market.rs
```diff
@@ -41,8 +42,8 @@ pub enum MarketInitError {
 /// Structure connecting all market objects.
 pub struct MarketService {
     pub matcher: Arc<Matcher>,
-    pub provider_negotiation_engine: Arc<ProviderNegotiationEngine>,
-    pub requestor_negotiation_engine: Arc<RequestorNegotiationEngine>,
+    pub provider_engine: Arc<ProviderNegotiationEngine>,
```
tworec, line 45: 👍

tworec:
```suggestion
    pub provider_engine: Arc<ProviderNegotiatior>,
```
maybe, consistent with `Matcher` ?

core/market/decentralized/src/matcher/matcher.rs
```diff
@@ -316,6 +326,12 @@ impl From<ErrorMessage> for MatcherError {
     }
 }
 
+impl From<SubscriptionParseError> for MatcherError {
```
tworec, line 329: unused

core/market/decentralized/src/matcher/matcher.rs
```diff
@@ -287,20 +297,20 @@ async fn on_offer_unsubscribed(db: DbExecutor, msg: OfferUnsubscribed) -> Result
...
     }
     .await
-    .or_else(|e| {
-        let reason = match e {
+    .or_else(|error| {
+        let reason = match error {
             UnsubscribeError::OfferExpired => StopPropagateReason::Expired,
             UnsubscribeError::AlreadyUnsubscribed => StopPropagateReason::AlreadyUnsubscribed,
-            _ => StopPropagateReason::Error(e.to_string()),
+            _ => StopPropagateReason::Error(format!("{}", error)),
```
tworec, line 313: isn't the `e.to_string()` more readable to you?

https://doc.rust-lang.org/std/fmt/#usage
> To convert a single value to a string, use the to_string method.

core/market/decentralized/src/matcher/matcher.rs
```diff
@@ -264,8 +274,8 @@ async fn on_offer_received(db: DbExecutor, msg: OfferReceived) -> Result<Propaga
         Result::<_, MatcherError>::Ok(propagate)
     }
     .await
-    .or_else(|e| {
-        let reason = StopPropagateReason::Error(format!("{}", e));
+    .or_else(|error| {
```
tworec, line 277: why `e` is not enough? Lets be consistent across our module, 

All below stats are with three usages from here

```
~/git/yagna market-decentralized/query-events (04ecd2c2)$ grep --include '*.rs' -rni '|e|' agent/ core/ exe-unit/ service-bus/ test-utils/ utils/ | wc
      75     336    8157
~/git/yagna market-decentralized/query-events (04ecd2c2)$ grep --include '*.rs' -rni '|error|' agent/ core/ exe-unit/ service-bus/ test-utils/ utils/ | wc
      32     132    3282
```

and

```
~/git/yagna market-decentralized/query-events (04ecd2c2)$ grep --include '*.rs' -rni '|error|' core/market/decentralized/ | wc
       5      18     505
~/git/yagna market-decentralized/query-events (04ecd2c2)$ grep --include '*.rs' -rni '|e|' core/market/decentralized/ | wc
      10      44    1209
```

core/market/decentralized/src/negotiation/notifier.rs
```diff
@@ -0,0 +1,81 @@
...
+        // TODO: How to handle this error?
+        let _ = sender.send(to_send);
+    }
+
+    pub async fn wait_for_event(
+        &self,
+        subscription_id: &SubscriptionId,
+    ) -> Result<(), NotifierError> {
+        let mut receiver = self.sender.subscribe();
+        while let Ok(value) = receiver.recv().await {
```
tworec, line 55: 👍 this is really nice usage of broadcast channel

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -43,7 +54,132 @@ impl RequestorNegotiationEngine {
...
         Ok(())
     }
+
+    pub async fn query_events(
+        &self,
+        subscription_id: &SubscriptionId,
+        timeout: f32,
+        max_events: Option<i32>,
+    ) -> Result<Vec<RequestorEvent>, QueryEventsError> {
+        let mut timeout = Duration::from_millis((1000.0 * timeout) as u64);
```
tworec, line 82:
```suggestion
        let mut timeout = Duration::from_secs_f32(timeout);
```

https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs_f32

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -43,7 +54,132 @@ impl RequestorNegotiationEngine {
...
+        }
+
+        loop {
+            let events = get_events_from_db(&self.db, subscription_id, max_events).await?;
+            if events.len() > 0 {
+                return Ok(events);
+            }
+
+            // Solves panic 'supplied instant is later than self'.
+            if stop_time < Instant::now() {
```
tworec, line 99: it might be clearer stated when you use `while` with this condition instead of `loop`

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -43,7 +54,132 @@ impl RequestorNegotiationEngine {
...
+        .take_requestor_events(subscription_id, max_events)
+        .await?;
+
+    // Map model events to client RequestorEvent.
+    let results = futures::stream::iter(events)
+        .then(|event| event.into_client_requestor_event(&db))
+        .collect::<Vec<Result<RequestorEvent, EventError>>>()
+        .await;
+
+    // Filter errors. Can we do something better with errors, than logging them?
```
tworec, line 140: I'd leave logging for MVP.

btw. can you join those two iterations? It will probably shorter and much efficient bc of no need to collect in between

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -43,7 +54,132 @@ impl RequestorNegotiationEngine {
...
+    while let Some(proposal) = proposal_receiver.recv().await {
+        let db = db.clone();
+        let notifier = notifier.clone();
+        match async move {
+            log::info!("Received proposal from matcher. Adding to events queue.");
+
+            // Add proposal to database together with Negotiation record.
+            let proposal = db
+                .as_dao::<ProposalDao>()
+                .new_initial_proposal(proposal.demand, proposal.offer)
```
tworec, line 166: If we have DraftProposal, why should we split it here and not use it as argument type here?

core/market/decentralized/src/rest_api.rs
```diff
@@ -54,20 +54,20 @@ pub struct QueryTimeoutCommandIndex {
 
 #[derive(Deserialize, Debug)]
 pub struct QueryTimeoutMaxEvents {
-    /// number of milliseconds to wait
+    /// number of seconds to wait
```
tworec, line 57: 👍

core/market/decentralized/tests/utils/mock_node.rs
```diff
@@ -89,7 +90,7 @@ impl MarketsNetwork {
         offer_received: impl CallbackHandler<OfferReceived>,
         offer_unsubscribed: impl CallbackHandler<OfferUnsubscribed>,
         retrieve_offers: impl CallbackHandler<RetrieveOffers>,
-    ) -> Result<Self> {
+    ) -> Result<Self, anyhow::Error> {
```
tworec, line 93: this does not add anything bc Result is imported from anyhow already

core/market/decentralized/tests/utils/mock_offer.rs
```diff
@@ -16,9 +26,56 @@ pub fn example_offer() -> Offer {
...
         "golem": {
+            "golem.node.debug.subnet": "blaa".to_string(),
             "node.id.name": "itstest".to_string(),
             "srv.comp.wasm.task_package": "test-package".to_string(),
         },
     });
     Demand::new(properties, "(golem.node.debug.subnet=blaa)".to_string())
 }
+
+impl MarketNode {
```
tworec, line 37: is it intentionally here and not is `mock_node.rs`?

core/market/decentralized/tests/test_initial_proposal.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        let node1 = network.get_node("Node-1");
+        let market1: Arc<MarketService> = network.get_market("Node-1");
+        let identity1 = network.get_default_id("Node-1");
+
+        let subscription_id = market1
+            .subscribe_demand(&example_demand(), &identity1)
+            .await?;
+
+        let market1: Arc<MarketService> = network.get_market("Node-1");
```
tworec, line 79: it is already defined

core/market/decentralized/tests/test_initial_proposal.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        // Inject proposal before timeout will elapse. We expect that Proposal
+        // event will be generated and query events will return it.
+        tokio::time::delay_for(Duration::from_millis(500)).await;
+        node1
+            .inject_proposal_for_demand(&example_offer(), &demand_id)
+            .await?;
+
+        // Protect from eternal waiting.
+        tokio::time::timeout(Duration::from_millis(1100), query_handle).await???;
```
tworec, line 101: 👍

core/market/decentralized/tests/test_initial_proposal.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        let events = market1
+            .requestor_engine
+            .query_events(&demand_id2, 0.0, Some(5))
+            .await?;
+        assert_eq!(events.len(), 1);
+        Ok(())
+    }
+
+    /// Run to query events in the same time.
```
tworec, line 262:
```suggestion
    /// Run two query events in the same time.
```

core/market/decentralized/tests/test_initial_proposal.rs
```diff
@@ -0,0 +1,335 @@
...
+        let node1 = network.get_node("Node-1");
+        let market1: Arc<MarketService> = network.get_market("Node-1");
+        let identity1 = network.get_default_id("Node-1");
+
+        let subscription_id = market1
+            .subscribe_demand(&example_demand(), &identity1)
+            .await?;
+
+        let demand_id1 = subscription_id.clone();
+        let demand_id2 = subscription_id.clone();
```
tworec, line 281: id2 leads to confision when reading. It is better to define it after first query again as id1

core/market/decentralized/src/market.rs
nieznanysprawiciel: Negotiatior is someone who negotiates. It sounds like agent job...

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: Removed after merging tables into one

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: Tabs

core/market/decentralized/migrations/2020-05-15-093513_init-db/up.sql
nieznanysprawiciel: Table removed

core/market/decentralized/src/db/models/proposal.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn new(demand: &ModelDemand, offer: &ModelOffer, role: OwnerType) -> Negotiation {
+        let subscription_id = match role {
+            OwnerType::Provider => offer.id.clone(),
+            OwnerType::Requestor => demand.id.clone(),
+        };
+
+        let identity_id = match role {
+            OwnerType::Provider => offer.node_id.clone(),
+            OwnerType::Requestor => demand.node_id.clone(),
+        };
```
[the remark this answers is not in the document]
nieznanysprawiciel: For this I would need to change fields in Negotiation and add Owner type to db. We agreed to keep Negotiation table unchanged.

core/market/decentralized/src/db/models/proposal.rs
```diff
@@ -0,0 +1,207 @@
...
+/// Proposal is generated on Requestor side, when matching Offer and Demand is found.
+/// Note that initial proposal for Requestor contains properties and constraints
+/// from matched Offer. The same applies to initial Proposal for Provider, which contains
+/// Demand properties and constraints.
+///
+/// Proposal id to, be unique, must be generated from Provider and Requestor
+/// subscription ids and creation timestamp.
+#[derive(Clone, Debug, Identifiable, Insertable, Queryable)]
+#[table_name = "market_proposal"]
+pub struct DbProposal {
```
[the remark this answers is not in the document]
nieznanysprawiciel: It turned out that it requires more changes than expected.
I reverted changes.

core/market/decentralized/src/market.rs
tworec: True.
So maybe `ProviderBroker` :)
https://en.wikipedia.org/wiki/Negotiation

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -43,7 +54,132 @@ impl RequestorNegotiationEngine {
...
+        max_events: Option<i32>,
+    ) -> Result<Vec<RequestorEvent>, QueryEventsError> {
+        let mut timeout = Duration::from_millis((1000.0 * timeout) as u64);
+        let stop_time = Instant::now() + timeout;
+        let max_events = max_events.unwrap_or(i32::max_value());
+
+        if max_events < 0 {
+            Err(QueryEventsError::InvalidMaxEvents(max_events))?
+        } else if max_events == 0 {
+            return Ok(vec![]);
```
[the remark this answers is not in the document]
nieznanysprawiciel: https://github.com/golemfactory/ya-client/issues/28

core/market/decentralized/src/market.rs
nieznanysprawiciel: Ok :D

core/market/decentralized/src/negotiation/requestor.rs
nieznanysprawiciel: Thanks to this I found potential bug. Conversion to u64 could do something strange in case of negative timeouts

core/market/decentralized/src/negotiation/requestor.rs
nieznanysprawiciel: I would prefer to do it this way. But it is required to check events in database at least once. If this would be loop condition, we could not enter loop at all, so I would have to duplicate checking db before loop. And it doesn't look so good as we would expect

core/market/decentralized/src/negotiation/requestor.rs
nieznanysprawiciel: I could if you propose something. Note where is the problem. First iteration makes async call. Second doesn't.
First uses futures iterator, second uses normal iterator on collection.

core/market/decentralized/src/negotiation/requestor.rs
nieznanysprawiciel: This way Dao uses only model types not strange types from outside.

core/market/decentralized/tests/utils/mock_offer.rs
nieznanysprawiciel: Yes, these functions are loosely related to test framework. They are rather functions with rank of example_offer or demand.

core/market/decentralized/src/negotiation/errors.rs
```diff
@@ -1,7 +1,39 @@
...
+
+impl From<TakeEventsError> for QueryEventsError {
+    fn from(e: TakeEventsError) -> Self {
+        match e {
+            TakeEventsError::SubscriptionExpired(id) => QueryEventsError::SubscriptionExpired(id),
+            TakeEventsError::SubscriptionNotFound(id) => QueryEventsError::Unsubscribed(id),
+            _ => QueryEventsError::FailedGetEvents(e),
+        }
+    }
+}
```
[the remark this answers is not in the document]
nieznanysprawiciel: QueryEventsError is api error and should be easily matched against possible errors. TakeEventsError says nothing to external user, but these errors are meaningful and don't need nested matches.

core/market/decentralized/src/db/dao/demand.rs
```diff
@@ -64,3 +72,21 @@ impl<'c> DemandDao<'c> {
         .await
     }
 }
+
+pub(super) fn demand_status(
```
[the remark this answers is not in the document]
nieznanysprawiciel: We could consider this in future, but I don't want do another refactor now, that touches everything

core/market/decentralized/src/db/models/events.rs
nieznanysprawiciel: We avoided many things. When you use already created types, you don't have to do any matching. This is only creation function. I could split it into 2 separate functions with different names and you would have any match, but i didn't want to duplicate code.

core/market/decentralized/src/db/models/events.rs
nieznanysprawiciel: `EventError::ProposalNotFound` doesn't belong there for sure. If you get proposal, dao doesn't know, if not existing proposal is error or not. Only external layer knows this and in this example, this place is external layer.
FailedGetProposal probably could have separate type, but it would only wrap DbError, so it doesn't make sense

core/market/decentralized/src/db/dao/events.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    pub async fn take_requestor_events(
+        &self,
+        subscription_id: &SubscriptionId,
+        max_events: i32,
+    ) -> Result<Vec<MarketEvent>, TakeEventsError> {
+        let subscription_id = subscription_id.clone();
+        Ok(do_with_transaction(self.pool, move |conn| {
+            match demand_status(conn, &subscription_id)? {
```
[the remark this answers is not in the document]
nieznanysprawiciel: We need to do it in single transaction.
Why are you worried about mixing Dao?
That's like saying that you are worried SQL queries with multiple tables.
But of course I don't like this place, because this way Broker needs Demand existence in database, what is implicit dependency :(

core/market/decentralized/src/market.rs
```diff
@@ -43,23 +45,22 @@ pub enum MarketInitError {
 /// Structure connecting all market objects.
 pub struct MarketService {
     pub matcher: Matcher,
-    pub provider_negotiation_engine: Arc<ProviderNegotiationEngine>,
-    pub requestor_negotiation_engine: Arc<RequestorNegotiationEngine>,
+    pub provider_engine: Arc<ProviderBroker>,
```
tworec, line 48: yeah! thx! 🥳

core/market/decentralized/src/matcher.rs
```diff
@@ -17,22 +17,27 @@ mod error;
 mod resolver;
 mod store;
 
+/// Stores proposal generated from resolver.
+#[derive(Debug)]
+pub struct DraftProposal {
```
tworec, line 22: I will change this into `RawProposal` in my PR #333 
because `Draft` in the context of `Proposal` is misleading. Resulting `Proposal` will be in the `Initial` state

nieznanysprawiciel: Ok

Issue #313, opened 2020-06-03
Use `constraints!` macro in both agents


tworec:
use [`constraints!`](https://github.com/golemfactory/yagna/pull/251)
- [x] in [provider agent](https://github.com/golemfactory/yagna/blob/c15bfa4c4a9f3deefa8832a281931b73ad756eac/agent/provider/src/provider_agent.rs#L116-L129)
- [x] in [mock requestor agent](https://github.com/golemfactory/yagna/blob/c15bfa4c4a9f3deefa8832a281931b73ad756eac/agent/requestor/src/market.rs#L45-L51)

AC:
- get rid of `format!`
- use only `constarints!`

PR #347, opened 2020-06-15
Use constraints from agreement utils

jiivan:
fixes: #313


agent/requestor/src/market.rs
```diff
@@ -28,29 +32,31 @@ pub(crate) fn build_demand(
...
-        constraints: format!(
-            "(&
-            (golem.inf.mem.gib>0.5)
-            (golem.inf.storage.gib>1)
-            (golem.com.pricing.model=linear)
-            {}
-        )",
-            subnet_constraint
-        ),
+        constraints: subnet_constraint,
```
nieznanysprawiciel, line 59: Just a note: I would do common_constraint.to_string() here, instead of in line 54. It would leave constraint object open for modification and adding new constraints.

agent/requestor/src/market.rs
```diff
@@ -28,29 +32,31 @@ pub(crate) fn build_demand(
...
     let subnet_constraint = match subnet {
         Some(subnet) => {
             log::info!("Using subnet: {}", subnet);
             properties.as_object_mut().unwrap().insert(
                 "golem.node.debug.subnet".to_string(),
                 serde_json::Value::String(subnet.clone()),
             );
-            format!("(golem.node.debug.subnet={})", subnet.clone())
+            common_constraint.and(Constraints::new_single(
+                ConstraintKey::new("golem.node.debug.subnet").equal_to(ConstraintValue::new(serde_json::Value::String(subnet.clone())))
```
nieznanysprawiciel, line 51: Why do you create ConstraintValue from serde_json::Value::String? Does it work?
Note that when we add subnet to properties (line 48), we must wrap it in serde_json::Value, because `properties` is serde structure. Here subnet can be used directly.

agent/requestor/src/market.rs
```diff
@@ -28,29 +32,31 @@ pub(crate) fn build_demand(
...
     });
 
+    let common_constraint = Constraints::new_clause(
+        And,
+        vec![
+            ConstraintKey::new("golem.inf.mem.gib").greater_than(ConstraintValue::new(0.5)),
+            ConstraintKey::new("golem.inf.storage.gib").greater_than(ConstraintValue::new(1)),
+            ConstraintKey::new("golem.com.pricing.model").equal_to(ConstraintValue::new("linear")),
+        ],
+    );
```
nieznanysprawiciel, line 42: This could be written using `constraints!` macro. Probably would be much more readable.

agent/provider/src/provider_agent.rs
```diff
@@ -118,13 +122,16 @@ impl ProviderAgent {
...
-            None => Ok(format!("(golem.srv.comp.expiration>0)")),
+            Some(subnet) => Ok(Constraints::new_clause(
+                And,
+                vec![
+                    ConstraintKey::new("golem.node.debug.subnet").equal_to(ConstraintValue::new(subnet)),
+                    expiration_constraint,
+                ],
+            ).to_string()),
+            None => Ok(Constraints::new_single(expiration_constraint).to_string()),
         }
```
mfranciszkiewicz, line 135:
```rust
        let mut constr = constraints!["golem.srv.comp.expiration" > 0,];
        if let Some(subnet) = self.node_info.subnet.clone() {
            constr = constr.and(constraints!["golem.node.debug.subnet" == subnet,]);
        }
        Ok(constr.to_string())
```

agent/requestor/src/market.rs
jiivan:
```
   Compiling ya-requestor v0.1.0 (/home/dariusz/repo/yagna/agent/requestor)
error[E0308]: mismatched types
  --> agent/requestor/src/market.rs:41:13
   |
41 |             subnet.clone(),
   |             ^^^^^^^^^^^^^^
   |             |
   |             expected enum `serde_json::value::Value`, found struct `std::string::String`
   |             help: try using a variant of the expected enum: `serde_json::value::Value::String(subnet.clone())`
```

nieznanysprawiciel: Hmm, maybe `ConstraintValue::from(subnet.clone())` could help?
New should work for all types that implement `Into<serde_json::Value>`, so I would expect it to work even with new. @filipgolem

filipgolem: Does `cnts = cnts.and(constraints!["golem.node.debug.subnet" == subnet.clone(),]);` work?

jiivan: Yes, it does. Thanks

agent/provider/src/provider_agent.rs
```diff
@@ -98,14 +99,14 @@ impl ProviderAgent {
             })?;
 
             // Create simple offer on market.
-            let constraints = self.build_constraints()?;
+            let cnts = self.build_constraints()?;
```
tworec, line 102: nitpick: `cnts` resembles rather `counts` than `constraints`

agent/provider/src/provider_agent.rs
```diff
@@ -98,14 +99,14 @@ impl ProviderAgent {
...
             let create_offer_message = CreateOffer {
                 preset,
                 offer_definition: OfferDefinition {
                     node_info: self.node_info.clone(),
                     service: Self::create_service_info(&exeunit_desc),
                     com_info,
-                    constraints,
+                    constraints: cnts,
```
tworec, line 109: @jiivan it was more readable originally, don't you think?

PR #349, opened 2020-06-17
HTTP Basic Auth in transfer URLs

mfranciszkiewicz:
Support HTTP basic auth in transfer URLs

Note: the `transfer` example was not functioning properly due to recent changes

### DIFF
--- a/utils/transfer/src/http.rs
+++ b/utils/transfer/src/http.rs
@@ -10,6 +10,36 @@ use futures::{SinkExt, StreamExt, TryStreamExt};
 use std::thread;
 use url::Url;
 
+enum HttpAuth<'s> {
+    None,
+    Basic {
+        username: &'s str,
+        password: Option<&'s str>,
+    },
+}
+
+impl<'s> From<&'s Url> for HttpAuth<'s> {
+    fn from(url: &'s Url) -> Self {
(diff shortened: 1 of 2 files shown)


utils/transfer/src/http.rs
```diff
@@ -10,6 +10,36 @@ use futures::{SinkExt, StreamExt, TryStreamExt};
...
+        }
+    }
+}
+
+fn request(method: Method, url: Url) -> awc::ClientRequest {
+    let builder = awc::ClientBuilder::new();
+    let client = match HttpAuth::from(&url) {
+        HttpAuth::None => builder.finish(),
+        HttpAuth::Basic { username, password } => builder.basic_auth(username, password).finish(),
+    };
```
etam, line 39:
```suggestion
    if let HttpAuth::Basic { username, password } = HttpAuth::from(&url) {
        builder.basic_auth(username, password);
    }
    let client = builder.finish();
```
I think it would be more readable, but I'm not insisting.

mfranciszkiewicz: Unfortunately `basic_auth` consumes and returns `self`, so it cannot be implemented this way. Also, `HttpAuth` will probably soon be extended with new variants

PR #375, opened 2020-06-26
Negotiation API structs function signatures

nieznanysprawiciel:


core/market/decentralized/src/protocol/negotiation/provider.rs
```diff
@@ -0,0 +1,238 @@
...
+    ) -> Result<(), AgreementError> {
+        log::debug!(
+            "Negotiation API: Agreement [{}] cancelled by [{}].",
+            &msg.agreement_id,
+            &caller
+        );
+        self.inner.agreement_cancelled.call(caller, msg).await
+    }
+
+    pub async fn bind_gsb(
```
tworec, line 187: pls use `ServiceBinder` for this job

core/market/decentralized/src/protocol/negotiation/requestor.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// Sent to provider, when Requestor will call cancel Agreement,
+    /// while waiting for approval.
+    /// TODO: Use model Agreement structu in api.
+    pub async fn cancel_agreement(
+        &self,
+        id: NodeId,
+        agreement_id: &str,
+        owner: NodeId,
+    ) -> Result<(), AgreementError> {
+        let msg = AgreementRejected {
```
tworec, line 132: I'd introduce separate message for this eg. `AgreementCanceled`

core/market/decentralized/src/protocol/negotiation/provider.rs
nieznanysprawiciel: What benefit do you see of using it?
Because I can't see any benefit. I will be just forced to get DbExecutor in parameter, that I don't use. But we will have the same amount of code.

nieznanysprawiciel: After discussion I used `ServiceBinder`. Check new version.

PR #403, opened 2020-07-08
Adjust payment-dummy-driver to new api

maaktweluit:
work in progress...

- moved `payment-driver/dummy.rs` to its own crate: `payment-driver/dummy/lib.rs`
- removed all GNT-driver related files
- update feature toggle to be on the import of `PaymentDriverService`
- Code is compiling, not working


core/payment-driver/dummy/src/service.rs
```diff
@@ -1,28 +1,34 @@
...
 use ya_persistence::executor::DbExecutor;
 use ya_service_bus::{typed as bus, RpcEndpoint};
 
+
+const BUS_ID_POSTFIX: &'static str = "dummy";
+
+
 pub fn bind_service(db: &DbExecutor, processor: PaymentDriverProcessor) {
     log::debug!("Binding payment driver service to service bus");
+    let bus_id: &str = &(BUS_ID_PREFIX.to_owned() + BUS_ID_POSTFIX);
```
maaktweluit, line 13: why are strings in rust so hard..

would like: ```
const BUS_ID: &'static str = BUS_ID_PREFIX + "dummy";
```

File: core/payment-driver/dummy/src/service.rs [hunk 2]
```diff
@@ -1,28 +1,34 @@
...
 use bigdecimal::BigDecimal;
 use ya_core_model::driver::*;
 use ya_persistence::executor::DbExecutor;
 use ya_service_bus::{typed as bus, RpcEndpoint};
 
+
+const BUS_ID_POSTFIX: &'static str = "dummy";
+
+
 pub fn bind_service(db: &DbExecutor, processor: PaymentDriverProcessor) {
```
maaktweluit, line 11: preferably we move this back to `payment-driver` and pass the full bus_id

core/payment-driver/dummy/src/lib.rs
```diff
@@ -54,65 +57,40 @@ impl PaymentDriver for DummyDriver {
...
         let payments = self.payments.clone();
         let details = PaymentDetails {
             recipient: recipient.to_string(),
             sender: sender.to_string(),
-            amount: amount.base_currency_amount,
+            amount: amount,
             date: Some(Utc::now()),
         };
-        let invoice_id = invoice_id.to_string();
+        // FIXME: generate payment_order_id?
```
maaktweluit, line 79: 2 questions on payment_order_id:
- Where should it be generated? just a new UUID?
- Should it be part of `PaymentDetails`?

core/payment-driver/dummy/src/lib.rs
```diff
@@ -54,65 +57,40 @@ impl PaymentDriver for DummyDriver {
...
+        let payment_order_id = "123".to_string();
 
         Box::pin(async move {
-            match payments.lock().await.entry(invoice_id.clone()) {
-                Entry::Occupied(_) => Err(PaymentDriverError::PaymentAlreadyScheduled(invoice_id)),
+            match payments.lock().await.entry(payment_order_id.clone()) {
+                // FIXME: need other properties to check if Payment is already Scheduled
```
maaktweluit, line 84: Depending on the question above, how can we ensure payments are only scheduled once? Or is this no longer the driver repsonsibility

core/payment-driver/dummy/src/lib.rs
```diff
@@ -126,11 +104,21 @@ impl PaymentDriver for DummyDriver {
...
-            currency: Currency::Gnt,
-            amount: utils::str_to_big_dec(&amount).unwrap(),
-        })))
+        Box::pin(future::ready(Ok(utils::str_to_big_dec(&amount).unwrap())))
+    }
+}
+
+pub struct PaymentDriverService;
+
+impl PaymentDriverService {
```
maaktweluit, line 115: could be moved back to `payment-driver` if we pass the ::new() function to use

core/serv/src/main.rs
```diff
@@ -24,6 +24,10 @@ use ya_activity::service::Activity as ActivityService;
 use ya_identity::service::Identity as IdentityService;
 use ya_net::Net as NetService;
 use ya_payment::PaymentService;
+
+#[cfg(feature = "dummy-driver")]
```
maaktweluit, line 28: This approach makes it only possible to import one driver at the time, think we need to fix that enum to import both.

core/payment-driver/dummy/src/service.rs
Wiezzel: You can use `lazy_static` to have it as `&'static str`.

core/payment-driver/dummy/src/lib.rs
Wiezzel: Just a new UUID. Even a counter incremented with every new order should do.

Nope, it doesn't need to go to payment details.

core/payment-driver/dummy/src/lib.rs
Wiezzel: Not the driver's responsibility. Twice ordered, twice served.

core/serv/src/main.rs
Wiezzel: I'm working on it but it's difficult.

core/payment-driver/dummy/src/lib.rs
maaktweluit: updated it to be an incremental counter

core/payment-driver/src/dao/payment.rs
```diff
@@ -48,49 +48,49 @@ impl<'c> PaymentDao<'c> {
...
-                *confirmation = PaymentConfirmation {
-                    confirmation: tx_hash,
-                };
-            }
-
-            Ok(Some(status))
-        })
-        .await
-    }
+    // pub async fn get_payment_status(&self, invoice_id: String) -> DbResult<Option<PaymentStatus>> {
```
maaktweluit, line 51: @bidzyyys these break my compilation, please bring them back ( or delete them ) when updating the GNT driver

core/serv/src/main.rs
Wiezzel: Not really worth the pain, let's leave it like this.

core/payment-driver/src/dao/payment.rs
maaktweluit: yes, when these functions are commented out it compiles :)

Thank you <3

PR #408, opened 2020-07-09
More verbose log message when path not found in container

kubkon:
This would have saved me some debugging time ;-)


exe-unit/src/service/transfer.rs
```diff
@@ -100,7 +100,7 @@ impl ContainerTransferProvider {
             }
             Ok(vol_base.join(path))
         } else {
-            log::warn!("not found!!");
+            log::warn!("path {} not found in container", container_path);
```
etam, line 103:
```suggestion
            log::warn!("path not found in container: {}", container_path);
```
it makes it less obscure if path is really long.

kubkon: Makes sense, but in that case, IMHO it also makes sense to tweak the error message below.

PR #419, opened 2020-07-14
various impromevents for exe unit runtime server

etam:
required by https://github.com/golemfactory/ya-runtime-vm/pull/39

### DIFF
--- a/exe-unit/runtime-api/examples/runtime-server-mock.rs
+++ b/exe-unit/runtime-api/examples/runtime-server-mock.rs
@@ -3,9 +3,7 @@ use std::env;
 
 use std::time::Duration;
 use tokio;
-use ya_runtime_api::server;
-use ya_runtime_api::server::AsyncResponse;
-use ya_runtime_api::server::RuntimeService;
+use ya_runtime_api::server::{self, AsyncResponse, RuntimeEvent, RuntimeService};
 
 struct RuntimeMock;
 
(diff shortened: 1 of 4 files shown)


exe-unit/runtime-api/examples/runtime-server-mock.rs
```diff
@@ -1,15 +1,23 @@
...
 
-struct RuntimeMock;
+// server
 
-struct EventMock;
+struct RuntimeMock<E>
+where
+    E: RuntimeEvent,
+{
+    event_emiter: E,
```
mfranciszkiewicz, line 17:
```suggestion
    event_emitter: E,
```

exe-unit/runtime-api/examples/runtime-server-mock.rs
```diff
@@ -41,10 +56,41 @@ impl server::RuntimeService for RuntimeMock {
...
+    RuntimeMock { event_emiter }
+}
+
+// client
+
+struct EventsData {
+    last_status: ProcessStatus,
+}
+
+struct EventMock(Arc<Mutex<EventsData>>);
```
mfranciszkiewicz, line 69: `EventsData` instead of `ProcessStatus` seems to be an extra indirection with no added value

etam: I need a type that implements both RuntimeEvent trait and Clone, where each clone shares the same underlying data.

etam: done

exe-unit/runtime-api/examples/runtime-server-mock.rs
etam: done

Issue #383, opened 2020-06-29
Agreement - wait for approval


nieznanysprawiciel:
- [ ] Implement Requestor POST `agreements/{agreement_id}/wait` endpoint
    - [x] Check state of Agreement in database immediately in case Provider already approved/rejected it (How to avoid race conditions here)
    - [x] Wait for approval notifications
    - [ ] Wait for rejection notifications
    - [x] Wait for timeout
        - [ ] Probably we should send cancel agreement on timeout
    - [ ] Wait for cancel Agreement (Requestor can cancel Agreement as long it wasn't approved)

Remaining points were moved to: #432

PR #420, opened 2020-07-14
Agreement - Wait for approval

nieznanysprawiciel:
resolves: #383

### DIFF
--- a/core/market/decentralized/src/db/dao/agreement.rs
+++ b/core/market/decentralized/src/db/dao/agreement.rs
@@ -19,8 +19,12 @@ impl<'a> AsDao<'a> for AgreementDao<'a> {
 
 #[derive(thiserror::Error, Debug)]
 pub enum StateError {
-    #[error("Can't update Agreement [{0}] state from {1} to {2}.")]
-    InvalidTransition(AgreementId, AgreementState, AgreementState),
+    #[error("Can't update Agreement [{id}] state from {from} to {to}.")]
+    InvalidTransition {
+        id: AgreementId,
+        from: AgreementState,
+        to: AgreementState,
+    },
     #[error("Failed to update state. Error: {0}")]
     DbError(DbError),
 }
@@ -74,11 +78,11 @@ impl<'c> AgreementDao<'c> {
             let agreement: Agreement = dsl::market_agreement.filter(dsl::id.eq(&id)).first(conn)?;
 
             if agreement.state != AgreementState::Pending {
-                Err(StateError::InvalidTransition(
-                    id.clone(),
-                    agreement.state,
-                    AgreementState::Approved,
-                ))?
+                Err(StateError::InvalidTransition {
+                    id: id.clone(),
+                    from: agreement.state,
+                    to: AgreementState::Approved,
+                })?
             }
 
             diesel::update(dsl::market_agreement.filter(dsl::id.eq(&id)))
--- a/core/market/decentralized/src/db/dao/agreement.rs
+++ b/core/market/decentralized/src/db/dao/agreement.rs
@@ -1,11 +1,11 @@
+use chrono::NaiveDateTime;
 use diesel::prelude::*;
 
 use ya_persistence::executor::{do_with_transaction, AsDao, ConnType, PoolType};
 
 use crate::db::model::{Agreement, AgreementId, AgreementState};
 use crate::db::schema::market_agreement::dsl;
-use crate::db::DbResult;
-use chrono::NaiveDateTime;
+use crate::db::{DbError, DbResult};
 
 pub struct AgreementDao<'c> {
     pool: &'c PoolType,
@@ -15,6 +15,14 @@ impl<'a> AsDao<'a> for AgreementDao<'a> {
     fn as_dao(pool: &'a PoolType) -> Self {
         Self { pool }
     }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum StateError {
+    #[error("Can't update Agreement [{0}] state from {1} to {2}.")]
+    InvalidTransition(AgreementId, AgreementState, AgreementState),
+    #[error("Failed to update state. Error: {0}")]
+    DbError(DbError),
 }
 
 impl<'c> AgreementDao<'c> {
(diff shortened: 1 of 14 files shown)


core/market/decentralized/src/negotiation/notifier.rs
```diff
@@ -1,80 +1,108 @@
...
         let _ = sender.send(to_send);
     }
 
-    pub async fn wait_for_event(
-        &self,
-        subscription_id: &SubscriptionId,
-    ) -> Result<(), NotifierError> {
-        let mut receiver = self.sender.subscribe();
-        while let Ok(value) = receiver.recv().await {
+    pub fn listen(&self, subscription_id: &Type) -> EventNotifierListener<Type> {
```
tworec, line 72: 👍

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -47,20 +62,25 @@ impl RequestorBroker {
...
         let api = NegotiationApi::new(
             move |caller: String, msg: ProposalReceived| {
                 broker1
                     .clone()
                     .on_proposal_received(caller, msg, OwnerType::Requestor)
             },
             move |_caller: String, _msg: ProposalRejected| async move { unimplemented!() },
-            move |_caller: String, _msg: AgreementApproved| async move { unimplemented!() },
+            move |caller: String, msg: AgreementApproved| {
+                on_agreement_approved(broker2.clone(), caller, msg, agreement_notifier2.clone())
```
tworec, line 74: nitpicking: should be second arg

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -39,6 +53,7 @@ impl RequestorBroker {
         store: SubscriptionStore,
         proposal_receiver: UnboundedReceiver<RawProposal>,
     ) -> Result<RequestorBroker, NegotiationInitError> {
+        let agreement_notifier = EventNotifier::new();
```
tworec, line 55: isn't that common? I mean to be placed in CommonBroker...

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -194,15 +214,113 @@ impl RequestorBroker {
...
     }
 
+    pub async fn wait_for_approval(
+        &self,
+        id: &AgreementId,
+        timeout: f32,
+    ) -> Result<ApprovalStatus, WaitForApprovalError> {
+        // TODO: Check if we are owner of Proposal
+        // TODO: What to do with 2 simultaneous calls to wait_for_approval??
+        //  should we reject one? And if so, how to discover, that to calls were made?
```
tworec, line 237:
```suggestion
        //  should we reject one? And if so, how to discover, that two calls were made?
```

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -194,15 +214,113 @@ impl RequestorBroker {
...
+        // TODO: Do we have race conditions here, between checking Agreement state and
+        //  waiting for notifications? Event could come, before we start waiting and
+        //  it will be lost.
```
tworec, line 240: As far as I understand we are using event only to be sure that the `Agreement.state` is already properly set in DB (ie. it is not `Pending`), we are safe here, because the `loop` below will return immediately after querying DB.

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -194,15 +214,113 @@ impl RequestorBroker {
...
+        let timeout = Duration::from_secs_f32(timeout.max(0.0));
+        let mut notifier = self.agreement_notifier.listen(id);
+
+        // Loop will wait for events notifications only one time. It doesn't have to be loop at all,
+        // but it spares us doubled getting agreement and mapping statuses to return results.
+        // So I think this simplification is worth confusion, that it cause.
```
tworec, line 246: Your comment will clear doubts for future readers I belive.
So 👍  let it be as it is.

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -194,15 +214,113 @@ impl RequestorBroker {
...
+    caller: String,
+    msg: AgreementApproved,
+    notifier: EventNotifier<AgreementId>,
+) -> Result<(), ApproveAgreementError> {
+    let agreement = broker
+        .db
+        .as_dao::<AgreementDao>()
+        .get(&msg.agreement_id)
+        .await
+        .map_err(|_e| RemoteAgreementError::NotFound(msg.agreement_id.clone()))?;
```
tworec: Non blocking comment: It is not precisely `NotFound` because there could be some DB or other issues

core/market/decentralized/src/negotiation/requestor.rs
```diff
@@ -194,15 +214,113 @@ impl RequestorBroker {
...
+                .as_dao::<AgreementDao>()
+                .get(id)
+                .await
+                .map_err(|e| WaitForApprovalError::FailedGetFromDb(id.clone(), e))?;
+
+            match agreement.state {
+                AgreementState::Approved => return Ok(ApprovalStatus::Ok),
+                AgreementState::Rejected => return Ok(ApprovalStatus::Rejected),
+                AgreementState::Cancelled => return Ok(ApprovalStatus::Cancelled),
+                AgreementState::Expired => Err(WaitForApprovalError::AgreementExpired(id.clone()))?,
```
tworec, line 261: `return` would be much more readable here than `?` because we have three `return`s above.

core/market/decentralized/src/db/dao/agreement.rs
```diff
@@ -16,6 +16,14 @@ impl<'a> AsDao<'a> for AgreementDao<'a> {
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum StateError {
+    #[error("Can't update Agreement [{0}] state from {1} to {2}.")]
+    InvalidTransition(AgreementId, AgreementState, AgreementState),
```
tworec, line 23: we have two of `AgreementState` here, so I'd go for named fields (struct) instead of a tuple.

core/market/decentralized/src/rest_api/error.rs
```diff
@@ -152,3 +152,21 @@ impl ResponseError for AgreementError {
...
+
+impl ResponseError for WaitForApprovalError {
+    fn error_response(&self) -> HttpResponse {
+        let msg = ErrorMessage::new(self.to_string());
+        match self {
+            WaitForApprovalError::NotFound(_) => HttpResponse::NotFound().json(msg),
+            WaitForApprovalError::AgreementExpired(_) => HttpResponse::Gone().json(msg),
+            WaitForApprovalError::AgreementTerminated(_)
+            | WaitForApprovalError::AgreementNotConfirmed(_) => {
+                HttpResponse::BadRequest().json(msg)
```
tworec, line 164: I believe we should return here `409 Conflict`, but it is not compatible with the spec. So spec needs to be updated

PR #427, opened 2020-07-15
Various path-related fixes

mfranciszkiewicz:
Provider:
- create `work` and `cache` sub-directories in a working directory
- use any existing path for hw configuration

Transfer:
- use url-decoded Url::path()

Deploy package path:
- fix deploy image's SHA-3 sum


agent/requestor/src/main.rs
```diff
@@ -15,7 +15,7 @@ mod market;
 mod payment;
 
 const DEFAULT_NODE_NAME: &str = "test1";
-const DEFAULT_TASK_PACKAGE: &str = "hash://sha3:38D951E2BD2408D95D8D5E5068A69C60C8238FA45DB8BC841DC0BD50:http://34.244.4.185:8000/rust-wasi-tutorial.zip";
+const DEFAULT_TASK_PACKAGE: &str = "hash://sha3:D5E31B2EED628572A5898BF8C34447644BFC4B5130CFC1E4F10AEAA1:http://34.244.4.185:8000/rust-wasi-tutorial.zip";
```
tworec, line 18: why do we need an update here?

core/market/resolver/tests/expression_resolve.rs
```diff
@@ -560,7 +560,7 @@ fn resolve_pseudo_function_array_sample_undefined() {
 #[test]
 fn resolve_wildcard_prop_sample_undefined() {
     // this syntax should work, and should return "undefined" for property declared with wildcards.
-    let f = r#"(golem.srv.comp.wasm.task_package=hash://sha3:38D951E2BD2408D95D8D5E5068A69C60C8238FA45DB8BC841DC0BD50:http://34.244.4.185:8000/rust-wasi-tutorial.zip)"#;
+    let f = r#"(golem.srv.comp.wasm.task_package=hash://sha3:D5E31B2EED628572A5898BF8C34447644BFC4B5130CFC1E4F10AEAA1:http://34.244.4.185:8000/rust-wasi-tutorial.zip)"#;
```
tworec, line 563: this is just a mock value, need not be valid, just arbitrary string

core/market/resolver/tests/sample/mod.rs
```diff
@@ -92,7 +92,7 @@ pub static POC_DEMAND_PROPERTIES_JSON: &str = r#"
   "golem.node.debug.subnet": "piotr",
   "golem.node.id.name": "test1",
   "golem.srv.comp.expiration": 1590765503361,
-  "golem.srv.comp.wasm.task_package": "hash://sha3:38D951E2BD2408D95D8D5E5068A69C60C8238FA45DB8BC841DC0BD50:http://34.244.4.185:8000/rust-wasi-tutorial.zip"
+  "golem.srv.comp.wasm.task_package": "hash://sha3:D5E31B2EED628572A5898BF8C34447644BFC4B5130CFC1E4F10AEAA1:http://34.244.4.185:8000/rust-wasi-tutorial.zip"
```
tworec, line 95: same here

agent/requestor/src/main.rs
mfranciszkiewicz: Consistency. We've decided to copy a valid URL with a valid hash in multiple places, now let's keep it up to date

agent/provider/src/execution/task_runner.rs
```diff
@@ -138,14 +138,15 @@ pub struct TaskRunner {
 }
 
 impl TaskRunner {
-    pub fn new(
+    pub fn new<P: AsRef<Path>>(
         client: ActivityProviderApi,
         config: TaskRunnerConfig,
         registry: ExeUnitsRegistry,
+        data_dir: P,
```
etam, line 145: What is the benefit of using `AsRef<Path>`, rather than just `&Path`?

agent/requestor/src/main.rs
tworec: cool

agent/provider/src/execution/task_runner.rs
tworec: it is more generic
You can input both `Path` and `&Path` and `PathBuf` and `&PathBuf` as a input. Namely all that implements `AsRef<Path>`

etam: I understand that there is a trade-off between
```rust
fn foo(path: &Path) { // <- simpler function
    use(path);
}

let path_buf: PathBuf = ...;
foo(path_buf.as_ref());
```
and
```rust
fn foo<P: AsRef<Path>>(path: P) {
    let path = path.as_ref();
    use(path);
}

let path_buf: PathBuf = ...;
foo(path_buf); // <- simpler calling
```

I just think that the complexity function in second example is bigger than complexity of calling it in first example. Also using such templates might make compilation longer and output binary bigger (based on experience from C++ templates).

tworec: pls read my answer carefully ;)
in template version `foo<P: AsRef<Path>>(path: P)` you can also do
```
let p : Path =...
foo(p)
```

and **any** other type that implements AsRef for Path

etam: You can't make `let p: Path = ...`. https://doc.rust-lang.org/std/path/struct.Path.html says
> This is an unsized type, meaning that it must always be used behind a pointer like & or Box. For an owned version of this type, see PathBuf.

But, let's take a `PathBuf` as an example:
```rust
fn foo(path: &Path) {}

let p: PathBuf = ...;
foo(&p);
```
This works just fine.

So can you provide a complete convincing example, where your version has a clear advantage?

mfranciszkiewicz: @etam Please take a look at https://doc.rust-lang.org/std/path/struct.Path.html#trait-implementations

PR #428, opened 2020-07-16
Provider agent: limit hardware resources via CLI or env vars

mfranciszkiewicz:
Introduces 3 new command line arguments for limiting max CPU cores, RAM, storage plus corresponding env variables. Temporarily affects existing hardware profiles, permanently affects newly created hardware profiles.


agent/provider/Cargo.toml
```diff
@@ -28,6 +28,7 @@ env_logger = "0.7"
 futures = "0.3"
 futures-util = "0.3.4"
 humantime = "2.0.0"
+lazy_static = "1.4"
```
etam, line 31: I guess you wanted to use it to cache `max_caps`

agent/provider/src/hardware.rs
```diff
@@ -65,24 +65,45 @@ pub struct Resources {
 }
 
 impl Resources {
-    pub fn new_empty() -> Self {
+    pub fn try_with_config(work_dir: &Path, config: &ProviderConfig) -> Result<Self, Error> {
+        if config.rt_cores.is_some() || config.rt_mem.is_some() || config.rt_storage.is_some() {
+            let max_caps = Self::max_caps(work_dir)?;
```
etam, line 70: Use `default_caps` here. It might be surprising that for example if you specify storage in config, it takes not 0.7 but max of memory (which is unrelated to storage).

mfranciszkiewicz: These caps serve as an upper limit for the currently active hardware profile.

On a second thought, I think that `default_caps` should only be used while creating a default profile, otherwise the user won't be able to go beyond the upper resource limit.

PR #440, opened 2020-07-20
[payment-driver] use payment.db

tworec:
Payment Driver service to use `payment.db`

### DIFF
--- a/core/serv/src/main.rs
+++ b/core/serv/src/main.rs
@@ -149,6 +149,7 @@ impl ServiceContext {
             Self::make_entry::<MarketService>(path, "market")?,
             Self::make_entry::<ActivityService>(path, "activity")?,
             Self::make_entry::<PaymentService>(path, "payment")?,
+            Self::make_entry::<PaymentDriverService>(path, "payment")?,
         ]
         .iter()
         .cloned()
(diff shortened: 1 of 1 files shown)


core/serv/src/main.rs
```diff
@@ -149,6 +149,7 @@ impl ServiceContext {
             Self::make_entry::<MarketService>(path, "market")?,
             Self::make_entry::<ActivityService>(path, "activity")?,
             Self::make_entry::<PaymentService>(path, "payment")?,
+            Self::make_entry::<PaymentDriverService>(path, "payment")?,
```
bidzyyys, line 152: `payment` twice? Why not `payment-driver`? Or even `gnt-driver`? Each db is payment platform specific.

Wiezzel: If the idea is that each service should have its own database, then this should be changed.

tworec: pls change this name to whatever you want. I'm just after splitting it from main `yagna.db`

Wiezzel, line 152:
```suggestion
            Self::make_entry::<PaymentDriverService>(path, "gnt-driver")?,
```

PR #452, opened 2020-07-23
Payment API: Cancelling invoices

Wiezzel:
#### Description

Implemented `/invoice/<invoice_id>/cancel` endpoint. Cancelled invoice is treated as it has never been issued. It cannot be acted upon (sent, accepted, etc.), does not count for total requested money amount. Invoice cancellation triggers an event on requestor side. It is impossible if the invoice has already been accepted.

----
#### Testing

1. Start `payment_api`:
```shell
$ cd ./core/payment/
$ cargo run --example payment_api -- --driver=dummy
```
2. Run `cancel_invoice` example:
```shell
$ cd ./core/payment/
$ cargo run --example cancel_invoice
```

### DIFF
--- a/core/payment/examples/cancel_invoice.rs
+++ b/core/payment/examples/cancel_invoice.rs
@@ -1,0 +1,176 @@
+use bigdecimal::BigDecimal;
+use chrono::Utc;
+use std::time::Duration;
+use ya_client::payment::{PaymentProviderApi, PaymentRequestorApi};
+use ya_client::web::WebClient;
+use ya_client_model::payment::{
+    Acceptance, DocumentStatus, EventType, NewAllocation, NewDebitNote, NewInvoice,
+};
+use ya_core_model::payment::local as pay;
+use ya_service_bus::typed as bus;
+
+async fn assert_requested_amount(
+    payer_addr: &str,
+    payee_addr: &str,
+    payment_platform: &str,
+    amount: &BigDecimal,
+) -> anyhow::Result<()> {
+    let payer_status = bus::service(pay::BUS_ID)
+        .call(pay::GetStatus {
+            platform: payment_platform.to_string(),
+            address: payer_addr.to_string(),
+        })
+        .await??;
+    assert_eq!(&payer_status.outgoing.requested, amount);
+
+    let payee_status = bus::service(pay::BUS_ID)
+        .call(pay::GetStatus {
+            platform: payment_platform.to_string(),
+            address: payee_addr.to_string(),
+        })
+        .await??;
+    assert_eq!(&payee_status.incoming.requested, amount);
+    Ok(())
+}
+
+#[actix_rt::main]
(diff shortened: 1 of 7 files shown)


core/payment/src/api/provider.rs
```diff
@@ -326,8 +328,50 @@ async fn cancel_invoice(
...
+        Ok(None) => return response::not_found(),
+        Err(e) => return response::server_error(&e),
+    };
+
+    match invoice.status {
+        DocumentStatus::Issued => (),
+        DocumentStatus::Received => (),
+        DocumentStatus::Rejected => (),
+        DocumentStatus::Cancelled => return response::ok(Null),
+        DocumentStatus::Accepted | DocumentStatus::Settled | DocumentStatus::Failed => {
```
maaktweluit, line 347: should failed get its own response error message?

core/payment/src/api/provider.rs
```diff
@@ -326,8 +328,50 @@ async fn cancel_invoice(
...
+                })
+                .await??;
+            dao.cancel(invoice_id, node_id).await?;
+            Ok(())
+        }
+        .await
+        {
+            Ok(_) => response::ok(Null),
+            Err(Error::Rpc(RpcMessageError::Cancel(CancelError::Conflict))) => {
+                response::conflict(&"Invoice already accepted by requestor")
```
maaktweluit, line 369: this magic string is repeated 2x, maybe we can make it a const or a response type?

core/payment/src/api/provider.rs
Wiezzel: If the payment failed, invoice must had been accepted beforehand. So the message is technically correct.

core/payment/src/api/provider.rs
Wiezzel: it's not _magic_ in that sense that nowhere is it assumed that these two messages are identical and changing it doesn't break the code in any way. But I can extract it to a variable.

maaktweluit: True dat :)

Think you can leave it as is since its only 2x and in the same file...

core/payment/examples/cancel_invoice.rs
```diff
@@ -0,0 +1,176 @@
...
+        .await??;
+    assert_eq!(&payer_status.outgoing.requested, amount);
+
+    let payee_status = bus::service(pay::BUS_ID)
+        .call(pay::GetStatus {
+            platform: payment_platform.to_string(),
+            address: payee_addr.to_string(),
+        })
+        .await??;
+    assert_eq!(&payee_status.incoming.requested, amount);
```
bidzyyys, line 32: Does it crash the daemon? Do we want this?

Wiezzel: No, it doesn't. That's just test code. The daemon is running in another process.

PR #459, opened 2020-07-28
Payment: initialize accounts from file

Wiezzel:
### Description

On startup Yagna daemon will automatically initialize all payment accounts listed in `accounts.json` file (path to this file can be
changed using `ACCOUNT_LIST` variable). If the file does not exist, it will be automatically created with the default identity listed as receive-only GNT account. Format of the file is specified by an example `accounts-example.json`.

A new command was added to the CLI in order to verify if the initialization is working correctly: `yagna payment accounts`. This
command lists all initialized payment accounts with information about platform, driver, and mode.

----

### Testing instrucitons

1. Start yagna daemon with a clean database.
```shell
$ yagna service run
```
2. Verify if the default payment account is initialized (address should be the same as default identity).
```shell
$ yagna payment accounts
┌────────────┬──────────────────────────────────────────────┬──────────┬────────┬────────┐
│  platform  │  address                                     │  driver  │  send  │  recv  │
├────────────┼──────────────────────────────────────────────┼──────────┼────────┼────────┤
│  GNT       │  0xdde2ccd1566294aa77db314501385ea50f1059c3  │  gnt     │        │  X     │
└────────────┴──────────────────────────────────────────────┴──────────┴────────┴────────┘
```
3. Verify if the `accounts.json` file has been created.
```shell
$ cat accounts.json
[{"driver":"gnt","address":"0xdde2ccd1566294aa77db314501385ea50f1059c3","send":false,"receive":true}]
```
4. Stop the daemon.
5. Edit `accounts.json` and set the default account into send mode.
```shell
$ cat accounts.json
[{"driver":"gnt","address":"0xdde2ccd1566294aa77db314501385ea50f1059c3","send":true,"receive":false}]
```
6. Start the daemon again (could take some time this time because it will be requesting money from faucet).
```shell
$ yagna service run
```
7. Verify if the account is initialized in send mode.
```shell
$ yagna payment accounts
┌────────────┬──────────────────────────────────────────────┬──────────┬────────┬────────┐
│  platform  │  address                                     │  driver  │  send  │  recv  │
├────────────┼──────────────────────────────────────────────┼──────────┼────────┼────────┤
│  GNT       │  0xdde2ccd1566294aa77db314501385ea50f1059c3  │  gnt     │  X     │        │
└────────────┴──────────────────────────────────────────────┴──────────┴────────┴────────┘
```
8. Verify if the funds have been obtained from faucet.
```shell
$ yagna payment status -p gnt 0xdde2ccd1566294aa77db314501385ea50f1059c3
---
amount: "1000"
incoming:
  accepted: "0"
  confirmed: "0"
  requested: "0"
outgoing:
  accepted: "0"
  confirmed: "0"
  requested: "0"
reserved: "0"
```

### DIFF
--- a/core/payment/src/accounts.rs
+++ b/core/payment/src/accounts.rs
@@ -1,0 +1,80 @@
+use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
+use std::env;
+use std::path::Path;
+use tokio::fs;
+use ya_core_model::driver::{driver_bus_id, AccountMode, Init};
+use ya_core_model::identity;
+use ya_service_bus::typed as bus;
+
+pub const DEFAULT_PAYMENT_DRIVER: &str = "gnt";
+
+lazy_static! {
+    pub static ref ACCOUNT_LIST_PATH: String =
+        env::var("ACCOUNT_LIST").unwrap_or("accounts.json".to_string());
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub(crate) struct Account {
+    pub driver: String,
+    pub address: String,
+    pub send: bool,
+    pub receive: bool,
+}
+
+pub(crate) async fn init_account(account: Account) -> anyhow::Result<()> {
+    log::debug!("Initializing payment account {:?}...", account);
+    let mut mode = AccountMode::NONE;
+    mode.set(AccountMode::SEND, account.send);
+    mode.set(AccountMode::RECV, account.receive);
+    bus::service(driver_bus_id(account.driver))
+        .call(Init::new(account.address, mode))
+        .await??;
+    log::debug!("Account initialized.");
+    Ok(())
+}
+
+/// Read payment accounts information from `ACCOUNT_LIST` file and initialize them.
+pub async fn init_accounts() -> anyhow::Result<()> {
+    log::debug!(
+        "Initializing payment accounts from file {} ...",
+        &*ACCOUNT_LIST_PATH
+    );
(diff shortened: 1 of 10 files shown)


core/payment/src/accounts.rs
```diff
@@ -0,0 +1,80 @@
...
+
+pub(crate) async fn init_account(account: Account) -> anyhow::Result<()> {
+    log::debug!("Initializing payment account {:?}...", account);
+    let mut mode = AccountMode::NONE;
+    mode.set(AccountMode::SEND, account.send);
+    mode.set(AccountMode::RECV, account.receive);
+    bus::service(driver_bus_id(account.driver))
+        .call(Init::new(account.address, mode))
+        .await??;
+    log::debug!("Account initialized.");
```
maaktweluit, line 33: i think this should also override / extend the file account json file

Wiezzel: You mean to persist all accounts initialized from CLI?

maaktweluit: yes, these are the scenarios i was thinking about
- Default id got updated, payment init called with new default address ( should override )
- Calling payment init -r on default identiry ( should update the line in json )
- payment init called on other identity ( should extend the json )

i'm just thinking out loud, we can make issues from scenarios that are too large to be picked up in this task

Wiezzel: These look out of the scope for this task. It was just supposed to initialize accounts from file. I threw in creating a file with the default identity as a bonus because it makes the default case easier. We can discuss improvement ideas in the next sprint.

PR #453, opened 2020-07-23
Clean up market DB artifacts

jiivan:
fixes: #377

PR #460, opened 2020-07-28
Market Discovery Protocol with cyclic broadcasting

nieznanysprawiciel:


core/market/decentralized/src/matcher/store.rs
```diff
@@ -218,4 +253,37 @@ impl SubscriptionStore {
...
+                                e
+                            );
+                            None
+                        }
+                    }
+                }
+            })
+            .collect::<Vec<SubscriptionId>>()
+            .await)
+    }
```
nieznanysprawiciel, line 288: Use batch_select. Make set difference.

core/market/decentralized/src/matcher.rs
```diff
@@ -140,60 +181,334 @@ impl Matcher {
...
+    _caller: String,
+    msg: GetOffers,
+) -> Result<Vec<Offer>, DiscoveryRemoteError> {
+    match resolver.store.get_offers_batch(msg.offers).await {
+        Ok(offers) => Ok(offers),
+        Err(e) => {
+            log::error!("Failed to get batch offers. Error: {}", e);
+            // TODO: Propagate error.
+            Ok(vec![])
+        }
```
nieznanysprawiciel, line 288: Return error instead of empty vector

core/market/decentralized/src/matcher.rs
```diff
@@ -140,60 +181,334 @@ impl Matcher {
...
+            let num_ours_offers = our_offers.len();
+            let num_to_broadcast = matcher.config.discovery.num_broadcasted_offers;
+
+            let all_offers = matcher
+                .store
+                .get_offers(None)
+                .await?
+                .into_iter()
+                .map(|offer| offer.id)
+                .collect::<Vec<SubscriptionId>>();
```
nieznanysprawiciel, line 377: Don't query full Offers from database if we only need ids

core/market/decentralized/src/config.rs
```diff
@@ -0,0 +1,48 @@
+use chrono::Duration;
+
+/// TODO: Decide where should this config be loaded from.
+///  We could deserialize it from .json file or use structopt and
+///  configure market through env variables.
+pub struct Config {
```
tworec, line 6: @jiivan pls look at this. We (with @nieznanysprawiciel ) would like to introduce some kind of marriage of this approach and yours from #453

core/market/decentralized/src/matcher/store.rs
nieznanysprawiciel: After rethinking: I can't use batch_select. It doesn't return unsubscribed and expired Offers. So if I remove from `offers` result of batch_select, it won't be what I expect here

core/market/decentralized/src/db/model/subscription_id.rs
```diff
@@ -173,6 +174,25 @@ impl From<SubscriptionParseError> for ErrorMessage {
...
+/// Pretty display vector, as list with all elements in separate row.
+/// Implementation uses Display trait for all elements (not Debug) to make
+/// them more readable.
+pub struct DisplayVec<'a, T>(pub &'a Vec<T>);
+
+impl<'a, T> Display for DisplayVec<'a, T>
+where
+    T: Display,
+{
+    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
```
tworec, line 186: there is https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.format_with which is already used by our deps. so adding this dep will not increase resulting binary size most probably

tworec (commented):
partial reviw

core/market/decentralized/src/matcher.rs
```diff
@@ -140,60 +181,334 @@ impl Matcher {
...
+        .choose_multiple(&mut rand::thread_rng(), num_to_select)
+        .cloned()
+        .collect::<Vec<SubscriptionId>>();
+    random_offers.extend(our_offers);
+    random_offers
+}
+
+fn randomize_interval(mean_interval: std::time::Duration) -> std::time::Duration {
+    let mut rng = rand::thread_rng();
+    (2 * mean_interval).mul_f64(rng.gen::<f64>())
```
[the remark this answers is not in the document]
nieznanysprawiciel: But look. We want to have mean interval in config, because it is human intuitive.
And this is how you compute random time from mean...

core/market/decentralized/src/db/dao/offer.rs
```diff
@@ -53,6 +53,23 @@ impl<'c> OfferDao<'c> {
         .await
     }
 
+    pub async fn batch_select(
```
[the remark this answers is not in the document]
tworec: Or even `Vec<node_id>` with similar semantic...

core/market/decentralized/src/identity.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    async fn list(&self) -> Result<Vec<NodeId>, IdentityError> {
+        Ok(bus::service(identity::BUS_ID)
+            .send(identity::List {})
+            .await
+            .map_err(|e| IdentityError::GsbError(e.to_string()))?
+            .map_err(|e| IdentityError::ListError(e.to_string()))?
+            .iter()
+            .map(|identity_info| identity_info.node_id)
```
tworec, line 49: what about locked identities? We should probably filter them out. 

I'm not sure if default identity can be locked. Probably yes... In such a case we should cease broadcasting for it probably

core/market/decentralized/src/matcher/store.rs
```diff
@@ -92,6 +111,18 @@ impl SubscriptionStore {
             .collect())
     }
 
+    pub async fn get_offers_batch(
+        &self,
+        ids: Vec<SubscriptionId>,
```
tworec, line 116: this should be also more generic if you'll decide to follow my comment on `OffersDao`

core/market/decentralized/src/market.rs
```diff
@@ -114,7 +119,11 @@ impl MarketService {
     }
 
     pub async fn get_offers(&self, id: Option<Identity>) -> Result<Vec<Offer>, MarketError> {
-        Ok(self.matcher.store.get_offers(id).await?)
+        Ok(self
```
tworec, line 122: I'm more and more concerned about why do we need those wrappers here in main market module. They should be covered within rest_api module. If you'll agree than please create issue based on that comment, as this should not be part of this PR

core/market/decentralized/src/matcher.rs
```diff
@@ -83,13 +103,21 @@ impl Matcher {
         offer: &ClientOffer,
         id: &Identity,
     ) -> Result<Offer, MatcherError> {
-        // TODO: Handle broadcast errors. Maybe we should retry if it failed.
         let offer = self.store.create_offer(id, offer).await?;
         self.resolver.receive(&offer);
 
+        log::info!(
+            "Subscribed new Offer: [{}] using identity: {} [{}]",
```
tworec, line 110: 👍 good to have this in logs

core/market/decentralized/src/matcher/store.rs
```diff
@@ -218,4 +253,37 @@ impl SubscriptionStore {
             false => Err(DemandError::NotFound(demand_id.clone())),
         }
     }
+
+    /// Returns Offers SubscriptionId from vector, that don't exist in our database.
+    pub async fn filter_existing(
```
tworec, line 258:
```suggestion
    pub async fn filter_out_existing(
```

core/market/decentralized/src/identity.rs
nieznanysprawiciel: Are you sure we want to cease broadcasting for locked identities?
That means that Offers from locked identities shouldn't appear on other Nodes? I think it's impossible.

Probably we want to disable negotiations for locked identities (and what more important signing Agreements). But still I don't think, we should filter these identities here

core/market/decentralized/src/protocol/discovery.rs
```diff
@@ -52,60 +58,81 @@ pub struct Discovery {
...
+        // TODO: We shouldn't use send_as. Put identity inside broadcasted message instead.
+        let _ = bus::service(broadcast::BUS_ID)
+            .send_as(default_id, bcast_msg) // TODO: should we send as our (default) identity?
             .await?;
         Ok(())
     }
 
-    pub async fn broadcast_unsubscribe(
+    /// Ask remote Node for specified Offers.
+    pub async fn get_offers(
```
tworec, line 92: to indicate it is remote operation, not local store select
```suggestion
    pub async fn get_offers_from(
```

core/market/decentralized/src/matcher.rs
```diff
@@ -140,60 +181,334 @@ impl Matcher {
...
+        Ok(our_offers)
     }
 }
 
-pub(crate) async fn on_offer_received(
+// =========================================== //
+// Discovery protocol messages handlers
+// =========================================== //
+
+pub(crate) async fn on_offer_ids_received(
```
tworec, line 226: misleading name.  Below is my proposition. more descriptive one
```suggestion
pub(crate) async fn filter_known_offer_ids(
```

core/market/decentralized/src/matcher.rs
```diff
@@ -140,60 +181,334 @@ impl Matcher {
...
-            SaveOfferError::Expired(_) => Ok(Propagate::No(Reason::Expired)),
-            // Below errors are not possible to get from checked_store_offer
-            SaveOfferError::SaveError(_, _)
-            | SaveOfferError::SubscriptionValidation(_)
-            | SaveOfferError::WrongState { .. } => {
-                Ok(Propagate::No(Reason::Error(format!("{}", e))))
+        .map_err(|e| log::warn!("Error filtering Offers. Error: {}", e))?)
+}
+
+pub(crate) async fn on_offers_received(
```
tworec, line 241: same here - maybe?
```suggestion
pub(crate) async fn receive_offers_to_propagate(
```

tworec (commented):
more comments

core/market/decentralized/src/db/dao/offer.rs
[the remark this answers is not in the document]
nieznanysprawiciel: I have to do 2 methods, because `get_offer` returns `OfferState` not `Offer`, so I can't write batch version that returns `Vec<OfferState>`.

core/market/decentralized/src/protocol/discovery.rs
nieznanysprawiciel: Hmmm I think it's obvious. You call these method on Discovery interface and pass NodeId in parameter. I don't think it can be missinterpreted

core/market/decentralized/src/matcher.rs
nieznanysprawiciel: For me your proposal is more misleading. Propagation isn't main goal of this callback, but name suggests it.

core/market/decentralized/src/market.rs
nieznanysprawiciel: I think, having it on market level isn't something bad. It is better api if you call `get_offers` instead dispatching `matcher.store...`.
But if you change it in future, I don't mind.

core/market/decentralized/src/matcher.rs
nieznanysprawiciel: Added comment for future

core/market/decentralized/src/db/model/subscription_id.rs
nieznanysprawiciel: I used this formatter within `DisplayVec` implementation. If you meant to use it not here, but in places I display this vector, I'm not sure it would be a good idea, because it seems to be to long

core/market/decentralized/src/matcher/store.rs
nieznanysprawiciel: I think the opposite. The bottom layer (Dao) should have the most generic functions as we can make. Top layers like Store should have specialized functions that hide unused details.
It makes `Store` a little bit more useful

core/market/decentralized/src/db/model/subscription_id.rs
tworec: That's what I ment. Thx

core/market/decentralized/src/identity.rs
tworec: I've performed a test using locked identity via yapapi and it has passed. So locked identities are... not locked in sense I'd expect. Let's leave it for now.

I'd expect that locked identity does not allow usage REST APIs at all, ie. no subscriptions, no activities, no debit notes nor invoices. But thats sth which must be discussed on higher level than just Marketplace.

tworec: Referenced this comment in #529

core/market/decentralized/src/matcher/store.rs
tworec: 👍 agreed

core/market/decentralized/src/identity.rs
nieznanysprawiciel: I agree to everything you said, that shouldn't be possible ;)
But I don't think, we shouldn't broadcast Offers, if they were already there.
To discuss later.

core/market/decentralized/src/market.rs
tworec: ok, I've made an issue for it #530

core/market/decentralized/src/protocol/discovery.rs
tworec: It was not to me. 
I propose below naming scheme for this file:
* sender `broadcast_offers` and its receiver `on_broadcast_offers`
* `broadcast_unsubscribes` and `on_broadcast_unsubscribes`
* `retrieve_offers` --> `on_retrieve_offers`

`get_offers` resembles local db operation rather than remote

tworec: And consequently `OffersReceived` rename to `OffersRetrieved` as it reflects this operation is triggered by us and not via broadcast.

nieznanysprawiciel: Fixed

core/market/decentralized/src/matcher.rs
tworec: Ok, my proposition is far from good ;) 

But my point is naming those fns as `on_*` suggests that they are handlers, and they are, but not directly. Their handler nature is depicted via inputting them to the `add_data_handler`.

To be consistent with above I'd rename:
*  `on_offers_received` to `save_and_match_offers` and  
* `filter_known_offer_ids` to `filter_out_known_offer_ids`.
* `on_get_offers` to simply `get_offers`
* `on_offer_unsubscribed` to `unsubscribe_offers`

This will give an idea of what are those fns purposes.
Current names only depicts **how** they are used, and not **what for**.

tworec: I'm preparing PR based on this PR with my proposals. They are to broad to be contained within comments

tworec: see #531

core/market/decentralized/src/matcher/store.rs
tworec: it was a pain for me that we do it that not optimal, so I've rewritten to use single select instead of multiple selects.
It was a broader effort of rewriting OfferDao

core/market/decentralized/src/matcher.rs
tworec: fixed

core/market/decentralized/src/matcher.rs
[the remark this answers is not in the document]
tworec: Every positive number here will result with proper mean value when you are using standard distribution.

core/market/decentralized/src/protocol/discovery.rs
tworec: rethinked again and also fixed

core/market/decentralized/src/matcher.rs
tworec: fixed

tworec (approved):

PR #461, opened 2020-07-29
SGX support

prekucki:


mfranciszkiewicz (dismissed):

mfranciszkiewicz (approved):

nieznanysprawiciel (approved):

Wiezzel (approved):

mfranciszkiewicz (approved):

core/activity/src/requestor/control.rs
```diff
@@ -15,51 +19,99 @@ use crate::common::{
...
     id: Identity,
 ) -> impl Responder {
-    let agreement_id = body.into_inner();
-    authorize_agreement_initiator(id.identity, &agreement_id).await?;
+    let agreement_id = body.agreement_id();
+    authorize_agreement_initiator(id.identity, agreement_id).await?;
 
     let agreement = get_agreement(&agreement_id).await?;
-    log::trace!("agreement: {:#?}", agreement);
+    log::info!("agreement: {:#?}", agreement);
```
tworec, line 83: debug!

nieznanysprawiciel: I would even leave it at trace level. I don't think we should log whole structures on debug level, because it prevents us from using this level..

PR #469, opened 2020-08-04
Renamed 'GNT' to 'NGNT'

Wiezzel:
'NGNT' is the approved ticker for testnet and shall be used in all user-facing situations (instructions, logs, CLI, etc.)


agent/provider/src/preset_cli.rs
```diff
@@ -64,7 +64,7 @@ impl PresetUpdater {
                 .cloned()
                 .unwrap_or(0.);
             let price = Input::<f64>::new()
-                .with_prompt(&format!("{} (GNT)", coefficient.to_readable()))
+                .with_prompt(&format!("{} (/NGNT)", coefficient.to_readable()))
```
maaktweluit, line 67: what is the extra `/` for?

Wiezzel: Just a typing mistake.

bidzyyys:
Then I will approve

PR #472, opened 2020-08-10
Add core SGX service

omeg:
SGX service with (so far) GSB API for verifying enclave quotes with Intel Attestation Service

### DIFF
--- a/core/model/Cargo.toml
+++ b/core/model/Cargo.toml
@@ -25,7 +25,7 @@ identity = []
 market = []
 net = []
 payment = ['bigdecimal']
-sgx = ['graphene_api']
+sgx = ['sgx-types']
 
 [dependencies]
 ya-client-model = "0.1"
@@ -37,4 +37,4 @@ log = "0.4"
 serde = { version = "^1.0", features = ["derive"] }
 serde_bytes = "0.11.3"
 thiserror = "1.0.9"
-graphene_api = { git = "https://github.com/golemfactory/graphene-rust.git", rev = "8e14e8f012e790ccce5c696f20c1d750ca3f4398", optional = true }
+sgx-types = { git = "https://github.com/golemfactory/sgx-types.git", rev = "157af8487d538062155d98712fee61a6d69c8ced", features = ['with-serde'], optional = true }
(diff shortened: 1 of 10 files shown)


exe-unit/src/service/metrics.rs
```diff
@@ -95,7 +95,7 @@ impl Handler<GetMetrics> for MetricsService {
         let mut metrics = Vec::with_capacity(self.usage_vector.len());
 
         for name in self.usage_vector.iter() {
-            let metric = self
+            /*let metric = self
```
mfranciszkiewicz, line 98: How about making this feature enabled, e.g. with `#[cfg(feature = "sgx")]`, `#[cfg(not(feature = "sgx"))]`

omeg: Oops, this commit was included here by mistake.

core/model/src/sgx.rs
```diff
@@ -0,0 +1,29 @@
+pub use graphene::ias::AttestationResponse;
+use serde::{Deserialize, Serialize};
+use ya_service_bus::RpcMessage;
+
+/// Public SGX bus address.
+pub const BUS_ID: &str = "/public/sgx";
```
mfranciszkiewicz, line 6: Will these endpoints be called via network? If not, bindings could be local

core/model/Cargo.toml
```diff
@@ -35,3 +37,4 @@ log = "0.4"
 serde = { version = "^1.0", features = ["derive"] }
 serde_bytes = "0.11.3"
 thiserror = "1.0.9"
+graphene_api = { git = "https://github.com/golemfactory/graphene-rust.git", rev = "8e14e8f012e790ccce5c696f20c1d750ca3f4398", optional = true }
```
mfranciszkiewicz, line 40: It might be a good idea to split that crate. AFAIU only models are needed here

mfranciszkiewicz (changes requested):

exe-unit/src/service/metrics.rs
omeg: Rebased without this commit, it'll be in a separate PR.

core/model/src/sgx.rs
omeg: At first I thought it would be needed but in reality this will only be called locally. Changed.

mfranciszkiewicz (approved):

PR #485, opened 2020-08-17
Retry limit for invoice issuing and sending in provider agent

Wiezzel:
Without the limit agent could get stuck forever on a failed operation and thus become idle. That was serious risk esp. in case of re-sending because requestor might have just disconnected making it impossible to deliver an invoice.


tworec:
Closing in favour of #486

PR #487, opened 2020-08-18
Refactored payment driver initialization

Wiezzel:
Every payment driver has a separate feature toggle allowing to use multiple drivers in any configuration. Specialized driver initialization code could be added into `start_payment_drivers` function. Applying `#[cfg(feature = ...)]` to variants in `Services` enum didn't work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,11 +10,12 @@ license = "GPL-3.0"
 edition = "2018"
 
 [features]
-default=['market-forwarding']
+default=['market-forwarding', 'gnt-driver']
 static-openssl=["openssl/vendored"]
 market-forwarding = ['ya-market-forwarding']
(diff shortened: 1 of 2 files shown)


Wiezzel:
@maaktweluit 
For testing you can run yagna with different configurations and see if drivers are properly started by calling `payment init` and/or `payment status`.
* `cargo run --release --bin yagna -- service run` – only nGNT driver
* `cargo run --release --features dummy-driver --bin yagna -- service run` – both nGNT and dummy driver
* `cargo run --release --no-default-features --features dummy-driver --bin yagna -- service run` – only dummy driver
You may also manipulate `accounts.json` between runs and change the driver there for auto-init.

maaktweluit (approved):
LGTM!

PR #488, opened 2020-08-18
[exe-unit] Single process killing

boryspoplawski:
These changes are needed to run Yagna in Graphene-SGX, which currently does not fully support `/proc` subsystem (used by `ProcessTree`).

### DIFF
--- a/exe-unit/src/runtime/process.rs
+++ b/exe-unit/src/runtime/process.rs
@@ -2,7 +2,11 @@ use crate::error::Error;
 use crate::message::{
     RuntimeCommand, RuntimeCommandResult, SetRuntimeMode, SetTaskPackagePath, Shutdown,
 };
-use crate::process::{ProcessTree, SystemError};
+#[cfg(feature = "sgx")]
+use crate::process::kill;
+#[cfg(not(feature = "sgx"))]
+use crate::process::ProcessTree;
+use crate::process::SystemError;
 use crate::runtime::event::EventMonitor;
 use crate::runtime::{Runtime, RuntimeArgs, RuntimeMode};
 use crate::ExeUnitContext;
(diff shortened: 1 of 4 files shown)


exe-unit/src/runtime/process.rs
```diff
@@ -164,12 +164,11 @@ impl RuntimeProcess {
                 .stderr(Stdio::piped())
                 .spawn()?;
 
-            let tree =
-                ProcessTree::try_new(child.id()).map_err(|e| Error::RuntimeError(e.to_string()))?;
+            let signle_child = child.id();
```
mfranciszkiewicz, line 167: typo in `single_child`

mfranciszkiewicz (changes requested):
Please don't remove `ChildProcess::Tree` and make it feature-enabled in the following fashion:

```rust
enum ChildProcess {
    #[cfg(feature = "sgx")]
    Single { pid: u32 },
    #[cfg(not(feature = "sgx"))]
    Tree(process::ProcessTree),		     
    Service(ProcessService)
}
```

or

```rust
enum ChildProcess {
    #[cfg(feature = "sgx")]
    Subprocess { pid: u32 },
    #[cfg(not(feature = "sgx"))]
    Subprocess { tree: process::ProcessTree },		     
    Service(ProcessService)
}
```

boryspoplawski:
> Please don't remove `ChildProcess::Tree` and [...]

What's the use of `ProcessTree` here? Seems that it always contains a single process anyway ...

exe-unit/src/runtime/process.rs
boryspoplawski: Done.

mfranciszkiewicz:
> > Please don't remove `ChildProcess::Tree` and [...]
> 
> What's the use of `ProcessTree` here? Seems that it always contains a single process anyway ...

We're not limiting runtimes to a single process. Replacing `ProcessTree` with `Single` is like reverting to a pre-poc implementation, which shouldn't be done on a whim

boryspoplawski:
> We're not limiting runtimes to a single process. Replacing `ProcessTree` with `Single` is like reverting to a pre-poc implementation, which shouldn't be done on a whim

I missed that `ProcessTree` iterates through all system processes.
Done.

mfranciszkiewicz (approved):

PR #497, opened 2020-08-20
[mkt] serial tests using serial_test crate

tworec:
All test using `MarketsNetwork::new` are now tagged to be executed serially with `serial_test` crate (https://github.com/palfrey/serial_test/pull/23).


core/market/decentralized/tests/test_initial_proposal.rs
```diff
@@ -22,7 +23,7 @@ async fn test_query_events_non_existent_subscription() -> Result<(), anyhow::Err
     let non_existent_id = "80da375cb604426fb6cddd64f4ccc715-85fdde1924371f4a3a412748f61e5b941c500ea69a55a5135b886a2bffcb8e55".parse()?;
 
     // We expect that no events are available for non existent subscription.
-    let result = market1.query_events(&non_existent_id, 1.2, Some(5)).await;
+    let result = market1.query_events(&non_existent_id, 0.2, Some(5)).await;
```
nieznanysprawiciel, line 26: This comment applies to all timeout changes:
Are you sure, you want to decrease them???
Note that in normal execution that doesn't hurt us at all, because we use notification to wake up from waiting.
In incorrect execution it matters, but I think it's good to have bigger margin, to avoid non deterministic tests.

nieznanysprawiciel (approved):
I think it's amazing :D
I found this crate somewhere back than, but it didn't have support for async tests, so I abandoned this idea.
And you've just modified it ;)

nieznanysprawiciel:
If we don't need to limit threads anymore (explicite), you could update readme.md and probably CI scripts.

tworec:
> If we don't need to limit threads anymore (explicite), you could update readme.md and probably CI scripts.

I will, but for now I would just give it a few weeks to see if it works correctly.
Furthermore only @prekucki has super powers to merge CI scripts currently (AFAIK).

core/market/decentralized/tests/test_initial_proposal.rs
tworec: I've decided to give it a try, but your comment given me opportunity to think again and I would not mix changing test execution (with macro) and other changes. This will enable easier reverting in case of failure of this approach with `serial_test`.

core/market/decentralized/Cargo.toml
```diff
@@ -56,5 +56,7 @@ env_logger = { version = "0.7", optional = true }
 ya-market-decentralized = { version = "0.1", features = ["testing"] }
 
 serde_json = "1.0"
+#serial_test = "0.4.0"
```
jiivan, line 59:
```suggestion
```

jiivan (approved):

jiivan (approved):

jiivan (approved):

prekucki (approved):

PR #504, opened 2020-08-24
Merge release/vm poc

mfranciszkiewicz:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "yagna"
-version = "0.3.0"
+version = "0.3.3-alpha.4"
```
tworec, line 3:
```suggestion
version = "0.3.3-alpha.5"
```

mfranciszkiewicz: @tworec I'd suggest `0.4.0` since ExeUnit uses a completely different communication api with the vm runtime

tworec: is it back compatible? I need some sync with you on that

tworec: but this seems legit afai understand

mfranciszkiewicz: @tworec on a second thought, `ya-exe-unit` and `ya-provider` are versioned separately, so I'm leaving this change to other reviewers

prekucki (dismissed):

mfranciszkiewicz:
Closing in favour of #543

PR #511, opened 2020-08-26
Test timing

nieznanysprawiciel:


.github/workflows/rust.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - market-decentralized/errors-experiments
```
prekucki, line 7: what is it needed for?

prekucki (changes requested):

.github/workflows/rust.yml
nieznanysprawiciel: Don't worry. This branch isn't meant to be ever merged to master ;)

PR #513, opened 2020-08-27
Add CLI config option for negotiator strategy

kubkon:
This seemed to have been missing in the provider CLI and overriding
the default behaviour of accepting only one agreement to accepting
all agreements is really useful for local testing.

### DIFF
--- a/agent/provider/src/startup_config.rs
+++ b/agent/provider/src/startup_config.rs
@@ -60,6 +60,11 @@ pub struct ProviderConfig {
         env = "YA_RT_STORAGE")
     ]
     pub rt_storage: Option<f64>,
+
+    /// Negotiator strategy used by this provider.
+    /// Can be one of: LimitAgreements, AcceptAll.
+    #[structopt(long, default_value = "LimitAgreements")]
+    pub negotiator_strategy: String,
 }
 
 impl ProviderConfig {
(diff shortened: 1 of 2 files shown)


agent/provider/src/startup_config.rs
```diff
@@ -60,6 +60,11 @@ pub struct ProviderConfig {
         env = "YA_RT_STORAGE")
     ]
     pub rt_storage: Option<f64>,
+
+    /// Negotiator strategy used by this provider.
+    /// Can be one of: LimitAgreements, AcceptAll.
+    #[structopt(long, default_value = "LimitAgreements")]
+    pub negotiator_strategy: String,
```
tworec, line 67: pls define a dedicated Enum with possible strategies

mfranciszkiewicz: `AcceptAll` doesn't seem to be a viable strategy with the current implementation

kubkon: @mfranciszkiewicz oh, how come?

mfranciszkiewicz: @kubkon `AcceptAllNegotiator` accepts every received proposal / agreement, so it can potentially spawn an unlimited number of activities. Providers don't currently manage free hardware resources

kubkon: Makes sense! In that case, I'm gonna go ahead and close this PR for the moment.

PR #518, opened 2020-08-28
[ya-provider] Spawn ExeUnit in its own process group

mfranciszkiewicz:
Introduces separation of process groups as a countermeasure for any potential issues with ExeUnit's process tree scanning

### DIFF
--- a/agent/provider/src/execution/exeunit_instance.rs
+++ b/agent/provider/src/execution/exeunit_instance.rs
@@ -34,6 +34,13 @@ impl ExeUnitInstance {
         let mut command = Command::new(&binary_path);
         command.args(args).current_dir(working_dir);
 
+        #[cfg(unix)]
+        {
+            // executes the command in its own process group
+            use ya_utils_process::CommandSetSid;
+            command.setsid();
+        }
+
         let child = ProcessHandle::new(&mut command).map_err(|error| {
             anyhow!(
                 "Can't spawn ExeUnit [{}] from binary [{}] in working directory [{}]. Error: {}",
--- a/agent/provider/src/execution/exeunit_instance.rs
+++ b/agent/provider/src/execution/exeunit_instance.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::time::Duration;
 
-use ya_utils_process::ProcessHandle;
+use ya_utils_process::{ProcessGroupExt, ProcessHandle};
 
 /// Working ExeUnit instance representation.
 #[derive(Display)]
@@ -32,14 +32,11 @@ impl ExeUnitInstance {
             .with_context(|| format!("Failed to spawn [{}].", binary_path.display()))?;
 
         let mut command = Command::new(&binary_path);
-        command.args(args).current_dir(working_dir);
-
-        #[cfg(unix)]
-        {
-            // executes the command in its own process group
-            use ya_utils_process::CommandSetSid;
-            command.setsid();
-        }
+        // new_process_group is a no-op on non-Unix systems
+        command
+            .args(args)
+            .current_dir(working_dir)
+            .new_process_group();
 
         let child = ProcessHandle::new(&mut command).map_err(|error| {
             anyhow!(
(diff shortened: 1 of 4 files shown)


utils/process/src/lib.rs
```diff
@@ -14,6 +14,25 @@ use {
...
+    fn setsid(&mut self) -> &mut T;
+}
+
+#[cfg(unix)]
+impl CommandSetSid<Command> for Command {
+    fn setsid(&mut self) -> &mut Command {
+        use std::os::unix::process::CommandExt;
+        unsafe {
+            self.pre_exec(|| {
+                libc::setsid();
```
kubkon, line 28: Is it intentional we don't check for errors? Also, perhaps it might be better to use a wrapper [`nix::setsid()`](https://docs.rs/nix/0.18.0/nix/unistd/fn.setsid.html)?

agent/provider/src/execution/exeunit_instance.rs
```diff
@@ -34,6 +34,13 @@ impl ExeUnitInstance {
         let mut command = Command::new(&binary_path);
         command.args(args).current_dir(working_dir);
 
+        #[cfg(unix)]
+        {
+            // executes the command in its own process group
+            use ya_utils_process::CommandSetSid;
+            command.setsid();
+        }
+
```
nieznanysprawiciel, line 43: Could we keep provider free from system dependent code inside compilation switches??
In this case we could for example implement this trait for all systems (in process/src/lib.rs), but on non unix system it would resolve to empty implementation.

This call in provider could look like this:
```
command.args(args).current_dir(working_dir).new_process_group();
```

nieznanysprawiciel (commented):
Great you've done this change :)

nieznanysprawiciel:
Hmm my second concern is:
If this is ExeUnit _Supervisor_'s fault, that to many processes are killed, than maybe _Supervisor_ should spawn _Runtime_ in separate group?? Because now correct behavior of _Supervisor_ depends on correct behavior of _Provider Agent_.

utils/process/src/lib.rs
mfranciszkiewicz: > Is it intentional we don't check for errors?

Kind of - we still want to continue execution (this shouldn't fail btw). I'll return an `Err` and add a debug statement in case of failure, even `warn` seems a bit too much.

>  Also, perhaps it might be better to use a wrapper `nix::setsid()`?

At first I've implemented it with `nix` in the provider agent, but eventually decided to move it to this crate without adding extra dependencies. I'll revert to `nix` anyway.

agent/provider/src/execution/exeunit_instance.rs
mfranciszkiewicz: @nieznanysprawiciel That was also my thought. But I'd personally expect this to also work on Windows (and that's not necessary - we're using `JobObject`s https://github.com/golemfactory/yagna/blob/e59a9ac9ca97922db0caabb25a708541d9257a78/exe-unit/src/process/win.rs)

mfranciszkiewicz:
@nieznanysprawiciel It's been fixed for a long time, no more additional processes are killed. The `release-poc` branch was based on an old version of `master`. Behaviour of the Supervisor does not depend on this change. This is an additional countermeasure. 

There are multiple angles on why it's implemented this way:

- metrics collection, `getrusage` does not take into account other process groups
- macOS compatibility: no cgroups and also metric collection
- terminating grandchildren is easier (though they still can change the group)

agent/provider/src/execution/exeunit_instance.rs
nieznanysprawiciel: I wasn't talking only about functionalities. For me it is ugly, that we expose system level details here.
We need abstraction level that deals with differences between systems. _Provider Agent_ is on higher level and should only use abstraction.
Compilation switches like this should be hidden behind this abstraction.

nieznanysprawiciel (approved):

agent/provider/src/execution/exeunit_instance.rs
mfranciszkiewicz: I've implemented it as requested (one of my previous iterations). 

I'd agree with you if there existed an implementation for all target platforms. Currently it just hides the fact that it doesn't do what is says (on Windows) and each use of this trait needs to be accompanied by a _doesn't work on Windows_ comment

PR #523, opened 2020-08-31
improve bus factor

etam:
Or suggest more people to be added


etam:
The conclusion: Modifying `.github` must be difficult, because actions are potentially dangerous. In case of bus factor issues, repo admins can override this limitation.

Issue #385, opened 2020-06-29
Get Proposal - Requestor


nieznanysprawiciel:
- [ ] Implement GET `demands/{subscription_id}/proposals/{proposal_id}` endpoint

Issue #384, opened 2020-06-29
Get proposal - provider


nieznanysprawiciel:
- [ ] Implement GET `offer/{subscription_id}/proposals/{proposal_id}` endpoint

PR #438, opened 2020-07-20
 Counter initial proposal - error handling

nieznanysprawiciel:
resolves: #402

Notes:
- Reject proposals with wrong generated ProposalId is not tested. It needs some work on our test utils.
- Probably we should do more security checks in negotiation API handlers, but this seems to be in another milestone.

PR #527, opened 2020-08-31
Get Proposal rest endpoints for both Requestor and Provider

nieznanysprawiciel:
resolves: #385 
resolves: #384 

It would be good to write rest endpoint test, but I would do it after merging:
https://github.com/golemfactory/yagna/pull/438
where some useful test utils are present.


core/market/decentralized/src/market.rs
```diff
@@ -158,6 +160,22 @@ impl MarketService {
         // TODO: shouldn't remove precede negotiation unsubscribe?
         Ok(self.matcher.unsubscribe_demand(demand_id, id).await?)
     }
+
+    pub async fn get_proposal(
+        &self,
+        proposal_id: &ProposalId,
+        _id: &Identity,
+    ) -> Result<Proposal, MarketError> {
+        // TODO: Autorization
```
jiivan, line 169: What are the dangers of unauthorized access to proposal?

jiivan (approved):
lgtm

core/market/decentralized/src/market.rs
nieznanysprawiciel: Negotiations are private, meaning that only negotiating parties should know content of Proposals (Other identities on the same daemon probably shouldn't have access either).
But we scheduled thinking about Identities to Milestone 2, so we may decide, that we don't need authorization here.

core/market/decentralized/tests/test_proposal.rs
```diff
@@ -0,0 +1,76 @@
+use ya_market_decentralized::testing::proposal_util::exchange_draft_proposals;
+use ya_market_decentralized::testing::{GetProposalError, MarketsNetwork, OwnerType};
+
+use ya_client::model::market::proposal::State;
+
+#[cfg_attr(not(feature = "market-test-suite"), ignore)]
+#[actix_rt::test]
+async fn test_get_proposal_rest() -> Result<(), anyhow::Error> {
```
tworec, line 8: REST test is missing.

tworec (changes requested):

core/market/decentralized/tests/test_proposal.rs
jiivan: added

core/market/decentralized/tests/test_proposal.rs
```diff
@@ -0,0 +1,76 @@
...
+    assert_eq!(proposal.issuer_id()?, &prov_id.identity.to_string());
+    assert!(proposal.prev_proposal_id().is_ok());
+    Ok(())
+}
+
+/// Try to query not existing Proposal.
+#[cfg_attr(not(feature = "market-test-suite"), ignore)]
+#[actix_rt::test]
+async fn test_get_proposal_not_found() -> Result<(), anyhow::Error> {
+    let network = MarketsNetwork::new("test_get_proposal_rest")
```
tworec, line 51:
```suggestion
    let network = MarketsNetwork::new("test_get_proposal_not_found")
```

tworec: Will change it within conflicts resolve

tworec (changes requested):
My review in form of PR #579

tworec (approved):

PR #528, opened 2020-09-01
reenable clippy

etam:


prekucki (changes requested):
Clippy was disabled for reason. Many rules there are only suggestions. and with no warnings rule it breaks too offen.

etam:
I've changed it, so it makes only warnings and no errors.

tworec:
> I've changed it, so it makes only warnings and no errors.

This adds over 5mins of build time, produces over 1.1k warnings which are then ignored. So causing no effect.
I agree we should reenable it at some point in the future, but currently it is to early, as we are rather quick-and-dirty delivering features in the first place.

tworec:
I suggest adding a request to Tech Council to consider this https://www.notion.so/golemnetwork/87c28c874fc54d76845127946e72e3a7?v=2fe57c56530c41839e6bbcf222c026e2

PR #538, opened 2020-09-03
[mkt] minor renames and upgrades for tests

tworec:


core/market/decentralized/tests/test_cyclic_broadcasts.rs
```diff
@@ -107,9 +107,6 @@ async fn test_unsubscribes_cyclic_broadcasts() -> Result<(), anyhow::Error> {
 
     let market3 = network.get_market("Node-3");
 
-    // Create some offers before we instantiate 3rd node.
-    // This way this 3rd node won't get them in first broadcasts, that
-    // are sent immediately, after subscription is made.
```
nieznanysprawiciel, line 112: Why is this removed?

core/market/decentralized/tests/test_cyclic_broadcasts.rs
```diff
@@ -310,9 +321,6 @@ async fn test_sharing_someones_else_unsubscribes() -> Result<(), anyhow::Error>
...
```
nieznanysprawiciel, line 315: Why is this removed?

nieznanysprawiciel (approved):

core/market/decentralized/tests/test_cyclic_broadcasts.rs
tworec: because this is not true in this test. Node-3 was created already in line 302

core/market/decentralized/tests/test_cyclic_broadcasts.rs
tworec: because this is not true in this test. Node-3 was created already in line 99

PR #572, opened 2020-09-10
remember TOS accepted

### LINKED ISSUE #571: save info about accepted terms
#474 introduced accepting terms and conditions, but asks for it on every run.

etam:
Resolves #571


etam:
As I understand, those terms are a temporary measure. User should agree to them in the installer.

Issue #570, opened 2020-09-10
Agreements naming conventions (`R-*` and `P-*`) causes problem in activity API


nieznanysprawiciel:
Activity API sends AgreementId to provider, which can't find it in hist database because of prefix.

- [x] Agreements should expose in API the same id for both sides. We translate it internally
- [x] Rethink if there are any problems with naming in case of proposals

Temporary workaround:
https://github.com/golemfactory/yagna/pull/562

PR #562, opened 2020-09-09
Market decentralized/get agreement fix

tworec:
fixes #559

PR #574, opened 2020-09-10
[mkt] get agreement should return for both sides: Provider and Requestor

nieznanysprawiciel:
Cherry-picked workaround from:
https://github.com/golemfactory/yagna/pull/562

Real fix in issue:
#570


jiivan (approved):
hesitantly approved :)

PR #575, opened 2020-09-10
ExeScript signatures with the SIGN command

mfranciszkiewicz:


exe-unit/src/handlers/local.rs
```diff
@@ -117,6 +117,37 @@ where
...
+    fn handle(&mut self, msg: SignExeScript, _: &mut Context<Self>) -> Self::Result {
+        let batch = self.state.batches.get(&msg.batch_id).ok_or_else(|| {
+            Error::RuntimeError(format!(
+                "signing an unknown ExeScript (batch [{}])",
+                msg.batch_id
+            ))
+        })?;
+
+        let script = batch.exe_script.clone();
+        let json = serde_json::to_string(&script)?;
```
omeg, line 133: Is the serialization deterministic?

mfranciszkiewicz: Yes, it will match the output generated by this command

exe-unit/src/handlers/local.rs
```diff
@@ -117,6 +117,40 @@ where
...
+        })?;
+
+        let mut response = SignExeScriptResponse {
+            script: batch.exe_script.clone(),
+            results: self.state.batch_results(&msg.batch_id),
+            digest: "sha3".to_string(),
+            sig: None,
+        };
+
+        let json = serde_json::to_string(&response)?;
```
prekucki, line 139: i think this is bad idea. serde_json::to_string could change the order between properties. (or spaces).
suggestion:
 - results as String (same sa script).
 - signature calculated from utf-8:

```rust
format!(r"ya-exe-unit execution signature\nscript: {}\nresults: {}, script, results)
```

prekucki (changes requested):

exe-unit/src/handlers/local.rs
mfranciszkiewicz: The order does not change (at least on the same machine) and we provide the output to calculate the hash from.

The only option to introduce indeterminism is to use a hash map (that could be remedied with the `preserve_order` feature in serde_json). But in the long run, if someone alters the model, I agree - something may go wrong.

mfranciszkiewicz: I've converted the output used for generating the signature to a string to make things more clear (and remain parsable)

prekucki (approved):

PR #577, opened 2020-09-11
Agreement naming fix

nieznanysprawiciel:
resolves: #570

### DIFF
--- a/core/market/decentralized/readme.md
+++ b/core/market/decentralized/readme.md
@@ -35,13 +35,13 @@ RUST_LOG=debug cargo test -p ya-market-decentralized --features ya-market-decent
 
 ### Building .deb
 Prerequisites: 
-- You need cargo-deb installed.
+- You need cargo-deb installed (`cargo install cargo-deb`).
 - Build .deb on the oldest operating system version, you want to support.
 Otherwise linking with GLIBC will fail.
 
 Build yagna with all binaries needed in .deb:
 ```
-cargo build --release --no-default-features --features market-decentralized --features gnt-driver -p yagna -p gftp -p ya-requestor -p ya-provider -p ya-exe-unit
+cargo build --release --no-default-features --features market-decentralized --features gnt-driver --workspace
 ```
 
 Run cargo-deb using binaries compiled in previous step:
--- a/core/market/decentralized/readme.md
+++ b/core/market/decentralized/readme.md
@@ -46,5 +46,5 @@ cargo build --release --no-default-features --features market-decentralized --fe
 
 Run cargo-deb using binaries compiled in previous step:
 ```
-cargo-deb --no-build
+cargo deb --no-build
 ```
--- a/core/market/decentralized/readme.md
+++ b/core/market/decentralized/readme.md
@@ -32,3 +32,19 @@ add `env_logger::init();` on the beginning.
 ```
 RUST_LOG=debug cargo test -p ya-market-decentralized --features ya-market-decentralized/market-test-suite 
 ```
+
+### Building .deb
+Prerequisites: 
+- You need cargo-deb installed.
+- Build .deb on the oldest operating system version, you want to support.
+Otherwise linking with GLIBC will fail.
+
+Build yagna with all binaries needed in .deb:
+```
+cargo build --release --no-default-features --features market-decentralized --features gnt-driver -p yagna -p gftp -p ya-requestor -p ya-provider -p ya-exe-unit
+```
+
+Run cargo-deb using binaries compiled in previous step:
+```
+cargo-deb --no-build
+```
--- a/core/market/decentralized/src/db/model/proposal_id.rs
+++ b/core/market/decentralized/src/db/model/proposal_id.rs
@@ -23,12 +23,18 @@ pub enum OwnerType {
     Requestor,
 }
 
+const HASH_SUFFIX_LEN: usize = 64;
+
 #[derive(Error, Debug, PartialEq)]
 pub enum ProposalIdParseError {
-    #[error("Proposal id [{0}] has invalid format.")]
+    #[error("Id [{0}] has invalid format.")]
     InvalidFormat(String),
-    #[error("Proposal id [{0}] has invalid owner type.")]
+    #[error("Id [{0}] has invalid owner type.")]
     InvalidOwnerType(String),
+    #[error("Id [{0}] contains non hexadecimal characters.")]
+    NotHexadecimal(String),
+    #[error("Id [{0}] hash has invalid length. Should be |{}|", HASH_SUFFIX_LEN)]
+    InvalidLength(String),
 }
 
 #[derive(thiserror::Error, Debug, PartialEq, Serialize, Deserialize)]
@@ -85,6 +91,27 @@ impl ProposalId {
         }
         Ok(())
     }
+
+    /// Clients on both Requestor and Provider side should use the same id,
+    /// because the comunicate with each other and exchange this id.
+    pub fn to_client(&self) -> String {
+        self.id.clone()
+    }
+
+    pub fn from_client(s: &str, owner: OwnerType) -> Result<ProposalId, ProposalIdParseError> {
+        if !s.chars().all(|character| character.is_ascii_hexdigit()) {
+            Err(ProposalIdParseError::NotHexadecimal(s.to_string()))?;
+        }
+
+        if s.len() != HASH_SUFFIX_LEN {
+            Err(ProposalIdParseError::InvalidLength(s.to_string()))?;
+        }
+
+        Ok(ProposalId {
+            owner,
+            id: s.to_string(),
+        })
+    }
 }
 
 pub fn hash_proposal(
@@ -121,10 +148,7 @@ impl FromStr for ProposalId {
             _ => Err(ProposalIdParseError::InvalidOwnerType(s.to_string()))?,
         };
 
-        Ok(ProposalId {
-            owner,
-            id: elements[1].to_string(),
-        })
+        ProposalId::from_client(elements[1], owner)
     }
 }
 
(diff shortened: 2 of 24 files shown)


core/market/decentralized/src/db/model/proposal_id.rs
```diff
@@ -85,6 +91,27 @@ impl ProposalId {
         }
         Ok(())
     }
+
+    /// Clients on both Requestor and Provider side should use the same id,
+    /// because the comunicate with each other and exchange this id.
```
jiivan, line 96:
```suggestion
    /// because they communicate with each other and exchange this id.
```

core/market/decentralized/src/market.rs
```diff
@@ -203,13 +203,13 @@ impl MarketService {
     pub async fn get_agreement(
         &self,
         agreement_id: &AgreementId,
-        _id: &Identity,
+        id: &Identity,
     ) -> Result<Agreement, AgreementError> {
         // TODO: Authorization
```
jiivan, line 208: Is id in select an authorization already?

service-bus/router/src/lib.rs
```diff
@@ -242,7 +242,10 @@ where
         let server_addr = match self.pending_calls.entry(msg.request_id.clone()) {
             Entry::Occupied(_) => Err("CallRequest with this ID already exists".to_string()),
             Entry::Vacant(call_entry) => match self.registered_endpoints.get(&msg.address) {
-                None => Err("No service registered under given address".to_string()),
+                None => Err(format!(
+                    "No service registered under given address '{}'.",
+                    &msg.address
```
jiivan, line 247: :+1:

core/market/decentralized/src/market.rs
nieznanysprawiciel: Hmmm we could think about it this way ;)
I'm not sure, that's why I left this comment

core/market/decentralized/src/rest_api/common.rs
```diff
@@ -6,18 +6,35 @@ use ya_service_api_web::middleware::Identity;
...
-        .get_agreement(&agreement_id, &id)
-        .await
-        .inspect_err(|e| log::error!("[GetAgreement] {}", e))
-        .map(|agreement| HttpResponse::Ok().json(agreement))
+    // We don't know, if we are requestor or provider. Try to get Agreement for both sides
+    // and check, if any will be returned. Note that we won't get Agreement if we aren't
+    // owner, so here is no danger, that Provider gets Requestor's Offer and opposite.
+    let body = body.into_inner();
+    let r_agreement_id = body.clone().to_id(OwnerType::Requestor)?;
+    let p_agreement_id = body.to_id(OwnerType::Provider)?;
```
tworec, line 24: This is awkward. It is clear that we do not need to store agreements using owner-prefixed-ids

tworec (changes requested):
I think it will be better to introduce a special type for `AgreementId` without `Owner` or make `ProviderId` generic to let set `OwnerType` as `()`. Or in the worst case, we can assume, that `AgreementId` is always owned by `Requestor`.

The Agreement is self-contained, so we will never need to store a separate copy for Requestor and Provider on the same node. 

If I am right, then it probably will be enough to change
[negotiation/provider.rs line 304](https://github.com/golemfactory/yagna/pull/577/files#diff-f4fe3ac66c934546460b0f35bd59411aR304)

core/market/decentralized/src/db/model/proposal_id.rs
```diff
@@ -85,6 +91,27 @@ impl ProposalId {
         }
         Ok(())
     }
+
+    /// Clients on both Requestor and Provider side should use the same id,
+    /// because they communicate with each other and exchange this id.
+    pub fn to_client(&self) -> String {
```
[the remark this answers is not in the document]
tworec: After reconsideration I'd refrain to communicate other ids to the client that we have internally..

nieznanysprawiciel:
@tworec Doing, what you propose, is very, very bad idea for multiple reasons.

Agreement has it's state and it can be different on `Provider` and `Requestor`.
- First it is created on `Requestor`; On `Provider` it doesn't exist at all at this moment
- When `Requestor` sends it to `Provider`, it is added to database on `Provider` side. If we have only one database entry, we need to have separate code for case when `Provider` speaks with `Requestor` on the same `Node` and separate for case when they are on different Nodes. In first case we must interpret error while adding to db as normal condition, in second case it is error that must be propagated to `Requestor` as fail. I dont' want to have these separate paths.
- But the worst comes here: When Requestor sent Agreement proposal, many things can happen, that affect Agreement state. `Provider` can accept or reject. Requestor waits, but can cancel too. This means that state of `Reqestor's` Agreement can be modified by `Provider`. It will be such a mess, that no tests can cover it, to ensure, that everything works correctly.
- At that's not all. We will implement Agreements signatures in nearest future. When `Provider` signs Agreement, Requestor have it already signed on his side, before event comes.

And even if there were no interactions like this in current code, I wouldn't like to do it either, because it is unpredictable, what problems this can cause in the future. Now we have clear separation between `Provider` and `Requestor` code, that handles Agreement. Breaking this separation can't lead to anything good.

The only way to make this idea work, is preventing from negotiations happening on the same node. But we didn't decided to do this in the past, so we must live with this decision now, at least for some time.

core/market/decentralized/src/market/agreement.rs
```diff
@@ -20,16 +19,24 @@ async fn get_agreement(
...
+    // On GSB we don't know if Provider or Requestor is calling, so we will try both versions.
+    let agreement_id = AgreementId::from_client(&msg.agreement_id, OwnerType::Provider)
         .map_err(|e| RpcMessageError::Market(e.to_string()))?;
+
+    // TODO: We should check Agreement owner, like in REST get_agreement implementation, but
+    //  I'm not sure we can trust `sender_id` value from gsb now.
     let dao = db.as_dao::<AgreementDao>();
     Ok(match dao
-        .select(&agreement_id, chrono::Utc::now().naive_utc())
+        .select(&agreement_id, None, chrono::Utc::now().naive_utc())
```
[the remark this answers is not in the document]
tworec: lets skip it

core/market/decentralized/src/rest_api/common.rs
```diff
@@ -6,18 +6,35 @@ use ya_service_api_web::middleware::Identity;
...
     body: Path<PathAgreement>,
     id: Identity,
 ) -> impl Responder {
-    let agreement_id = body.into_inner().agreement_id;
-    market
-        .get_agreement(&agreement_id, &id)
-        .await
-        .inspect_err(|e| log::error!("[GetAgreement] {}", e))
-        .map(|agreement| HttpResponse::Ok().json(agreement))
+    // We don't know, if we are requestor or provider. Try to get Agreement for both sides
```
tworec, line 19: we should know. We have identity

nieznanysprawiciel: You mean `Identity.role`? Or how you determine, who we are?

tworec:
Ok, let's leave it as it is for now. 
I would argue with some of the points above, but first of all we need to make it work. 
Only then we can care for code simplicity and clarity.

core/market/decentralized/src/protocol/negotiation/provider.rs
```diff
@@ -233,6 +233,8 @@ impl NegotiationApi {
         public_prefix: &str,
         _local_prefix: &str,
     ) -> Result<(), NegotiationApiInitError> {
+        log::info!("Negotiation (Provider) protocol version: mk1");
```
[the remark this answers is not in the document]
nieznanysprawiciel: Note that negotiations don't have versioning in gsb binding points. I wanted to introduce them, but it would be incompatible with our current market, so I decided not to introduce breaking changes in this PR.
And the second thing: This is not `DISCOVERY_VERSION`, but Negotiation protocol version. It should be separate const.

core/market/decentralized/readme.md
```diff
@@ -32,3 +32,19 @@ add `env_logger::init();` on the beginning.
...
+
+### Building .deb
+Prerequisites: 
+- You need cargo-deb installed.
+- Build .deb on the oldest operating system version, you want to support.
+Otherwise linking with GLIBC will fail.
+
+Build yagna with all binaries needed in .deb:
+```
+cargo build --release --no-default-features --features market-decentralized --features gnt-driver -p yagna -p gftp -p ya-requestor -p ya-provider -p ya-exe-unit
```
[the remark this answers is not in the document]
nieznanysprawiciel: Did you checked if it work?

core/market/decentralized/readme.md
```diff
@@ -32,3 +32,19 @@ add `env_logger::init();` on the beginning.
...
+```
+
+Run cargo-deb using binaries compiled in previous step:
+```
+cargo-deb --no-build
```
tworec, line 49:
```suggestion
cargo deb --no-build
```

core/market/decentralized/src/market.rs
nieznanysprawiciel: Ok, if you think so, I will remove this comment

core/market/decentralized/src/db/model/proposal_id.rs
```diff
@@ -23,12 +23,18 @@ pub enum OwnerType {
     Requestor,
 }
 
+const HASH_SUFFIX_LEN: usize = 64;
```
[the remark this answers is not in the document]
nieznanysprawiciel: Changed in both proposal_id and subscription_id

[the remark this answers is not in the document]
nieznanysprawiciel: Checked. It's not possible. Can't initialize const using not compile time expression

tworec (approved):

Issue #533, opened 2020-09-02
monitoring page for Market


tworec:
Continuation of #506 

### Required

- Monitor would request for yagna REST endpoints
  - It needs to expose statistics @ new endpoint (REST)
  - No local persistence of statistics
- Monitor would request only known node (nodes)
- We need two services
- Metric collector service (eg. prometheus + grafana; timeseries; quantitative)
- Dashboard / charts configuration
- Yagna daemon GUI (live; only local data; detailed; custom frontend)

### Optional

- Broader network info will be retrieved using own nodes (via Offers)
- Provider would be also monitored

### Comments

- We are not expecting new provider incarnations
- We do not care to other (hypothetical) Provider compatibility

### PoC functional requirements:
- [x] enough is to have a snapshot of current network status (history not needed) 
- [x] UI: CLI is enough, (Grafana GUI is nice-to-have/optional)
- [x] single Metrics:
 - [x] agreements cnt

### MVP functional reqs:
- [x] all above and
- [x] UI depends on results of PoC
- [x] Metrics
  - [x] nodes cnt (per status: running, shut down)
  - [x] offers cnt
  - [x] demands cnt
  - [x] agreements cnt (per status)
  - [x] activities cnt
  - [x] invoices cnt

### Version with history:
- [x] central active backend storage (Prometeus)
  - active means it sends requests to nodes
- [x] simplistic UI: GUI or CLI (which is simpler) eg. Grafana

_Originally posted by @tworec in https://github.com/golemfactory/yagna/issues/506#issuecomment-692678637_

PR #588, opened 2020-09-16
Metrics prototype

nieznanysprawiciel:
resolves: #533
resolves: #671


tworec (commented):
This is an awesome thing. Keep going!

core/market/decentralized/src/matcher/cyclic.rs
```diff
@@ -73,6 +84,11 @@ pub(super) async fn bcast_unsubscribes(matcher: Matcher) {
                 .discovery
                 .bcast_unsubscribes(our_and_random_ids)
                 .await?;
+
+            let end = Instant::now();
+            counter!("market.offers.unsubscribes.broadcasts", 1);
+            timing!("market.offers.unsubscribes.broadcasts.time", start, end);
```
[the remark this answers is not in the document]
nieznanysprawiciel: I don't know how could this work. You need two macros whatsoever, so i prefer to leave it like this. Or maybe you can propose something better?

jiivan (dismissed):

Wiezzel (approved):
I would advise to make the `counter!` macro variadic to skip the second argument if we want to increase by 1 (this seems to be the majority of cases):
`counter!("foo")` == `counter!("foo", 1)`

nieznanysprawiciel:
> I would advise to make the `counter!` macro variadic to skip the second argument if we want to increase by 1 (this seems to be the majority of cases):
> `counter!("foo")` == `counter!("foo", 1)`
Note that `counter!` is macro from metrics crate, not ours.

Wiezzel:
> Note that `counter!` is macro from metrics crate, not ours.

Didn't see that. Only reviewed the payments part. I guess in this case we'll have to stick with `counter!("foo", 1)`.

nieznanysprawiciel:
Activity exposes metric for usage vector. Interpretation of elements in this vector belongs to Grafana.
Note: that order of elements depends on Agreement and this depends from Offer created by Provider.

prekucki (approved):

tworec (approved):

jiivan (dismissed):

mfranciszkiewicz (approved):

tworec (dismissed):

jiivan (dismissed):

core/payment/src/processor.rs
```diff
@@ -245,6 +246,7 @@ impl PaymentProcessor {
 
         // TODO: Implement re-sending mechanism in case SendPayment fails
 
+        counter!("payment.invoices.requestor.paid", 1);
```
Wiezzel, line 249: This is not semantically accurate. Single payment can be used to pay off multiple invoices if they have the same payee address. (We don't do that now, but data model allows it.) So it should be rather 'sent payments' not 'paid invoices'. If you really want paid invoices then you have to look at the length of `agreement_payments` vector here.

nieznanysprawiciel: Thanks for noticing this.

tworec (approved):

PR #596, opened 2020-09-19
[Windows] Fix for 'An existing connection was forcibly closed by the remote host' on GSB server when client shuts down. 

stranger80:
'An existing connection was forcibly closed by the remote host' on GSB server when client shuts down. This can be observed eg. after Yagna CLI calls to Yagna Daemon - a CLI command leaves 'An existing connection was forcibly closed by the remote host' in Daemon logs.

This seems to be caused by the fact that tokio::net::TcpStream implementation on Windows requires an explicit shutdown() of the connection - this isn't called when TcpStream is being removed. 

Proposed resolution:

As the TcpTransport<> type requires an object implementing AsyncRead+AsyncWrite, a SafeTcpStream wrapper has been added, wrapping TcpStream and implementing Drop trait, where on drop() an explicit shutdown() on TcpStream is called.

No more 'An existing connection was forcibly closed by the remote host' error on Yagna CLI calls on both Windows and Linux.


prekucki (changes requested):
in my opinion, the change as a whole makes no sense. in practice of creating network services I have rarely encountered a situation where it would make sense to distinguish between the method of closing the connection. 
This is due to the fact that either the developer of the application depends on the reliability of the delivery, or they do not depend on it.

If the developer of the application wants to have a guarantee of delivery, relying on the fact that the `shudwon` has finished without timeout is incorrect because it only means that the application took the data from the buffer but it could crash right after.

the only correct approach seems to be to treat `RST` as equal to a normal close. especially when running on `localhost`

stranger80:
Closing as superseded by https://github.com/golemfactory/yagna/pull/616

PR #601, opened 2020-09-22
Metrics as a service

tworec:
relates to #588


jiivan (approved):

core/metrics/src/service.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use ya_service_api_interfaces::Provider;
+
+use crate::metrics::Metrics;
+
+pub struct MetricsService;
+
+lazy_static! {
+    static ref METRICS: Arc<Mutex<Metrics>> = Metrics::new();
+}
```
nieznanysprawiciel, line 14: There's no need for lazy static as long we don't use both gsb and rest.

core/metrics/src/service.rs
```diff
@@ -0,0 +1,38 @@
...
+
+impl MetricsService {
+    // currently just to produce log entry that service is activated
+    pub async fn gsb<C: Provider<Self, ()>>(_ctx: &C) -> anyhow::Result<()> {
+        Ok(())
+    }
+
+    pub fn rest<C: Provider<Self, ()>>(_ctx: &C) -> actix_web::Scope {
+        Scope::new("metrics-api/v1")
+            // TODO:: add wrapper injecting Bearer to avoid hack in auth middleware
```
nieznanysprawiciel, line 29: For me, we shouldn't do this. No need to move hacks between different places in code. Next step should be final solution with identities.

nieznanysprawiciel (approved):

core/metrics/src/service.rs
tworec: it was failing when was not static since we have multiple workers / threads serving REST API

core/metrics/src/service.rs
tworec: might be

PR #499, opened 2020-08-22
[master] Windows 10 compatibility fixes

stranger80:
Pull request to aggregate the Windows 10 compatibility changes, introduces as a result of 'Alpha testing' round of yagna.

- The testnet faucet request was failing on WIn 10 machines. It appears the Actix Web Client is useless when trying to call a URL including a DNS name, which requires DNS resolution, on Windows machine with VM virtual switches configured. It is a known issue, which noone seems to be addressing (https://github.com/actix/actix-web/issues/858) - AWC uses non-default configuration of trust-dns-resolver package, and it seems impossible to easily override it.
Therefore AWC client has been replaced with a 'hyper' HTTP client.

PR #617, opened 2020-09-30
Replace awc with hyper to bypass trust-dns issue on win 10

maaktweluit:
For some windows 10 users the faucet's simlpe http get request gets a timeout instantly.
When investigating this is caused by trust-dns not being able to resolve the hostname, when using a direct IP it works.

This PR solves it by replacing `awc` for `hyper` to resolve the issue.

Payment task: https://golemproject.atlassian.net/browse/PAY-62

Originally fixed in: https://github.com/golemfactory/yagna/pull/517
Extension/replacement of fix in: https://github.com/golemfactory/yagna/pull/499

Original bug in actix: https://github.com/actix/actix-web/issues/1047

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2091,10 +2091,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "http-body"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


maaktweluit:
> migraction to hyper is overkill. just resolve addres usieing google dns before call using https://crates.io/crates/trust-dns-client

Closing in favor of #620

Issue #613, opened 2020-09-28
payments (GNT received/send) metrics


tworec:
- [x] received/earned GNT
- [x] send/paid GNT
as reported by Payment Service

counter takes `u64`, a GNT is `BigDecimal`, so we will use rounding to 10^-9 GNT

PR #622, opened 2020-09-30
Metrics: gnt sent/received

jiivan:
fixes: #613

### DIFF
--- a/core/metrics/src/lib.rs
+++ b/core/metrics/src/lib.rs
@@ -3,3 +3,14 @@ mod metrics;
 mod service;
 
 pub use service::MetricsService;
+
+pub mod utils {
+    const CRYPTOCURRENCY_PRECISION: u64 = 1000000000;
+    use bigdecimal::ToPrimitive;
+    pub fn cryptocurrency_to_u64(amount: &bigdecimal::BigDecimal) -> u64 {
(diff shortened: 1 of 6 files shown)


nieznanysprawiciel (approved):

core/payment/src/service.rs
```diff
@@ -491,8 +493,11 @@ mod public {
             return Err(SendError::BadRequest("Invalid payer ID".to_owned()));
         }
 
+        let driver = payment.payment_platform.clone();
```
Wiezzel, line 496: Driver is not the same thing as platform. It doesn't matter for now, but it's a semantic error.

core/metrics/src/lib.rs
```diff
@@ -3,3 +3,14 @@ mod metrics;
 mod service;
 
 pub use service::MetricsService;
+
+pub mod utils {
+    const CRYPTOCURRENCY_PRECISION: u64 = 1000000000;
```
Wiezzel, line 8: That's of course problematic if we allow other currencies than NGNT.

core/payment/src/service.rs
tworec: fixed via #683

core/metrics/src/lib.rs
tworec: no such plans afaik

PR #632, opened 2020-10-02
Update account not registered error with hint

maaktweluit:
To help users resolve this error themselves: Add a hint that solves this problem in 99% of the cases.

### DIFF
--- a/core/payment/src/error.rs
+++ b/core/payment/src/error.rs
@@ -112,7 +112,7 @@ pub mod processor {
     use ya_core_model::payment::public::SendError;
 
     #[derive(thiserror::Error, Debug)]
-    #[error("Account not registered. platform={platform} address={address} mode={mode:?}")]
+    #[error("Account not registered. Hint: Did you run `yagna payment init -r` after restarting? platform={platform} address={address} mode={mode:?}")]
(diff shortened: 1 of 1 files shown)


core/payment/src/error.rs
```diff
@@ -112,7 +112,7 @@ pub mod processor {
     use ya_core_model::payment::public::SendError;
 
     #[derive(thiserror::Error, Debug)]
-    #[error("Account not registered. platform={platform} address={address} mode={mode:?}")]
+    #[error("Account not registered. Hint: Did you run `yagna payment init -r` after restarting? platform={platform} address={address} mode={mode:?}")]
```
Wiezzel, line 115: The flag in the error message should depend on `mode` field. In case of receiving payments it's `-p`.

maaktweluit: i thought this was not needed since -p is called by default

Wiezzel: Well, it's called by default but only for the default identity. It is possible to create a new identity and not call init.

maaktweluit: good one! added the -r or -p dependent on the mode

Wiezzel (approved):

PR #674, opened 2020-10-07
Handle faucet queue requests by waiting

### LINKED ISSUE #650: decrease log severity for 'Failed to request Eth from Faucet'
All attempts that will be retried should be WARN (with "Retrying" in the msg), and the last (6th) one should be ERROR.

### LINKED ISSUE #670: Faucet shows error when donation is queued
When the faucet has queued the donation the `payment init -r` command returns an error:

deamon log:
```[2020-10-07T12:12:50Z ERROR ya_gnt_driver::gnt::faucet] Failed to request Eth from Faucet: LibraryError("{\"paydate\":1602072800121,\"address\":\"0x67f67edd4be92a2c32023cb48099557814c1b77f\",\"amount\":10000000000000000}")```

init response:
```
$ yagna payment init -r
Error: Library error: Cannot request Eth from Faucet
```

It works well when the donation is made immediately.

Proposed fix:
- Add sleep untill `paydate` and try again.

maaktweluit:
Resolves #670

When requesting a donation at the faucet the "happy path" knows 3 ways:
- Sufficient funds ✅ 
- Transaction for donation made ✅ 
- Donation put in queue, estimated at `<paydate>` ❌ 

The last scenario was not handled properly.

This PR adds reading the `paydate` from the faucet queue response and waiting for max 60 seconds to re-try the request

### DIFF
--- a/core/payment-driver/gnt/src/gnt/faucet.rs
+++ b/core/payment-driver/gnt/src/gnt/faucet.rs
@@ -1,7 +1,12 @@
 use crate::utils;
 use crate::{GNTDriverError, GNTDriverResult};
+use bigdecimal::BigDecimal;
+use chrono::Utc;
+use core::num::TryFromIntError;
 use ethereum_types::Address;
-use std::{env, time};
+use serde::{Deserialize, Serialize};
+use std::convert::TryInto;
+use std::{cmp::min, env, time};
 use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
 use trust_dns_resolver::TokioAsyncResolver;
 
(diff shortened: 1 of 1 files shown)


Wiezzel (approved):
All this type conversion boilerplate is a bit ugly, but I guess we've got to live with that.

maaktweluit:
@Wiezzel fully agree, but strange that the `timestamp()` returns a i64 instead of a u64..

It would also be nicer if all responses would be parsed by serde and then handled.
But this change is too much for a quick fix, and in logic we will not use much from next release ( zksync ).

Wiezzel:
> bit strange that the `timestamp()` returns a i64 instead of a u64..

That's necessary to represent dates before 1970, I guess.

PR #675, opened 2020-10-07
Decentralized Market as default + ver bump + .env-template refresh

tworec:
Decentralized Market is already a default in Alpha.dwa release.
Now after all teams confirmed that we can go, I'm making it a default in master.

It is worth noting that @maaktweluit has confirmed decentralized market impl is compatible with the test-harness via
https://github.com/golemfactory/yagna-integration/runs/1197888812?check_suite_focus=true

Also refreshing `.env-template` as requested by @Wiezzel

### DIFF
--- a/.env-template
+++ b/.env-template
@@ -15,12 +15,15 @@ YAGNA_DATADIR="."
 #GSB_URL=tcp://127.0.0.1:7464
 
 ## REST API
-# Default URL for all APIs.
+#
+# Default HOST:PORT for all REST APIs.
 #YAGNA_API_URL=http://127.0.0.1:7465
-# Specific API URLs.
-#YAGNA_MARKET_URL=http://3.249.139.167:8080/market-api/v1/  # this is default
-#YAGNA_ACTIVITY_URL=http://127.0.0.1:5001/activity-api/v1/
-#YAGNA_PAYMENT_URL=http://127.0.0.1:5001/payment-api/v1/
+#
+# Specific API URLs
+# (default values are derived from YAGNA_API_URL)
+#YAGNA_MARKET_URL=http://127.0.0.1:7465/market-api/v1/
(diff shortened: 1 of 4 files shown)


prekucki (approved):

maaktweluit (approved):

Wiezzel (commented):
Aren't you removing the centralized market completely? My suggestion regarding env was:
> I'd say the var should be removed from env-template in the same commit/PR that **removes the possibility to use market forwarding**.

tworec:
> Aren't you removing the centralized market completely? My suggestion regarding env was:

Not yet. It will be the next step.

> 
> > I'd say the var should be removed from env-template in the same commit/PR that **removes the possibility to use market forwarding**.

The lack of this env var in `.env-template` is intentional. If someone will try to a centralized (forwarding) market impl, which now will become obsolete, it is on him to configure it properly. I do not want to bother average user.

Wiezzel:
> The lack of this env var in `.env-template` is intentional. If someone will try to use a centralized (forwarding) market impl, which now becomes obsolete and requires additional compilation flags, it is on him to configure it properly. I do not want to bother the average user.

The 'average user' wouldn't use the vast majority of the env variables so that's not a valid argument.

nieznanysprawiciel (approved):

tworec:
> > The lack of this env var in `.env-template` is intentional. If someone will try to use a centralized (forwarding) market impl, which now becomes obsolete and requires additional compilation flags, it is on him to configure it properly. I do not want to bother the average user.
> 
> The 'average user' wouldn't use the vast majority of the env variables so that's not a valid argument.

Sorry, I don't have any better, or any other tbh. By average user of `.env-template` I mean one of golem internal devs.
They should not be using market forwarding. And if they are, I let them, but do not support them anymore.

PR #680, opened 2020-10-08
hide access logs for events endpoints

tworec:


prekucki (approved):
As a temporary solution. 
Service should have a logging configuration decorator.

tworec:
yup, just a temporary

PR #685, opened 2020-10-09
remove artifact storing for GH CI build action

tworec:
I believe we do not need it any more


mfranciszkiewicz (approved):

tworec:
Build time gain is marginal, and GH storage is released automatically after 90 days. Closing

PR #687, opened 2020-10-13
NGNT driver: wait for ETH before requesting NGNT

Wiezzel:
Instead of sleeping a fixed amount of time, check the account balance periodically until ETH from faucet is received or timeout (3 minutes) is reached.


maaktweluit (approved):
LGTM!

PR #688, opened 2020-10-13
use ya-client (0.3.1) & ya-client-model (0.1.1) from crates.io

tworec:


prekucki (changes requested):
I do not see the advantages of this change, but only potential problems:

- you make the application bind with any ya-client > = 0.3.0, so it may stop compiling.
- you make it necessary to restore this patch to test changes. (we still want to use patch in during development)

## Why it doesn't matter:

- cargo publish automatically disables patch.
- this also applies to those modules that will hit crates.io, i.e. the patch only applies to building from this repo.

tworec:
My intention was to communicate which version of `ya-client` we are compatible with. But this is minor, so I'm closing.

PR #701, opened 2020-10-15
Disable cache on "CI" build

maaktweluit:
Last tuesday the release branch for alpha.2 [failed to build](https://github.com/golemfactory/yagna/actions/runs/304398982) AFTER PR's got merged.
During the PR's the [builds passed](https://github.com/golemfactory/yagna/actions/runs/304269972) properly.

My suspicion was this was caused by the cache. And confirmed by a [successful build](https://github.com/golemfactory/yagna/actions/runs/304782292) of this branch ( on top of release/alpha2 )

I do not think this is the proper solve. But a good start of an investigation.


tworec (approved):

prekucki (changes requested):
1. We know that the cache sometimes crashes and sometimes you have to manually delete it.
2. But we don't really want to permanently disable the cache. Thanks to this, the build can take 20 minutes instead of 1 hour.

maaktweluit:
Closing, we can manually delete cache when its needed :)

PR #724, opened 2020-10-22
ZkSync faucet

Wiezzel:
Added a faucet integration for the zkSync payment driver. If account is initialized in send mode the driver requests funds via an HTTP request and waits for deposit (max. 1 minute).

Moved wallet unlocking from `schedule_payment` to `init`.

----
Testing: Try payment_api + invoice_flow **without** importing keys from PoC.


maaktweluit (approved):
LGTM!

Thanks for moving the tests, i forgot about those..

Tested the scenario and it works without pre-unlocked keys now :D

PR #744, opened 2020-10-28
Extract payment driver base crate.

maaktweluit:
In preparation for adding persistence.

No special test scenario's, they should all work as before.

Extracted:
- Bus interaction logic
- PaymentDriver trait
- Account storage


core/payment-driver/base/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+*/
+
+extern crate log;
+
+pub mod account;
+pub mod bus;
+pub mod driver;
+pub mod utils;
+
+pub use ya_core_model::driver as model;
```
maaktweluit, line 14: @Wiezzel ( and maybe @prekucki ) what do you think about sharing the model from the base crate this way?

Wiezzel: I'm just  `¯\_(ツ)_/¯`

core/payment-driver/base/src/bus.rs
```diff
@@ -0,0 +1,125 @@
...
+use ya_core_model::payment::local as payment_srv;
+use ya_service_bus::{
+    typed::{service, ServiceBinder},
+    RpcEndpoint,
+};
+
+// Local uses
+use crate::driver::PaymentDriver;
+
+pub fn bind_service(driver: Arc<dyn PaymentDriver>) {
```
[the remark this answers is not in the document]
maaktweluit: nice one! fixed :)

core/payment-driver/base/src/account.rs
```diff
@@ -0,0 +1,77 @@
...
+// Local uses
+use crate::driver::NodeId;
+
+// Public types
+pub type AccountsRc = Rc<RefCell<Accounts>>;
+pub type AccountsRefMut<'a> = RefMut<'a, Accounts>;
+
+pub struct Accounts {
+    accounts: HashMap<String, NodeId>,
+}
```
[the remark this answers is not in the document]
maaktweluit: haha, if you put it like that..

The idea was that this `Account` type needs to be stored on the Driver struct, while it. is managed from the `PaymentDriver` trait.

Would be nice if traits could have property stubs as well.

core/payment-driver/base/src/bus.rs
```diff
@@ -0,0 +1,125 @@
...
+        .map_err(GenericError::new)?;
+    Ok(signature)
+}
+
+pub fn notify_payment(
+    driver: &(dyn PaymentDriver),
+    details: &PaymentDetails,
+    confirmation: Vec<u8>,
+) -> String {
+    let order_id = Uuid::new_v4().to_string();
```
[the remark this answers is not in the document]
maaktweluit: was in doubt if i should generate it here or pass it as an argument

updated to an argument now :)

maaktweluit:
Thanks for. the review!

Made the changes, will do some more testing and push it tomorrow morning

core/payment-driver/base/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+use ya_core_model::driver::{GenericError, PaymentConfirmation, PaymentDetails, SchedulePayment};
+
+pub fn to_payment_details(msg: SchedulePayment) -> PaymentDetails {
+    PaymentDetails {
+        recipient: msg.recipient().to_string(),
+        sender: msg.sender().to_string(),
+        amount: msg.amount(),
+        date: Some(Utc::now()),
+    }
+}
```
Wiezzel, line 19: ~I think it would be better to implement `From<SchedulePayment>` for `PaymentDetails` in `ya_core_model`.~
On second thoughts... this looks easy to misuse. The problem is with `date`. It should be the moment when the payment has been made and this has no guarantee to be called in that moment. I believe it should accept `date` as parameter.

core/payment-driver/base/src/bus.rs
```diff
@@ -0,0 +1,120 @@
...
+        amount: details.amount.clone(),
+        sender: details.sender.clone(),
+        recipient: details.recipient.clone(),
+        order_ids: vec![order_id.to_string()],
+        confirmation: PaymentConfirmation { confirmation },
+    };
+
+    // Spawned because calling payment service while handling a call from payment service
+    // would result in a deadlock.
+    Arbiter::spawn(async move {
```
Wiezzel, line 114: That spawn is only necessary if `notify_payment` is called from `schedule_payment`. I believe that won't be the case if there's a separate thread with send loop.

core/payment-driver/base/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+
+// Used by the DummyDriver to have a 2 way conversion between details & confirmation
+pub fn to_confirmation(details: &PaymentDetails) -> Result<Vec<u8>, GenericError> {
```
Wiezzel, line 22: This is totally dummy-driver-specific. Do we really want it here? Other drivers will rather use transaction hash or ID as confirmation.

core/payment-driver/base/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+    Ok(serde_json::to_string(details)
+        .map_err(GenericError::new)?
+        .into_bytes())
+}
+
+pub fn from_confirmation(
```
Wiezzel, line 28: Ditto.

core/payment-driver/zksync/src/driver.rs
```diff
@@ -0,0 +1,131 @@
...
+
+// Local uses
+use crate::{zksync::wallet, DRIVER_NAME, PLATFORM_NAME};
+
+pub struct ZksyncDriver {
+    active_accounts: AccountsRc,
+}
+
+impl ZksyncDriver {
+    pub fn new(accounts: AccountsRc) -> Self {
```
Wiezzel, line 29: Any legit cases when we'll be passing non-empty `accounts` to the constructor?

core/payment-driver/zksync/src/zksync/mod.rs
```diff
@@ -0,0 +1,9 @@
+/*
+    Private mod to encapsulate all zksync logic, revealed from the `wallet`.
+*/
+
+pub mod wallet;
+
+mod faucet;
```
Wiezzel, line 7: Not quite sure if `faucet` belongs here but I'm not going to push it.

core/payment-driver/base/src/account.rs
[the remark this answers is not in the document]
maaktweluit: ok, thanks! will have to update my way of thinking to this..

I propose to leave it as is for now, we can always:
- Extend it to  get more purpose
- Refactor it away into a plain `HashMap`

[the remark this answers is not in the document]
Wiezzel: Yep, just leave for now and adjust if our needs change.

core/payment-driver/base/src/driver.rs
```diff
@@ -0,0 +1,72 @@
...
+    // -- Required to implement
+    async fn get_account_balance(
+        &self,
+        db: (),
+        caller: String,
+        msg: GetAccountBalance,
+    ) -> Result<BigDecimal, GenericError>;
+
+    // used to update the active accounts, see `accounts.rs` for more details
+    fn get_accounts(&self) -> AccountsRefMut;
```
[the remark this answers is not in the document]
maaktweluit: haha, it did not start as 2 lines when this fn was made :P

good thoughts, will update

core/payment-driver/base/src/utils.rs
maaktweluit: I like the from idea, can it also take a date?

If not i'll change this to take date as argument

core/payment-driver/base/src/bus.rs
maaktweluit: good catch, will update

core/payment-driver/base/src/utils.rs
Wiezzel: You can `impl From<(SchedulePayment, DateTime<Utc>)>` but I think it's ugly. Better just add a param.

core/payment-driver/base/src/utils.rs
maaktweluit: not really, looked too late at GNT driver to find out this was dummy stuff..

core/payment-driver/zksync/src/driver.rs
maaktweluit: When fetching accounts on startup it could.
But then this could also fetch the accounts from the bus

relates to https://golemproject.atlassian.net/browse/PAY-113

core/payment-driver/zksync/src/zksync/mod.rs
maaktweluit: why not? Just curious :)

Wiezzel: Well, I have this feeling like it's not really part of the zkSync infrastructure but and external service. But it's tightly coupled, so it's fine.

core/payment-driver/zksync/src/driver.rs
Wiezzel: I just thought it might have a param-less constructor. But also we can `#[derive(Default)]` (on `AccountsRc` as well then).

Wiezzel (approved):

PR #751, opened 2020-10-29
Provider Agent market refactored 

tworec:
It was the workaround for Centralized Market being restarted every night. In Decentralized Market it is not needed.


nieznanysprawiciel (changes requested):
It will be needed to resubscribe expired Offers, unless we decide to handle expiration on market level

tworec:
I've restored the original behavior of resubscribing when the `collect` returns 404. 
Now it only left a bare refactoring, but still, I see a value here, because I'm making it over 50 lines shorter.

nieznanysprawiciel (approved):

PR #723, opened 2020-10-22
Make integration tests part of CI workflow

zakaprov:
Resolves: https://github.com/golemfactory/yagna-integration/issues/225

The recently added `integration-test` workflow is currently not part of blocking checks on PRs. This makes integration tests part of the `rust` CI workflow. The current running time for these tests on Actions is ~4:30.

PR #784, opened 2020-11-13
Make zksync default, add gnt-driver feature flag

maaktweluit:


.github/workflows/integration-test.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: 'golemfactory/yagna-integration'
+          branch: 'mwu/multi-payment'
```
maaktweluit, line 18: TODO: Revert before merging, after merging `ya-int`

Cargo.toml
```diff
@@ -10,7 +10,7 @@ license = "GPL-3.0"
 edition = "2018"
 
 [features]
-default = ['gnt-driver', 'gftp/bin']
+default = ['zksync-driver', 'gftp/bin']
```
Wiezzel, line 13: I'd say we should include both in the default build.

maaktweluit: Great idea! updated

Wiezzel (dismissed):

.github/workflows/integration-test.yml
```diff
@@ -15,6 +15,7 @@ jobs:
...
           token: ${{ secrets.YAGNA_WORKFLOW_TOKEN }}
 
       - name: Configure python 3.7
```
zakaprov, line 21: This workflow is going to be removed in https://github.com/golemfactory/yagna/pull/723. If you'd like to get it running until then anyway, please also update the python version used here to 3.8.

maaktweluit: Good one, will do when removing the test branch

zakaprov (approved):

zakaprov (approved):

Wiezzel (approved):

PR #786, opened 2020-11-13
Update web3==0.13 and ethereum-tx-sign==3.0

maaktweluit:
- `updated web3==0.13` same as `zksync`
- `ethereum-tx-sign==3.0` using the same version of `ethereum-types`
- removed patch versions from `gnt/cargo.toml`
- moved custom logic from bdyyyzz private forks to `crate::eth_hacks`


core/payment-driver/gnt/src/lib.rs
```diff
@@ -8,6 +8,7 @@ pub mod migrations {
 
 mod dao;
 mod error;
+mod eth_hacks;
```
Wiezzel, line 11: Can we call it `eth_utils` not `eth_hacks`?

Wiezzel (approved):
Nice cleanup. :+1:

Wiezzel (approved):

PR #790, opened 2020-11-16
Set timeout for zksync payment init to 3 minutes

maaktweluit:


core/payment-driver/zksync/src/driver.rs
```diff
@@ -149,7 +150,7 @@ impl PaymentDriver for ZksyncDriver {
         //     return Err(GenericError::new("Can not init, account not active"));
         // }
 
-        wallet::init_wallet(&msg).await?;
+        wallet::init_wallet(&msg).timeout(Some(180)).await.map_err(GenericError::new)??;
```
Wiezzel, line 153: Maybe we can just use `std::time::Duration::from_secs(180)` here and delete the copied code?

maaktweluit: `.timeout()` does not exist without the copied code

maaktweluit: moved it to a neater place :)

Wiezzel (dismissed):

Wiezzel (approved):

PR #807, opened 2020-11-19
Fix/provider round cost

Wiezzel:
NGNT precision is 18 decimal places. Rounding up prevents payment errors.


maaktweluit (approved):
LGTM!

PR #824, opened 2020-11-24
Mwu/0.5.0 rc.4 to master

maaktweluit:
Merging back (more) quickly since @zakaprov needs it for his current work in `goth`


Wiezzel (approved):

tworec:
yeah, we need to be as much `up to` master as possible both ways

PR #830, opened 2020-11-25
Clean provider data dir on start / on demand

mfranciszkiewicz:
- removes files older than 30d on start
- `clean` command for manual disk clean up


agent/provider/src/dir.rs
```diff
@@ -0,0 +1,117 @@
...
+        "h" => 3600,
+        "d" | "" => 3600 * 24,
+        "w" => 3600 * 24 * 7,
+        "M" => 3600 * 24 * 30,
+        "y" => 3600 * 24 * 365,
+        u => bail!("Invalid unit: {}", u),
+    };
+
+    Ok(Duration::from_secs(base * mul))
+}
```
nieznanysprawiciel, line 117: Humantime https://docs.rs/humantime/2.0.1/humantime/ is what you need here ;)
It is already dependency of provider

nieznanysprawiciel (dismissed):

nieznanysprawiciel (approved):

PR #831, opened 2020-11-26
Merge SGX metrics fix present in master

mfranciszkiewicz:


exe-unit/src/service/metrics.rs
```diff
@@ -24,16 +24,68 @@ impl MetricsService {
         backlog_limit: Option<usize>,
         supervise_caps: bool,
     ) -> Result<Self, MetricError> {
-        let caps = move |ctx: &ExeUnitContext, id| match supervise_caps {
+        let caps = move |ctx: &ExeUnitContext, id: &str| match supervise_caps {
             true => ctx.agreement.usage_limits.get(id).cloned(),
             _ => None,
         };
-        let metrics: HashMap<String, MetricProvider> = vec![
+        let metrics = Self::metrics(&ctx, backlog_limit, caps);
```
tworec, line 31: can we expose those metrics to the grafana via `ya_metrics` crate?

prekucki (approved):

exe-unit/src/service/metrics.rs
mfranciszkiewicz: IMHO it would be better to do that in the provider agent

Issue #727, opened 2020-10-23
Terminate Agreement endpoint


nieznanysprawiciel:
- [x] Propagate terminate to Provider/Requestor
- [x] Change Agreement state to Terminated
- [x] Endpoint should have optional reason parameter (not used yet)
- [x] If can't connect to other party - return error and don't change state

- [x] exclude Demand-Offer pair from matching  if both Provider and Requestor have the same identity. 

This has been discussed with @stranger80 and enable us to detect which side of the Agreement called the `terminate` operation basing on identity from REST API.

PR #792, opened 2020-11-16
WIP TerminateAgreement endpoint

jiivan:
fixes: #727

PR #832, opened 2020-11-26
TerminateAgreement endpoint

jiivan:
fixes: #727

(just a branch name change from #792 to trigger market tests)


nieznanysprawiciel (changes requested):
I realized that we can't distinguish between Provider and Requestor Agreement in REST endpoints (as in GetAgreement), but here the problem is much bigger, because we have to know it, to fill `terminator` field in `AgreementTerminatedEvent`. Bu the way note, that we broke GetAgreement endpoint by adding AppSessionId.
I'm afraid we have to address these issues on specification level. @tworec

core/market/src/negotiation/common.rs
```diff
@@ -279,6 +291,144 @@ impl CommonBroker {
...
+                if agreement.requestor_id != caller {
+                    // Don't reveal, that we know this Agreement id.
+                    Err(RemoteAgreementError::NotFound(msg.agreement_id.clone()))?
+                }
+            }
+        }
+
+        dao.terminate(&msg.agreement_id, msg.reason, owner_type)
+            .await
+            .map_err(|_e| RemoteAgreementError::InternalError(err_msg.agreement_id.clone()))?;
```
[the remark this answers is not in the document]
jiivan: [from chat]: Error is not leaked, but it should be logged locally.

core/market/src/negotiation/common.rs
```diff
@@ -445,3 +595,12 @@ pub fn expect_state(
...
 }
+
+fn verify_reason(reason: Option<&String>) -> Result<(), ReasonError> {
+    if let Some(s) = reason {
+        Reason::from_json_reason(JsonReason {
+            json: serde_json::from_str(s)?,
+        })?;
+    };
+    Ok(())
+}
```
nieznanysprawiciel, line 606:
```suggestion
fn verify_reason(reason: Option<&String>) -> Result<(), ReasonError> {
    reason.map(|reason| serde_json::from_str::<Reason>(reason)?);
}
```

core/market/src/negotiation/common.rs
```diff
@@ -445,3 +596,12 @@ pub fn expect_state(
         AgreementState::Terminated => AgreementStateError::Terminated(agreement.id.clone()),
     })?
 }
+
+fn verify_reason(reason: Option<&String>) -> Result<(), ReasonError> {
+    if let Some(s) = reason {
+        Reason::from_json_reason(JsonReason {
+            json: serde_json::from_str(s)?,
+        })?;
+    };
```
nieznanysprawiciel, line 605:
```suggestion
    if let Some(s) = reason {
        serde_json::from_str::<Reason>(s)?
    };
```

core/market/src/negotiation/common.rs
```diff
@@ -431,7 +431,14 @@ impl CommonBroker {
 
         dao.terminate(&agreement_id, msg.reason, owner_type)
             .await
-            .map_err(|_e| RemoteAgreementError::InternalError(agreement_id.clone()))?;
+            .map_err(|e| {
+                log::info!(
```
nieznanysprawiciel, line 435: I would say it is even `warn`, because it would leave Provider and Requestor in inconsistent state

core/market/src/negotiation/common.rs
nieznanysprawiciel: Or even:
```
    Ok(if let Some(s) = reason {
        serde_json::from_str::<Reason>(s)?;
    })
```

core/market/src/negotiation/common.rs
```diff
@@ -279,6 +287,134 @@ impl CommonBroker {
...
+    pub async fn terminate_agreement(
+        &self,
+        id: Identity,
+        agreement_id: AgreementId,
+        reason: Option<String>,
+    ) -> Result<(), AgreementError> {
+        verify_reason(reason.as_ref())?;
+        let dao = self.db.as_dao::<AgreementDao>();
+        log::debug!(
+            "Getting agreement. id: {:?}, agrid: {}, reason: {:?}",
```
nieznanysprawiciel, line 300: Please don't display using debug if it is possible

nieznanysprawiciel (dismissed):

nieznanysprawiciel (approved):

PR #851, opened 2020-12-07
Backport exe-unit core count fix to release/v0.5

mfranciszkiewicz:
Contains cherry-picked commits from #844 (already merged to master)

### DIFF
--- a/exe-unit/src/runtime.rs
+++ b/exe-unit/src/runtime.rs
@@ -53,7 +53,7 @@ impl RuntimeArgs {
         let mut mem_gib = None;
         let mut storage_gib = None;
         if with_inf {
-            cpu_cores = agreement.infrastructure.get("cpu.cores").cloned();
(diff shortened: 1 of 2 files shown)


exe-unit/src/runtime.rs
```diff
@@ -53,7 +53,7 @@ impl RuntimeArgs {
         let mut mem_gib = None;
         let mut storage_gib = None;
         if with_inf {
-            cpu_cores = agreement.infrastructure.get("cpu.cores").cloned();
+            cpu_cores = agreement.infrastructure.get("cpu.threads").cloned();
```
tworec, line 56: we are mixing up cores and threads, but we have more important issues, so let it be

tworec (approved):

exe-unit/src/runtime.rs
mfranciszkiewicz: yes, requestors currently demand logical threads not physical cores

PR #853, opened 2020-12-07
Verify zksync transaction hash

maaktweluit:
- Remove dummy implementation of transaction verification
- Update payment service to allow larger amounts then on the invoice
- Add verify method on wallet, using the old v0.1 zksync API to query the data

### DIFF
--- a/core/payment/src/processor.rs
+++ b/core/payment/src/processor.rs
@@ -442,7 +442,7 @@ impl PaymentProcessor {
             .await??;
 
         // Verify if amount declared in message matches actual amount transferred on blockchain
-        if &details.amount != &payment.amount {
+        if &details.amount < &payment.amount {
             return VerifyPaymentError::amount(&details.amount, &payment.amount);
         }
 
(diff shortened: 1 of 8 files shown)


core/payment/src/processor.rs
```diff
@@ -292,14 +292,14 @@ impl PaymentProcessor {
             .await??;
 
         // Verify if amount declared in message matches actual amount transferred on blockchain
-        if &details.amount != &payment.amount {
+        if &details.amount < &payment.amount {
```
Wiezzel, line 295: I believe we should rather aim for having the exact amount sent in `SendPayment` message. Instead, we can put `<` here: https://github.com/golemfactory/yagna/blob/60ecce975d10a7fb9ace34540a02f8f54cc04947/core/payment/src/processor.rs#L53

Wiezzel (commented):
The hack really makes me :crying_cat_face: 
Also, I think we could send the exact payment amount in `SendPayment` message.

Wiezzel (approved):

Wiezzel (approved):

PR #855, opened 2020-12-09
Update notify_payment to send list of order_ids

maaktweluit:
- Fix unique constraint bug when multiple invoices are paid in the same batch
- Fix to not call `NotifyPayment` when the payment failed

### DIFF
--- a/core/payment-driver/zksync/src/driver.rs
+++ b/core/payment-driver/zksync/src/driver.rs
@@ -237,18 +237,32 @@ impl PaymentDriverCron for ZksyncDriver {
             // Check_tx returns None when the result is unknown
             if let Some(result) = wallet::check_tx(&tx_hash).await {
                 let payments = self.dao.transaction_confirmed(&tx.tx_id, result).await;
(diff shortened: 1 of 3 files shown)


core/payment-driver/zksync/src/driver.rs
```diff
@@ -241,18 +241,32 @@ impl PaymentDriverCron for ZksyncDriver {
...
+                if !result {
+                    log::warn!("Payment failed, will be re-tried.");
+                    continue;
                 }
+
+                let order_ids: Vec<String> = payments
+                    .clone()
+                    .into_iter()
+                    .map(|payment| payment.order_id)
+                    .collect();
```
Wiezzel, line 253: We could clone `order_id`s only instead of the whole `payments`.
```suggestion
                let order_ids = payments
                    .iter()
                    .map(|payment| payment.order_id.clone())
                    .collect();
```

Wiezzel (approved):
LGTM, I only have one tiny optimization suggestion.

Wiezzel (approved):

PR #858, opened 2020-12-10
Reason simplification

tworec:
depends on https://github.com/golemfactory/ya-client/pull/71


nieznanysprawiciel (approved):
Is there easy way to convert from `Reason` to custom structure?

PR #861, opened 2020-12-10
Mwu/pay event new params

maaktweluit:
First update to input params, to review before applying the same input to all other calls.

Required some updates in `ya-client`
https://github.com/golemfactory/ya-client/pull/73

### DIFF
--- a/core/payment/src/api.rs
+++ b/core/payment/src/api.rs
@@ -78,7 +78,7 @@ pub struct EventParams {
 #[derive(Deserialize)]
 pub struct FilterParams {
     #[serde(rename = "maxItems", default)]
-    pub max_items: Option<i64>,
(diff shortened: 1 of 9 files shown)


core/payment/src/api.rs
```diff
@@ -75,6 +75,14 @@ pub struct EventParams {
     pub later_than: Option<DateTime<Utc>>,
 }
 
+#[derive(Deserialize)]
+pub struct FilterParams {
+    #[serde(rename = "maxItems", default)]
+    pub max_items: Option<i64>,
```
Wiezzel, line 81: I'd rather go with `u64` since negative values do not make sense here.

Wiezzel (approved):

PR #871, opened 2020-12-14
Payment API: filtering by app_session_id

Wiezzel:
Added app_session_id filed to pay_agreement DB model. Enabled filtering all events and payments by app_ssession_id.

### DIFF
--- a/core/payment/src/dao/debit_note_event.rs
+++ b/core/payment/src/dao/debit_note_event.rs
@@ -1,9 +1,12 @@
 use crate::error::DbResult;
 use crate::models::debit_note_event::{ReadObj, WriteObj};
+use crate::schema::pay_activity::dsl as activity_dsl;
+use crate::schema::pay_agreement::dsl as agreement_dsl;
+use crate::schema::pay_debit_note::dsl as debit_note_dsl;
 use crate::schema::pay_debit_note_event::dsl;
 use crate::schema::pay_event_type::dsl as event_type_dsl;
 use chrono::NaiveDateTime;
-use diesel::{ExpressionMethods, QueryDsl, RunQueryDsl};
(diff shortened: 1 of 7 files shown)


core/payment/src/dao/debit_note_event.rs
```diff
@@ -59,18 +62,36 @@ impl<'c> DebitNoteEventDao<'c> {
         node_id: NodeId,
         after_timestamp: Option<NaiveDateTime>,
         max_events: Option<u32>,
-        _app_session_id: Option<String>,
+        app_session_id: Option<String>,
     ) -> DbResult<Vec<DebitNoteEvent>> {
         readonly_transaction(self.pool, move |conn| {
             let mut query = dsl::pay_debit_note_event
                 .inner_join(event_type_dsl::pay_event_type)
+                .inner_join(
```
maaktweluit, line 70: Seems the join is not always required, can we join inside the `if Some( .. ) ..` block?

maaktweluit (changes requested):
Great start!

Some questions / thoughts:
- Commented once on a join to be in `let Some( .. ) ..`, this could apply to all added joins.
- Soo many joins...
-- maybe we can solve this with an extra column on invoice, debit_note OR event
-- maybe we can solve this by adding a new view in the database
- Do we need to preserve data on the down migration? For Allocations i just cleared the table and didnt care about the data.. We should either:
-- remove hacks from agreement down script
-- add data persistance to allocations

core/payment/src/dao/debit_note_event.rs
maaktweluit: not possible :(

maaktweluit (approved):

maaktweluit:
As discussed we made a new issue to look into the views: https://golemproject.atlassian.net/browse/PAY-181

Approved!

PR #873, opened 2020-12-15
Respawn proposal after termination

nieznanysprawiciel:
resolves: #728


core/market/src/db/dao/negotiation_events.rs
```diff
@@ -79,7 +80,7 @@ impl<'c> NegotiationEventsDao<'c> {
 
             let events = dsl::market_negotiation_event
                 .filter(dsl::subscription_id.eq(&subscription_id))
-                .order_by(dsl::timestamp.asc())
+                .order_by(sql::<sql_types::Bool>("RANDOM()"))
```
tworec, line 83: kinda controversial, because we will be hiding `AgreementEvent`s, which should have priority.
We should only shuffle `ProposalEvent`s...

core/market/src/negotiation/common.rs
```diff
@@ -333,6 +336,23 @@ impl CommonBroker {
             &agreement.id,
             reason.display(),
         );
+
+        // If we are Requestor we would like to be able to negotiate with the same
+        // provider for the second time, so we must generate new Proposal for him.
+        // Note: Regeneration failure isn't propagated to Requestor, because termination
+        // succeeded at this point.
+        if let OwnerType::Requestor = agreement.id.owner() {
+            self.regenerate_proposal(&agreement)
```
tworec, line 345: it should be spawned to the side task, not to hold terminate agreement op, which should be as fast as possible

whats more, whole this logic should be enclosed within `regenerate_proposal`

core/market/src/negotiation/common.rs
```diff
@@ -333,6 +336,23 @@ impl CommonBroker {
...
+                .await
+                .map_err(|e| {
+                    log::warn!(
+                        "Failed to regenerate Proposal after Agreement [{}] termination. {}",
+                        &agreement.id,
+                        e
+                    )
+                })
```
tworec, line 353: this logic should be enclosed within `regenerate_proposal`

core/market/tests/test_initial_proposal.rs
```diff
@@ -548,3 +553,98 @@ async fn test_counter_initial_proposal() {
...
+    let agreement_id = info.r_agreement;
+
+    let timestamp = Utc::now();
+
+    req_market
+        .terminate_agreement(req_id, agreement_id.clone(), Some(gen_reason("Success")))
+        .await
+        .unwrap();
+
+    let proposal = requestor::query_proposal(&req_market, &info.negotiation.demand_id, 4)
```
tworec, line 589: we should also check that there is no proposal on the provider side

core/market/tests/test_proposal.rs
```diff
@@ -78,3 +82,68 @@ async fn test_get_proposal_not_found() {
...
+        .filter_map(|event| match event {
+            RequestorEvent::ProposalEvent { proposal, .. } => Some(proposal.issuer_id),
+            _ => None,
+        })
+        .collect::<Vec<NodeId>>();
+
+    // If proposals were really shuffled, we expect incoming order to be different
+    // from initialization order.
+    assert_eq!(incoming_ids.len(), ids.len());
+    assert_ne!(incoming_ids, ids);
```
tworec, line 148: buggy. it will sometimes match ;)

tworec (changes requested):

tworec:
pls look at https://github.com/golemfactory/yagna/pull/930

core/market/src/negotiation/common.rs
nieznanysprawiciel: It wouldn't be elegant. Check `regenerate_proposal` internals. I would have to add this warning in 3 places or create wrapper function that does only this

core/market/src/db/dao/negotiation_events.rs
nieznanysprawiciel: We decided to create issue for later:  #932

core/market/tests/test_proposal.rs
nieznanysprawiciel: I can increase number of Offers to minimize probability ;)

tworec: ;) that's even worse, leave for now until first fail

core/market/src/negotiation/common.rs
tworec: ok, let it be

core/market/src/negotiation/common.rs
tworec: I see you've addressed it

tworec (approved):

PR #875, opened 2020-12-15
Use poetry to set up goth

zakaprov:


.github/workflows/rust.yml
```diff
@@ -162,14 +162,16 @@ jobs:
...
+          poetry-version: 1.1.4
 
-      - name: Print installed package versions
-        run: pip freeze
+      - name: Install dependencies
+        run: poetry install --no-root
 
       - name: Log in to GitHub Docker repository
-        run: echo ${{ secrets.YAGNA_WORKFLOW_TOKEN }} | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+        run: echo ${{ secrets.GITHUB_TOKEN }} | docker login docker.pkg.github.com -u ${{github.actor}} --password-stdin
```
shadeofblue, line 174: this is the only change I don't understand -> why we're suddenly changing the name of the token? ...

shadeofblue (approved):
looks okay

.github/workflows/rust.yml
zakaprov: It's not the name, but the token itself. Since `goth` is now public we no longer need `YAGNA_WORKFLOW_TOKEN`, which is a manually generated token with some extra permissions. `GITHUB_TOKEN` is auto-generated by Actions, only has basic permissions (as described [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token)) and gets revoked at the end of a workflow run.

PR #877, opened 2020-12-15
Persist data during event-api/down migration

maaktweluit:


core/payment/migrations/2020-12-11-123456_event_api/down.sql
```diff
@@ -13,6 +14,17 @@ CREATE TABLE pay_allocation(
...
 
+INSERT INTO pay_allocation_tmp(id, owner_id, payment_platform, address, total_amount, spent_amount, remaining_amount, timeout, make_deposit, released)
+SELECT id, owner_id, payment_platform, address, total_amount, spent_amount, remaining_amount, timeout, make_deposit, released FROM pay_allocation;
+
+
+DROP TABLE pay_allocation;
+
+ALTER TABLE pay_allocation_tmp RENAME TO pay_allocation;
+
+PRAGMA foreign_keys=on;
```
Wiezzel, line 25: You forgot to disable it first.

And, actually this could be done once for the hole migration.

Wiezzel (changes requested):

Wiezzel (approved):

core/payment/migrations/2020-12-11-123456_event_api/down.sql
maaktweluit: good catch! forgot a test with data...

Updated the file and ran test for empty AND filled DB

PR #887, opened 2020-12-16
Run payment cron jobs more often

maaktweluit:
reduce cron intervals from 30->10 (payment) and 10->5 (confirmation)

This is done so all invoices are handled in the 30 seconds of wait in `ya*api`

Also lowered most log levels to not spam the logs.

Resolves https://github.com/golemfactory/yapapi/issues/190


Wiezzel (dismissed):
:+1:

Wiezzel (approved):

PR #879, opened 2020-12-15
Payment API: examples adjusted for event API

Wiezzel:
* Removed remaining deprecated references to `PaymentProviderApi` and `PaymentRequestorApi`.
* Introduced `--app-session-id` parameter for `payment_api`, `invoice_flow`, and `debit_note_flow` examples in order to test event filtering.
* Changed the way provider and requestor sub-services are used to avoid the hack with `YAGNA_PAYMENT_URL`.

----
Testing instructions:

1. Run `payment_api` example with `--app-session-id=whatever`.
2. Try running `debit_note_flow` and `invoice_flow` with `--app-session-id=somthing-else` (should fail on event unwrapping :x: ).
3. Run `debit_note_flow` and `invoice_flow` with `--app-session-id=whatever` (should work :heavy_check_mark: ).
4. Run all other examples (should work :heavy_check_mark: ).

PR #888, opened 2020-12-16
Payment: DB views for querying events

Wiezzel:
Simplified and optimized event queries by introducing DB views.

----
Testing instructions:

Just repeat scenarios from #879

### DIFF
--- a/core/payment/src/schema.rs
+++ b/core/payment/src/schema.rs
@@ -88,6 +88,17 @@ table! {
 }
 
 table! {
+    pay_debit_note_event_read (debit_note_id, event_type) {
+        debit_note_id -> Text,
+        owner_id -> Text,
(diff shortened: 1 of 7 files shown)


core/payment/src/schema.rs
```diff
@@ -87,6 +87,17 @@ table! {
     }
 }
 
+table! {
```
maaktweluit, line 90: table? same macro for views?

maaktweluit (approved):
LGTM!

Running some tests now

core/payment/src/schema.rs
Wiezzel: Yep, there's no separate macro for views. Also, as far as selects are concerned, there isn't much difference between these two.

PR #895, opened 2020-12-21
logging to file

### LINKED ISSUE #882: add file logging to `yagna` and `ya-provider`
look at logger used by `exe-unit` which is capable of multiplying logs stream to both: console and file

etam:
closes #882


utils/path/src/data_dir.rs
```diff
@@ -42,6 +42,6 @@ impl FromStr for DataDir {
 
 impl Display for DataDir {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
-        write!(f, "{:?}", self.0)
+        write!(f, "{}", self.0.to_string_lossy())
```
etam, line 45: This is required, because otherwise flexi logger tries to create a path like `"/home/user/.local/share/yagna"/yagna.log` and those extra quotes are causing problems.

core/serv/src/main.rs
```diff
@@ -410,25 +414,82 @@ async fn me(id: Identity) -> impl Responder {
...
+            .print_message()
+            .duplicate_to_stderr(Duplicate::All);
+    }
+    logger = logger
+        .adaptive_format_for_stderr(AdaptiveFormat::Custom(log_format, log_format_color))
+        .set_palette("9;11;2;7;8".to_string());
+
+    logger.start()?;
+    Ok(())
+}
```
nieznanysprawiciel, line 483: Can we extract common part somewhere?

nieznanysprawiciel (approved):

core/serv/src/main.rs
```diff
@@ -410,25 +414,82 @@ async fn me(id: Identity) -> impl Responder {
...
+            DateTime::<Utc>::from(*now.now()).to_rfc3339_opts(SecondsFormat::Secs, true),
+            style(level, level),
+            record.module_path().unwrap_or("<unnamed>"),
+            record.args()
+        )
+    }
+
+    let log_spec = LogSpecification::env_or_parse(log_level)?;
+    let log_spec = LogSpecBuilder::from_module_filters(log_spec.module_filters())
+        .module("actix_web::middleware::logger", log::LevelFilter::Warn)
```
tworec, line 456: WMiddleware should be @ `warn` level only when the default setting is used.
Otherwise it should be taken from RUST_LOG.
Pls remove this line, and revert the original workaround below.

core/serv/src/main.rs
```diff
@@ -410,25 +414,82 @@ async fn me(id: Identity) -> impl Responder {
...
     dotenv::dotenv().ok();
     let args: CliArgs = CliArgs::from_args();
 
-    env::set_var(
-        "RUST_LOG",
-        env::var("RUST_LOG").unwrap_or(format!(
-            "{},actix_web::middleware::logger=warn",
-            args.log_level()
-        )),
-    );
```
tworec, line 424: pls revert this workaoround

core/serv/src/main.rs
```diff
@@ -410,25 +414,82 @@ async fn me(id: Identity) -> impl Responder {
...
+    }
+
+    let log_spec = LogSpecification::env_or_parse(log_level)?;
+    let log_spec = LogSpecBuilder::from_module_filters(log_spec.module_filters())
+        .module("actix_web::middleware::logger", log::LevelFilter::Warn)
+        .module("actix_http::response", log::LevelFilter::Off)
+        .module("tokio_core", log::LevelFilter::Info)
+        .module("tokio_reactor", log::LevelFilter::Info)
+        .module("hyper", log::LevelFilter::Info)
+        .module("web3", log::LevelFilter::Info)
```
tworec, line 461: hard set `h2` and `reqwest` on `info` also

exe-unit/Cargo.toml
```diff
@@ -42,7 +42,7 @@ async-trait = "0.1.24"
 chrono = "0.4.10"
 derivative = "2.1"
 dotenv = "0.15.0"
-flexi_logger = { version = "0.15", features = ["colors"] }
+flexi_logger = { version = "0.16", features = ["colors"] }
```
tworec, line 45: we cannot merge it wo @mfranciszkiewicz or @prekucki 
I'd revert this

tworec (changes requested):

tworec (approved):

utils/path/src/data_dir.rs
jiivan: This comment was very helpful for me! Could you put this comment in source also?

tworec:
It would be intuitive to use `data_dir` as a default place to store logs. 

Currently, `log_dir` is separate setting which only shares default value with `data_dir`

utils/path/src/data_dir.rs
etam: yes

core/serv/src/main.rs
etam: yes

core/serv/src/main.rs
```diff
@@ -82,9 +83,19 @@ struct CliArgs {
     json: bool,
 
     /// Log verbosity level
-    #[structopt(long, default_value = "info")]
+    #[structopt(long, default_value = "info", set = clap::ArgSettings::Global)]
```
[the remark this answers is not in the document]
etam: On a second thought: other commands might want to do some logging too. Maybe not to the same file, or not to a file at all, but still some debug info might be needed sometimes).

etam:
> It would be intuitive to use `data_dir` as a default place to store logs.
> 
> Currently, `log_dir` is separate setting which only shares default value with `data_dir`.
> 
> https://www.notion.so/golemnetwork/Event-API-testing-session-81e92bf06e5e49f8a65eea9f66d6a83b#e97d8f778b5340c7b6d822acebc2fff9

Done

core/serv/src/main.rs
```diff
@@ -81,10 +83,6 @@ struct CliArgs {
     #[structopt(long, set = clap::ArgSettings::Global)]
     json: bool,
 
-    /// Log verbosity level
-    #[structopt(long, default_value = "info")]
-    log_level: String,
-
```
tworec, line 87: 👍

core/serv/src/main.rs
```diff
@@ -262,7 +253,10 @@ enum CliCommand {
 impl CliCommand {
     pub async fn run_command(self, ctx: &CliCtx) -> Result<CommandOutput> {
         match self {
-            CliCommand::Commands(command) => command.run_command(ctx).await,
+            CliCommand::Commands(command) => {
+                set_logging("warn", &Path::new(""), &vec![])?;
```
tworec, line 257: ok, let it be. Just  with `info` as a default log level

tworec (dismissed):

core/serv/src/main.rs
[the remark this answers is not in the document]
etam: It is the case (not "when it will be") and it is by default logging to stderr.

core/serv/src/main.rs
etam: I used warn, because on info we get unwanted output. For example:
```
$ yagna id list
[2021-01-07T08:56:33Z INFO  ya_service_bus::remote_router] trying to connect to: 127.0.0.1:7464
[2021-01-07T08:56:33Z INFO  ya_service_bus::connection] started connection to gsb
┌───────────┬──────────┬─────────┬──────────────────────────────────────────────┐
│  default  │  locked  │  alias  │  address                                     │
├───────────┼──────────┼─────────┼──────────────────────────────────────────────┤
│  X        │          │         │  0x0a50cb00895929cf40a4c2e323989dbaa72c801b  │
└───────────┴──────────┴─────────┴──────────────────────────────────────────────┘
[2021-01-07T08:56:33Z INFO  ya_service_bus::connection] stopped connection to gsb
```

tworec (approved):

PR #869, opened 2020-12-14
AgreementEvents with common fields - compatibility with updated ya-client

nieznanysprawiciel:
resolves: #868
related to changes in ya-client: https://github.com/golemfactory/ya-client/pull/79

Remember!!! Update referenced branch on ya-client to master, before merging this PR

PR #896, opened 2020-12-21
[mkt] negotiation metrics fix

tworec:
resolves https://github.com/golemfactory/yagna/pull/869/files/fa24515e8a6b85360ca5ebd38f50af427e8c747c#r546956515


nieznanysprawiciel (changes requested):
There is the same problem as we had with RejectProposal. We probably should have 4 metrics for 4 pairs (Owner, Terminator).
My initial intention was to pass owner, not terminator here, so even if we don't introduce new metrics, these changes are invalid

PR #902, opened 2021-01-05
Event API

mfranciszkiewicz:


Wiezzel (dismissed):

nieznanysprawiciel (approved):
I assumed, that all changes in `event-api/master` were reviewed, before merging to this branch.

tworec:
> I assumed, that all changes in `event-api/master` were reviewed, before merging to this branch.

not true ;)

Cargo.toml
```diff
@@ -171,8 +172,11 @@ ya-sb-router = { path = "service-bus/router" }
 ya-sb-util = { path = "service-bus/util" }
 
 ## CLIENT
-ya-client = { git = "https://github.com/golemfactory/ya-client.git", rev = "e68bbd90c903c6f74f77fddb80a432d20e4e2b2a"}
-ya-client-model = { git = "https://github.com/golemfactory/ya-client.git", rev = "e68bbd90c903c6f74f77fddb80a432d20e4e2b2a"}
+ya-client = { git = "https://github.com/golemfactory/ya-client.git", branch = "event-api/master"}
```
tworec, line 175: we do not want to merge it like this to the master branch!

tworec (changes requested):
need to point specific ya-client gitrev

core/activity/src/common.rs
```diff
@@ -39,10 +39,15 @@ pub struct QueryTimeoutCommandIndex {
 }
 
 #[derive(Deserialize, Debug)]
-pub struct QueryTimeoutMaxEvents {
+pub struct QueryEvents {
     /// number of milliseconds to wait
-    #[serde(rename = "timeout", default = "default_query_timeout")]
-    pub timeout: Option<f32>,
+    #[serde(rename = "appSessionId")]
```
tworec, line 44: above comment is irrelevant
cc @mfranciszkiewicz @nieznanysprawiciel

tworec: forthermore I'd use `#[serde(rename_all = "camelCase")]` which is not only more concise, but also less error-prone

tworec: might be adressed in separate PR

mfranciszkiewicz: afair this code was auto-generated and only slightly modified later on

core/activity/src/dao/event.rs
```diff
@@ -99,65 +106,66 @@ impl<'c> EventDao<'c> {
...
                 .order(dsl_event::event_date.asc())
-                .limit(limit as i64)
+                .limit(limit)
                 .load::<Event>(conn)
                 .optional()?;
 
-            if let Some(r) = &results {
-                let ids = r.iter().map(|event| event.id).collect::<Vec<_>>();
-                if !ids.is_empty() {
-                    diesel::delete(dsl_event::activity_event.filter(dsl_event::id.eq_any(ids)))
```
tworec, line 135: we need some cleanup mechanism for those events

Wiezzel (dismissed):

core/activity/src/dao/event.rs
```diff
@@ -99,65 +106,66 @@ impl<'c> EventDao<'c> {
...
 
         log::trace!("get_events: starting db query");
-        let identity_id = identity_id.to_owned();
-
-        do_with_transaction(self.pool, move |conn| {
-            let results: Option<Vec<Event>> = dsl_event::activity_event
+        readonly_transaction(self.pool, move |conn| {
+            let mut query = dsl_event::activity_event
                 .inner_join(schema::activity::table)
-                .filter(dsl_event::identity_id.eq(identity_id))
```
tworec, line 119: is it on purpose to skip identity filtering here? it seems that now every identity can sniff all events, even those not intended for it

core/market/src/db/dao/negotiation_events.rs
```diff
@@ -40,17 +40,28 @@ impl<'c> AsDao<'c> for NegotiationEventsDao<'c> {
 }
 
 impl<'c> NegotiationEventsDao<'c> {
-    pub async fn add_proposal_event(
+    pub async fn add_proposal_event(&self, proposal: &Proposal, role: Owner) -> DbResult<()> {
+        let event = MarketEvent::from_proposal(proposal, role);
```
tworec, line 44: we need to create it under the transaction and with a timestamp on the rust code side, sam as we did with agreement events

core/market/src/db/dao/negotiation_events.rs
```diff
@@ -40,17 +40,28 @@ impl<'c> AsDao<'c> for NegotiationEventsDao<'c> {
...
+
+    pub async fn add_proposal_rejected_event(
         &self,
-        proposal: Proposal,
-        owner: OwnerType,
-    ) -> DbResult<Proposal> {
+        proposal: &Proposal,
+        reason: Option<String>,
+    ) -> DbResult<()> {
+        let event = MarketEvent::proposal_rejected(proposal, reason);
```
tworec, line 59: same here

core/activity/src/dao/event.rs
mfranciszkiewicz: reverted

core/activity/src/dao/event.rs
mfranciszkiewicz: Do you mean events in general? The count of activity events is going to be few orders of magnitude lower than payments / market

core/market/src/negotiation/common.rs
```diff
@@ -125,12 +133,56 @@ impl CommonBroker {
...
+        let proposal = self.get_proposal(subs_id.clone(), proposal_id).await?;
+
+        self.validate_proposal(&proposal, caller_id, caller_role)
+            .await?;
+
+        if proposal.body.issuer == Issuer::Us && proposal_id.owner() == caller_role {
+            let e = ProposalValidationError::OwnProposal(proposal.body.id.clone());
+            log::warn!("{}", e);
+            Err(e)?;
+        }
```
tworec, line 159: I've written it myself, but it smells to me. This should be part of `validate_proposal`

tworec: addressed in #914

core/market/src/negotiation/common.rs
```diff
@@ -316,53 +366,48 @@ impl CommonBroker {
...
             .await
             .map_err(|e| AgreementError::UpdateState((&agreement.id).clone(), e))?;
-
         self.notify_agreement(&agreement).await;
 
         inc_terminate_metrics(&reason, agreement.id.owner());
         log::info!(
-            "Agent {} terminated Agreement [{}]. Reason: {}",
+            "{:?} {} terminated Agreement [{}]. Reason: {}",
+            agreement.id.owner(),
```
tworec, line 379: minor: we are not sure here that the owner is terminating

core/payment/src/api.rs
```diff
@@ -5,22 +5,27 @@ use ya_client_model::payment::PAYMENT_API_PATH;
...
 }
 
 pub fn web_scope(db: &DbExecutor) -> Scope {
     Scope::new(PAYMENT_API_PATH)
         .data(db.clone())
-        .service(provider_scope())
-        .service(requestor_scope())
+        .service(api_scope(Scope::new("")))
+    // TODO: TEST
+    // Scope::new(PAYMENT_API_PATH).extend(api_scope).data(db.clone())
```
tworec, line 28: both defaults below should be 5 seconds according to actual spec:
- pollTmeout: https://github.com/golemfactory/ya-client/blob/87f106498a7a7b1519d82d0981fc21b260a89cf2/specs/common.yaml#L18
- ackTimeout: https://github.com/golemfactory/ya-client/blob/87f106498a7a7b1519d82d0981fc21b260a89cf2/specs/common.yaml#L32

core/activity/src/dao/event.rs
```diff
@@ -64,14 +64,14 @@ impl<'c> EventDao<'c> {
...
 
-        let now = Utc::now().naive_utc();
         log::trace!("creating event_type: {:?}", event_type);
 
         let app_session_id = app_session_id.to_owned();
         let activity_id = activity_id.to_owned();
         let identity_id = identity_id.to_owned();
 
         do_with_transaction(self.pool, move |conn| {
+            let now = Utc::now().naive_utc();
```
tworec, line 74: 👍

tworec (approved):
minor issues only, all of which can and should be addressed in separate PRs

core/activity/src/dao/event.rs
tworec: I think like, at most one order of magnitude.
But yes, in general.
We have a cleanup mechanism in mkt

mfranciszkiewicz: | I think like, at most one order of magnitude.

E.g. debit note events may significantly outnumber create / destroy activity events for long running services.

| we need some cleanup mechanism for those events

We haven't established a common cleanup policy and event removal isn't a part of the spec (for whatever reason). I'll abstain from implementing a heuristic solution for now - there is simply no need to do that now and the number of events won't affect the database in a significant way

tworec: True dat!
I was not clear in my intention here. I just wanted to express that such cleanup is needed, but not pushing to implement it right away. 👍

tworec: referenced it #918

core/market/src/db/dao/negotiation_events.rs
tworec: referenced in #917

core/payment/src/api.rs
tworec: referenced in #919

PR #909, opened 2021-01-11
zksync-driver withdrawal

pnowosie:
PAY-65: Withdrawal from zksync-driver's wallet to L1.

This is example that eventually refactors to payment's withdraw command.

- Run without parameters reuses preconfigured account to quickly withdraw default 5 NGNT from [0xfaca...91da](https://rinkeby.zkscan.io/explorer/accounts/0xfaca77397cede4f059a101093f076c99568d91da)
```
> cargo run --example withdrawal  
    Finished dev [unoptimized + debuginfo] target(s) in 0.24s
     Running `/home/pnowosie/Proj/yagna/target/debug/examples/withdrawal`
[2021-01-14T08:13:04Z INFO  withdrawal] Private key: 312776bb901c426cb62238db9015c100948534dea42f9fa1591eff4beb35cc13
    Account address 0xfaca77397cede4f059a101093f076c99568d91da
[2021-01-14T08:13:04Z INFO  withdrawal] Funding an account
[2021-01-14T08:13:04Z INFO  withdrawal] Creating wallet
[2021-01-14T08:13:07Z INFO  withdrawal] Deposit successful 68.899914443 NGNT available
[2021-01-14T08:13:07Z INFO  withdrawal] Obtaining withdrawal fee
[2021-01-14T08:13:08Z INFO  withdrawal] Withdrawal transaction fee 1.02400
[2021-01-14T08:13:08Z INFO  withdrawal] Withdrawal of 5.00000 NGNT started
[2021-01-14T08:13:09Z INFO  withdrawal] Waiting for receipt - this takes LOOONG to complete...
    Check it here: https://rinkeby.zkscan.io/explorer/accounts/0xfaca77397cede4f059a101093f076c99568d91da
...
```

- It allows to specify `amount` and use of randomly generated address
```
>  cargo run --example withdrawal -- --gen-key --amount=100
```
Here is an example of zeroing the balance of the account [[zkscan](https://rinkeby.zkscan.io/explorer/accounts/0xe3083125822420b5ec68056eaaa93b35d80a6db4)] [[etherscan](https://rinkeby.etherscan.io/address/0xe3083125822420b5ec68056eaaa93b35d80a6db4#tokentxns)]

**Better (earlier) error handling**
Trying to withdraw too much [[zkscan](https://rinkeby.zkscan.io/explorer/accounts/0x8204b73c22f4b2138a7b3849580d13f045ce698b)]
```
[2021-01-18T10:08:19Z INFO  withdrawal] Waiting for receipt - this takes LOOONG to complete...
[2021-01-18T10:08:19Z INFO  withdrawal] Check it here: https://rinkeby.zkscan.io/explorer/accounts/0x8204b73c22f4b2138a7b3849580d13f045ce698b
[2021-01-18T10:08:20Z WARN  withdrawal] Withdraw has failed. Reason: Some("Not enough balance")
```


core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+    let args: Args = Args::from_args();
+    let private_key = if args.genkey {
+        debug!("Using randomly generated key");
+        H256::random()
+    } else {
+        debug!("Using hardcoded key");
+        let mut bytes = [0u8; 32];
+        hex::decode_to_slice(PRIVATE_KEY, &mut bytes)
+            .expect("Cannot decode bytes from hex-encoded PK");
+        H256::from(bytes)
```
maaktweluit, line 39: there is no need for a `mut` here, you can:
```suggestion
        H256::from_str(PRIVATE_KEY).expect("Cannot decode bytes from hex-encoded PK")
```

Just from the top of my head, untested code here.
You also need `use std::str::FromStr;` at the top of the file to make this work

core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+    let pk_hex: String = private_key.encode_hex();
```
maaktweluit, line 42: on the previous line you went from `String` to `bytes[]`, now back to `String`..

is `private_key` used elsewhere? otherwise i think its better to keep it as hex_string

core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+    let signer = PrivateKeySigner::new(private_key);
+    let address = signer.get_address().await?;
+    let addr_hex = format!("0x{}", address.encode_hex::<String>());
+    info!("Private key: {}\nAccount address {}", pk_hex, addr_hex);
```
maaktweluit, line 46: I would try to avoid `\n` in log lines, unfortunately its used more often in yagna, but logs are cleaner without it.

We have logging guidelines that propose a log message structure like this:
```
This is what happened. key1={value1}, key2={value2}
```

Source: https://github.com/golemfactory/golem-internal/wiki/Logging-guidelines#log-message-structure

Because this is just an example it does not matter that much, just FYI :)

core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+    info!("Creating wallet");
+    let provider = Provider::new(Network::Rinkeby);
+    let cred = WalletCredentials::from_eth_signer(address, signer, Network::Rinkeby).await?;
+    let wallet = Wallet::new(provider, cred).await?;
+
+    if wallet.is_signing_key_set().await? == false {
+        info!("Unlocking account");
+        let unlock = wallet
+            .start_change_pubkey()
+            .fee_token("GNT")?
```
maaktweluit, line 61: there is a const for this "Magic string"

core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+            .send()
+            .await?;
+        debug!("unlock={:?}", unlock);
+        unlock.wait_for_commit().await?;
+    }
+
+    let balance = wallet.get_balance(BlockStatus::Committed, TOKEN).await?;
+    info!(
+        "Deposit successful {} NGNT available",
```
maaktweluit, line 70: NGNT is a deprecated name now, please use GLM ( or tGLM )

core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+        utils::big_uint_to_big_dec(balance.clone())
+    );
+
+    info!("Obtaining withdrawal fee");
+    let amount = utils::big_dec_to_big_uint(BigDecimal::from(args.amount))?;
```
maaktweluit, line 75: its dangerous ( and limiting ) to use `f64` -> `BigDecimal`

I would recommend to use a string as input param ( for both `BigDecimal`'s and `big_int`'s to support higher precision

core/payment-driver/zksync/src/zksync/mod.rs
```diff
@@ -4,6 +4,6 @@
 
 pub mod wallet;
 
-mod faucet;
+pub mod faucet;
```
maaktweluit, line 7: faucet was private on purpose, it was "exposed" through the `wallet.rs`

Maybe you can use wallet::init_wallet(msg)?

Not 100% sure about this one, since you would need to make the message that is un-related to the purpose of this example

core/payment-driver/zksync/examples/withdrawal.rs
```diff
@@ -0,0 +1,111 @@
...
+        .amount(withdraw_amount)
+        .to(address)
+        .send()
+        .await?;
+
+    debug!("Withdraw: {:?}", withdraw_handle);
+
+    info!(
+        "Waiting for receipt - this takes LOOONG to complete...\nCheck it here: {}",
+        format!("https://rinkeby.zkscan.io/explorer/accounts/{}", addr_hex)
```
maaktweluit, line 105: info!() already does a format, do we need a second one?

maaktweluit (commented):
Great start!

Quite a bit of small comments, nothing breaking..

Will approve on reply / fix :)

core/payment-driver/zksync/examples/withdrawal.rs
pnowosie: Excellent!
(Believe in yourself more - this was exactly what I need ;))

core/payment-driver/zksync/examples/withdrawal.rs
pnowosie: So it's only needed if one chose random option (`--gen-key`), to display the key (just if needed)...
Because it's easy to add I will remove this code now.

core/payment-driver/zksync/examples/withdrawal.rs
pnowosie: Note to self: TODO: Copy link to guidelines

core/payment-driver/zksync/examples/withdrawal.rs
pnowosie: Great catch! Fixed.

core/payment-driver/zksync/src/zksync/mod.rs
pnowosie: I'll try this... leaving for the another commit

Wiezzel: `wallet::init_wallet` is not an option here, because it creates a wallet using `YagnaEthSigner` instead of `PrivateKeySigner`.

Wiezzel (commented):
Could you please extract the withdrawal logic (lines 64–109) into a function and put it in `wallet` module? You can then import that function in the example.

core/payment-driver/zksync/src/zksync/mod.rs
pnowosie: Yeah - I just wanted to respond that it's using [wallet::get](https://github.com/golemfactory/yagna/blob/dde6dc99c6150d657544f4bb6467d95998f1285a/core/payment-driver/zksync/src/zksync/wallet.rs#L68) which I shoud stay away from ;)

maaktweluit: Thanks for trying, lets leave it as is :)

core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -184,3 +185,53 @@ async fn unlock_wallet<S: EthereumSigner + Clone>(wallet: Wallet<S>) -> Result<(
...
+        .await
+        .map_err(GenericError::new)?
+        .total_fee;
+    info!(
+        "Withdrawal transaction fee {:.5}",
+        utils::big_uint_to_big_dec(withdraw_fee.clone())
+    );
+
+    let amount = amount
+        .ok_or(GenericError::new("unused"))
```
pnowosie, line 216: Other options?
This is unused error just to satisfy type-check, I cannot use `Option::and_then` directly because fn used needs to return `Option` not `Result<>` as `utils::big_dec_to_big_uint` do.

So I map `Option` to `Result` - what kind of error is more appropriate here?
It won't be used anyway because of `.unwrap_or(balance.clone())` ...

WDYT :question:

Wiezzel: How about something along these lines?
```rust
let amount = match amount {
    Some(amount) => utils::big_dec_to_big_uint(amount)?,
    None => balance.clone(),
};
```

maaktweluit (approved):
LGTM!

Wiezzel (approved):

PR #912, opened 2021-01-12
Payments: Mainnet/testnet preparation

Wiezzel:
Introduced changes to the payment service and payment driver API necessary for supporting both mainnet and testnet in a single Yagna daemon process. This changes require drivers to be network-aware.

TODO:
- [ ] Update ERC-20 (NGNT) driver
- [ ] Update zkSync driver
- [ ] Update README


core/payment-driver/dummy/src/service.rs
```diff
@@ -43,9 +55,10 @@ async fn init(_db: (), _caller: String, msg: Init) -> Result<Ack, GenericError>
     let mode = msg.mode();
 
     let msg = payment_srv::RegisterAccount {
-        platform: PLATFORM_NAME.to_string(),
         address,
         driver: DRIVER_NAME.to_string(),
+        network: "".to_string(),
```
maaktweluit, line 60: would it be nicer to have an `Option<String>` here instead of an empty one?

core/payment/src/processor.rs
```diff
@@ -123,41 +197,29 @@ impl DriverRegistry {
...
-            if reg_mode.contains(mode) {
-                return Ok(driver.to_owned());
+            if details.mode.contains(mode) {
+                return Ok(details.driver.clone());
             }
         }
 
         // If it's recv-only mode or no-mode (i.e. checking status) we can use any driver that
         // supports the given platform and doesn't require init for receiving.
         if !mode.contains(AccountMode::SEND) {
```
maaktweluit, line 211: bit odd to check `if not send` to handle `if receive`.. will work till we get more modes :)

maaktweluit (commented):
LGTM!

2 nitty comments / questions

checking ya-client side now

core/payment/src/processor.rs
Wiezzel: The mode can be:
* Send
* Receive
* Send+Receive
* None

This conditional is specifically designed to capture Receive and None as stated by the comment above. It is assumed that if receiving funds doesn't require initialization then checking status will nit require it either.

core/payment-driver/dummy/src/service.rs
Wiezzel: That's actually a bug! I forgot to put `NETWORK_NAME` and `TOKEN_NAME` here.

Wiezzel: ...and fixed!

nieznanysprawiciel (approved):
Approve for provider part

core/payment/src/processor.rs
maaktweluit: You are right, thanks for the explanation!

PR #915, opened 2021-01-12
Event api/fix ya provider runner sess

tworec:


zakaprov (approved):

mfranciszkiewicz (dismissed):

mfranciszkiewicz (approved):

agent/provider/src/provider_agent.rs
```diff
@@ -155,6 +155,7 @@ impl ProviderAgent {
         // Generate session id from node name and process id to make sure it's unique.
         let name = args.node.node_name.clone().unwrap_or(app_name.to_string());
         args.market.session_id = format!("{}-[{}]", name, std::process::id());
+        args.runner.session_id = args.market.session_id.clone(); // TODO: unwind this dirty fix
```
mfranciszkiewicz, line 158: Why is this fix dirty? `session_id` was designed to be a part of the configuration.

A similar solution was already merged to `event-api/master` from https://github.com/golemfactory/yagna/pull/850/files#diff-4404c9c07948836afc87344e867a1bdcb8ff5d0f0e487f6313b7a96f34cc58adR141-R143

tworec: 👌 let it be then ! :)
I'm having a tendency to be too strict

PR #926, opened 2021-01-14
ZkSync driver: retry submitting transaction

Wiezzel:
If submitting transaction fails (e.g. due to operator downtime), retry for some time (hard-coded 15 minutes).

Retrying transactions that failed verification is **not included** because it is a much more complicated matter and cannot be easily solved without major rework of the driver. Transactions that failed verification will in general *fail it again* if re-submitted (by definition of stupidity: "Doing the same thing and expecting different results"). Possible scenario when it would make sense to retry a transaction that failed to verify is "not enough funds" but it requires detecting that account was topped up before re-sending the transaction -- and that's a non-trivial task. Therefore I decide to leave it out of the scope since we have more important issues to solve at this moment.

---
###### Regression test
Just run `payment_api --driver=zksync` and `invoice_flow` to check if transactions work.

###### Functionality test
Modify `core/payment-driver/zksync/src/zksync/wallet.rs` and put this code at the first line of `make_transfer`:
```rust
pub async fn make_transfer(details: &PaymentDetails, nonce: u32) -> Result<String, GenericError> {
    return Err(GenericError::new("This is a test".to_string()));
```
I know it's kinda lame but that's the fastest way to simulate operator error. You can also lower the hard-coded timeout to make the test faster. Then run `payment_api --driver=zksync` and `invoice_flow`. The server (`payment_api`) should output warnings like this until deadline is reached:
```
[2021-01-14T21:01:02Z WARN  ya_zksync_driver::driver] Failed to submit zkSync transaction. Payment will be retried until 2021-01-14 21:01:14.138368546 UTC. details=PaymentEntity { ... } error=This is a test
```
Finally, it should display an error:
```
[2021-01-14T21:01:32Z ERROR ya_zksync_driver::driver] Failed to submit zkSync transaction. Retry deadline reached. details=PaymentEntity { ... } error=This is a test
```

### DIFF
--- a/core/payment-driver/zksync/src/dao.rs
+++ b/core/payment-driver/zksync/src/dao.rs
@@ -118,27 +118,19 @@ impl ZksyncDao {
         tx_id
     }
 
-    pub async fn transaction_confirmed(&self, tx_id: &str, result: bool) -> Vec<PaymentEntity> {
-        let status = if result {
-            TransactionStatus::Confirmed
-        } else {
-            TransactionStatus::Failed
-        };
-
+    pub async fn transaction_confirmed(&self, tx_id: &str) -> Vec<PaymentEntity> {
         if let Err(e) = self
             .transaction()
-            .update_tx_status(tx_id.to_string(), status.into())
+            .update_tx_status(tx_id.to_string(), TransactionStatus::Confirmed.into())
             .await
         {
             log::error!("Failed to update tx status for {:?} : {:?}", tx_id, e)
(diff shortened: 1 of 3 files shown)


maaktweluit (dismissed):
LGTM!

Just reviewed the code, will run some tests after finishing the high priority issues

maaktweluit (approved):
LGTM!

Ran all the test and they worked as expected.

Also restarted the payment_api example and it (tried to) send the tx after starting

PR #930, opened 2021-01-15
Market/regenerate proposal after termination fix

tworec:
improovement for #873


core/market/src/negotiation/error.rs
```diff
@@ -159,12 +159,10 @@ pub enum ProposalError {
 }
 
 #[derive(Error, Debug)]
+#[error("Failed regenerate proposal: {0}.")]
```
nieznanysprawiciel, line 162: :O I was looking for something like this

PR #936, opened 2021-01-19
Payments: multi-network support

Wiezzel:
Added `platform` field to `PaymentDetails` and `SchedulePayment`. Supporting multiple networks invalidates the assumption that every driver supports a single payment platform. Therefore, this additional information is needed to be able to make appropriate payment.

Implemented validation of payment platforms in PaymentProcessor to secure against cross-network replay attacks (i.e. making payment on testnet and claiming it to be a mainnet payment).


core/payment/src/cli.rs
```diff
@@ -17,27 +17,39 @@ use crate::{wallet, DEFAULT_PAYMENT_DRIVER};
...
+    //     #[structopt(long)]
+    //     account: Option<String>,
+    //     #[structopt(long, default_value = DEFAULT_PAYMENT_DRIVER)]
+    //     driver: String,
+    //     #[structopt(long)]
+    //     network: Option<String>,
+    //     #[structopt(long)]
+    //     amount: String,
+    // },
+    Exit {
```
tworec, line 31: pls add `///` comments to fill `--help` for all commands

core/payment/src/cli.rs
```diff
@@ -56,17 +68,19 @@ pub enum PaymentCli {
         #[structopt(subcommand)]
```
tworec, line 68: I'm not sure, but seems that it is the only `address` within payment CLI. Shouldn't it be renamed to `account` to unify CLI?

Wiezzel: This is node ID, not payment account address. I'm not sure what to do with it. I don't even know if anyone is using this command.

core/payment/src/cli.rs
Wiezzel: Good idea. Will do if there's enough time.

maaktweluit (approved):
LGTM!

Tested all CLI Commands and invoice_flow scenario flavours

tworec (approved):

PR #939, opened 2021-01-19
Implement exit command for zksync driver

pnowosie:
Implements `exit` command based on already added `wallet::withdraw`.

**Changes:**

- extend `Exit` (message) structure of `sender` and accessor methods,
- retrieve confirmed transaction's hash,
- improve messages returned from command to the caller
- improve an example (tracking url to the zk-sync transaction)

**Tests:**

- :heavy_check_mark: Exiting with just amount 
```
#> cargo run payment exit --amount=10
Withdrawal has been accepted by the zkSync operator. It may take some time until the funds are available on Ethereum blockchain. 
Tracking link: https://rinkeby.zkscan.io/explorer/transactions/c0dedebfedb6825c17b548769792b120498f50722f4c6e48c02f9058f7c6d310
```

- :heavy_check_mark: Exiting to different address
- :heavy_check_mark: Cleaning account without parameters

Check it here: [[zk-scan @ 0x5943...9d15](https://rinkeby.zkscan.io/explorer/accounts/0x59432583686171d091c0c058f187d34b93339d15)]

### DIFF
--- a/core/model/src/driver.rs
+++ b/core/model/src/driver.rs
@@ -298,6 +298,13 @@ impl Exit {
             token,
         }
     }
+
+    pub fn amount(&self) -> Option<BigDecimal> {
+        match &self.amount {
+            Some(a) => Some(a.clone()),
+            None => None,
+        }
+    }
 }
 
 impl RpcMessage for Exit {
--- a/core/model/src/driver.rs
+++ b/core/model/src/driver.rs
@@ -278,6 +278,7 @@ impl RpcMessage for Enter {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Exit {
+    sender: String,
     to: Option<String>,
     amount: Option<BigDecimal>,
     network: Option<String>,
@@ -286,12 +287,14 @@ pub struct Exit {
 
 impl Exit {
     pub fn new(
+        sender: String,
         to: Option<String>,
         amount: Option<BigDecimal>,
         network: Option<String>,
         token: Option<String>,
     ) -> Exit {
         Exit {
+            sender,
             to,
             amount,
             network,
@@ -299,12 +302,9 @@ impl Exit {
         }
     }
 
-    pub fn amount(&self) -> Option<BigDecimal> {
-        match &self.amount {
-            Some(a) => Some(a.clone()),
-            None => None,
-        }
-    }
+    pub fn amount(&self) -> Option<BigDecimal> { self.amount.clone() }
+    pub fn sender(&self) -> String { self.sender.clone() }
+    pub fn to(&self) -> Option<String> { self.to.clone() }
 }
 
 impl RpcMessage for Exit {
--- a/core/model/src/driver.rs
+++ b/core/model/src/driver.rs
@@ -302,9 +302,15 @@ impl Exit {
         }
     }
 
-    pub fn amount(&self) -> Option<BigDecimal> { self.amount.clone() }
-    pub fn sender(&self) -> String { self.sender.clone() }
-    pub fn to(&self) -> Option<String> { self.to.clone() }
+    pub fn amount(&self) -> Option<BigDecimal> {
+        self.amount.clone()
+    }
+    pub fn sender(&self) -> String {
+        self.sender.clone()
+    }
+    pub fn to(&self) -> Option<String> {
+        self.to.clone()
+    }
 }
 
 impl RpcMessage for Exit {
--- a/core/payment-driver/zksync/src/driver.rs
+++ b/core/payment-driver/zksync/src/driver.rs
@@ -128,8 +128,9 @@ impl PaymentDriver for ZksyncDriver {
         _caller: String,
         msg: Exit,
     ) -> Result<String, GenericError> {
-        log::info!("EXIT = Not Implemented: {:?}", msg);
-        Ok("NOT_IMPLEMENTED".to_string())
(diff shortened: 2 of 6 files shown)


core/model/src/driver.rs
```diff
@@ -298,6 +298,13 @@ impl Exit {
             token,
         }
     }
+
+    pub fn amount(&self) -> Option<BigDecimal> {
+        match &self.amount {
+            Some(a) => Some(a.clone()),
```
pnowosie, line 304: `BigDecimal` does not clone as primitive types which makes me trouble... is it OK, alternatives :question: 
:bow:

core/payment-driver/zksync/src/driver.rs
```diff
@@ -128,8 +128,9 @@ impl PaymentDriver for ZksyncDriver {
         _caller: String,
         msg: Exit,
     ) -> Result<String, GenericError> {
-        log::info!("EXIT = Not Implemented: {:?}", msg);
-        Ok("NOT_IMPLEMENTED".to_string())
+        wallet::exit(_caller, &msg)
+            .await
+            .and(Ok("Success!".to_string()))
```
pnowosie, line 133: Or should I return link to [[zkscan](https://rinkeby.zkscan.io/)] ...

core/model/src/driver.rs
Wiezzel: This is OK, but I think you can simplify the code by simply calling `self.amount.clone()` and get rid of the `match`.

core/payment-driver/zksync/src/driver.rs
Wiezzel: I believe `wallet::exit` should return transaction hash and here it should be formatted into some kind of nice message like "Withdrawal has been accepted by the zkSync operator. It may take some time until the funds are available on Ethereum blockchain. Tracking link: {}"

core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -71,6 +71,20 @@ pub async fn init_wallet(msg: &Init) -> Result<(), GenericError> {
...
+pub async fn exit(address: String, msg: &Exit) -> Result<(), GenericError> {
+    let wallet = get_wallet(&address).await?;
+    let tx_handle = withdraw(wallet, msg.amount()).await?;
+    let tx_info = tx_handle
+        .wait_for_commit()
+        .await
+        .map_err(GenericError::new)?;
+    match tx_info.success {
+        Some(true) => Ok(()),
+        Some(false) => Err(GenericError::new(tx_info.fail_reason.unwrap())),
```
maaktweluit, line 83: Would use `unwrap_or("Unknown failure reason")`

core/payment-driver/zksync/src/driver.rs
```diff
@@ -128,8 +128,9 @@ impl PaymentDriver for ZksyncDriver {
         _caller: String,
         msg: Exit,
     ) -> Result<String, GenericError> {
-        log::info!("EXIT = Not Implemented: {:?}", msg);
-        Ok("NOT_IMPLEMENTED".to_string())
+        wallet::exit(_caller, &msg)
```
maaktweluit, line 131: the `_` is in front of the param because its is in the "interface" but its not used yet.

When you start using it you can remove the `_`

core/payment-driver/zksync/src/driver.rs
maaktweluit, line 133: since `exit()` returns a result you can unwrap it with `?`

I would also make it one line one thought:

```
let tx_hash = wallet::exit(_caller, &msg).await?;
Ok(tx_hash)
```

This requires exit to return the `tx_hash` or `tx_info`

maaktweluit (commented):
Looks good!

Some small comments :)

core/payment-driver/zksync/src/zksync/wallet.rs
Wiezzel: That's how I did it in another PR: https://github.com/golemfactory/yagna/pull/926/files#diff-932bad1dbb320024ab301b2c407b0bbddfd848c87c78f7dd2c1b44b46e7a5425

core/payment-driver/zksync/src/driver.rs
Wiezzel: Oh, I think this is serious bug that we have overlooked. `_caller` **should be ignored**. This **not** the address to withdraw from. The address should be in the message (but it's not)!

Wiezzel (changes requested):

core/payment-driver/zksync/src/driver.rs
maaktweluit: you are right, i was already wondering what `caller` was doing..

`msg.sender` should be used ( and checked if its in the `self. is_account_active(&msg.sender)` )

core/model/src/driver.rs
```diff
@@ -278,6 +278,7 @@ impl RpcMessage for Enter {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Exit {
+    sender: String,
```
maaktweluit, line 281: i think you should add sender to the CLI command too now.

You can use `resolve_address()` in `core/payment/src/cli.rs` to remove optional and get your `sender` 

Would call this argument `address` to match the `init` command

Wiezzel (approved):

PR #940, opened 2021-01-19
enclose version functionality in ya-compile-time-utils

tworec:
It encloses version reporting in ya-compile-time-utils
made on top of #933

### DIFF
--- a/utils/compile-time-utils/src/lib.rs
+++ b/utils/compile-time-utils/src/lib.rs
@@ -1,7 +1,8 @@
 use git_version::git_version;
+use metrics::value;
 use semver::Version;
 
-/// Returns latest tag or version from Cargo.toml`.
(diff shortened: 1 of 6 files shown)


utils/compile-time-utils/src/lib.rs
```diff
@@ -20,39 +21,51 @@ pub fn build_date() -> &'static str {
...
+pub fn semver_str() -> String {
     let mut version = git_tag();
     for prefix in ["pre-rel-", "v"].iter() {
         if version.starts_with(prefix) {
             version = &version[prefix.len()..];
         }
     }
-    version
+    if let Some(bn) = build_number() {
+        [version, bn].join("+")
```
etam, line 37: I don't think it's a good idea to smuggle in a change in version format. Build number is already included in `version_describe`.

PR #956, opened 2021-01-21
[golemsp] lower default price per hour of computations from 1 to 0.1 GLM

tworec:
As discussed with all reviewers


Wiezzel (approved):

83tb:
good to go!

83tb (approved):
good to go!

PR #957, opened 2021-01-22
Zksync multi network

maaktweluit:


core/model/src/driver.rs
```diff
@@ -162,6 +162,14 @@ impl Init {
...
         self.mode.clone()
     }
+
+    pub fn network(&self) -> Option<String> {
+        self.network.clone()
+    }
+
+    pub fn token(&self) -> Option<String> {
+        self.token.clone()
+    }
```
Wiezzel, line 172: These are already added above. Please remove duplicates.

core/payment-driver/base/migrations/2021-01-19-123456_multi_platform/down.sql
```diff
@@ -0,0 +1,28 @@
+-- HACK: All this code below is just to drop column timestamp from table pay_allocation
```
Wiezzel, line 1: Copy-paste comment not accurate. :P

core/payment-driver/zksync/src/driver.rs
```diff
@@ -209,9 +210,9 @@ impl PaymentDriver for ZksyncDriver {
             .map_err(GenericError::new)??;
 
         let mode = msg.mode();
-        let network = DEFAULT_NETWORK; // TODO: Implement multi-network support
-        let token = DEFAULT_TOKEN; // TODO: Implement multi-network support
-        bus::register_account(self, &address, network, token, mode).await?;
+        let network = msg.network().unwrap_or(DEFAULT_NETWORK.to_string());
+        let token = msg.token().unwrap_or(DEFAULT_TOKEN.to_string());
```
Wiezzel, line 214: It should be the default token **for this network** not universally-default token. If `msg.network` is `"mainnet"` and `msg.token` is `None` then `token` should be `"GLM"` not `"tGLM"`.

Please check if you didn't make the same error somewhere else.

core/payment-driver/zksync/src/zksync/faucet.rs
```diff
@@ -25,8 +25,8 @@ lazy_static! {
     static ref MAX_WAIT: Duration = Duration::minutes(1);
 }
 
-pub async fn request_ngnt(address: &str) -> Result<(), GenericError> {
-    let balance = account_balance(address).await?;
+pub async fn request_ngnt(address: &str, network: Network) -> Result<(), GenericError> {
```
Wiezzel, line 28: We won't request funds on mainnet, so I think this parameter is not needed. (It will be only if we introduce more testnets, I believe).

core/model/src/driver.rs
maaktweluit: whoops, rebase mistake.. fixing :)

core/payment-driver/base/migrations/2021-01-19-123456_multi_platform/down.sql
maaktweluit: good catch, fixing :)

core/payment-driver/zksync/src/driver.rs
maaktweluit: was replacing all `DEFAULT_TOKEN` with `get_network_token()` yesterday, will fix this and check for the other locations

core/payment-driver/zksync/src/zksync/faucet.rs
maaktweluit: indeed, but the get_provider and get_balance functions need this argument.

I think its also neater to already include network here for when test-nets get added

Wiezzel: Alright.

core/payment-driver/zksync/src/dao.rs
```diff
@@ -56,6 +62,8 @@ impl ZksyncDao {
         let recipient = msg.recipient().to_owned();
         let gnt_amount = utils::big_dec_to_u256(msg.amount());
         let gas_amount = Default::default();
+        // TODO: Remove unwrap? default to rinkeby?
+        let (network, _token) = platform_to_network_token(msg.platform()).unwrap();
```
Wiezzel, line 66: Error should be propagated to `schedule_payment`. Invalid platform would be a sign that there's a bug in payment service,

core/payment-driver/zksync/src/dao.rs
```diff
@@ -129,6 +138,17 @@ impl ZksyncDao {
...
 
+    pub async fn get_first_payment(&self, tx_hash: &str) -> Option<PaymentEntity> {
+        match self
+            .payment()
+            .get_first_by_tx_hash(tx_hash.to_string())
+            .await
+        {
+            Ok(payment) => Some(payment),
+            Err(_) => None,
+        }
```
Wiezzel, line 149: Maybe just use `optional()` in `get_first_by_tx_hash` and remove this match?

core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -4,75 +4,83 @@
...
+    let mut balance_com = acc_info
         .committed
         .balances
-        .get(ZKSYNC_TOKEN_NAME)
+        .get(&token)
         .map(|x| x.0.clone())
         .unwrap_or(BigUint::zero());
+    // Hack to get GNT balance for backwards compatability
+    // TODO: Remove this if {} and the `mut` from `let mut balance_com`
+    if network == Network::Rinkeby && balance_com == BigUint::zero() {
```
Wiezzel, line 45: Please contact Matter Labs team regarding this issue.

core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -135,7 +151,7 @@ pub async fn make_transfer(details: &PaymentDetails, nonce: u32) -> Result<Strin
         .nonce(nonce)
         .str_to(&details.recipient[2..])
         .map_err(GenericError::new)?
-        .token(ZKSYNC_TOKEN_NAME)
+        .token(token.as_ref()) // TODO: use `token`
```
Wiezzel, line 154: Is this `TODO` still relevant? Looks like already using `token`.

core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -170,16 +186,23 @@ pub async fn check_tx(tx_hash: &str) -> Option<bool> {
 //     }
 // }
 
-fn get_provider() -> Provider {
-    (*PROVIDER).clone()
+fn get_provider(network: Network) -> Provider {
+    let provider: Provider = match env::var("ZKSYNC_RPC_ADDRESS").ok() {
```
Wiezzel, line 190: I think we should no longer have a single variable for zkSync RPC address but rather separate variables for every network.

core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -188,24 +211,43 @@ async fn get_wallet(address: &str) -> Result<Wallet<YagnaEthSigner>, GenericErro
     Ok(wallet)
 }
 
-async fn unlock_wallet<S: EthereumSigner + Clone>(wallet: Wallet<S>) -> Result<(), GenericError> {
+fn get_zk_network(network: Network) -> ZkNetwork {
```
Wiezzel, line 214: Not sure if this deserves to be a separate function.

core/payment-driver/zksync/src/dao.rs
maaktweluit: good one, updated return value and removed unwrap

core/payment-driver/zksync/src/dao.rs
maaktweluit: then the error is not matched, and the cron can not really handle errors :(

core/payment-driver/zksync/src/zksync/wallet.rs
maaktweluit: this one is the "known" issue, balances returns "GNT" still, will reach out to them about transfer failing to work with tGLM

core/payment-driver/zksync/src/zksync/wallet.rs
maaktweluit: nope, removed :)

core/payment-driver/zksync/src/zksync/wallet.rs
maaktweluit: good idea, but i propose to create an issue for this and move on with finishing `fund`

core/payment-driver/zksync/src/zksync/wallet.rs
maaktweluit: its only nice for the "central" default value, now it makes little sense ( only called 2x )

core/payment-driver/zksync/src/dao.rs
Wiezzel: Alright, let it be that way.

core/payment-driver/zksync/src/zksync/wallet.rs
Wiezzel: OK

Wiezzel (approved):

PR #959, opened 2021-01-22
Don't clean old version entries

jiivan:


core/market/tests/test_agreement.rs
```diff
@@ -1049,9 +1049,7 @@ async fn test_terminate_from_wrong_states() {
                 from: AgreementState::Proposal,
                 to: AgreementState::Terminated,
             },
-        )) => {
-            assert_eq!(id, agreement_id)
-        }
+        )) => assert_eq!(id, agreement_id),
```
jiivan, line 1052: `cargo fmt` changed this file

PR #961, opened 2021-01-22
Version/service

tworec:


core/version/src/lib.rs
```diff
@@ -63,25 +74,38 @@ pub mod notifier {
...
-                log::warn!("New version of yagna available {}! Close yagna and run `{}` to install or `{}` to mute this notification", db_release, UPDATE_CURL, SILENCE_CMD);
-            };
+            match release_dao.pending_release().await {
+                Ok(Some(release)) => {
+                    if !release.seen {
+                        log::warn!(
+                            "New Yagna version {} ({}) is available. TODO: add howto here",
+                            release.name,
+                            release.version
+                        )
```
jiivan, line 103: Why not use `Display` of `Release` as before?

core/version/src/lib.rs
```diff
@@ -1,19 +1,20 @@
...
 
 pub mod notifier {
+    use metrics::counter;
     use std::time::Duration;
-    use tokio::time;
 
     use ya_persistence::executor::DbExecutor;
 
-    const UPDATE_CURL: &'static str = "curl -sSf https://join.golem.network/as-provider | bash -";
-    const SILENCE_CMD: &'static str = "yagna update skip";
```
jiivan, line 16: Are those wrong?

core/version/src/lib.rs
tworec: I'm using `thiserror` to implement ToString and Display for the release.

core/version/src/lib.rs
tworec: sorry. Removed it by misunderstanding. Reverted and fixed skip command as it is `yagna version skip`

core/version/src/db/dao.rs
```diff
@@ -0,0 +1,117 @@
...
+        .filter(release::version.eq(&ver))
+        .first::<DBRelease>(conn)
+        .optional()?
+        .map(|db_rel| db_rel.into()))
+}
+
+fn get_pending_release(conn: &ConnType) -> anyhow::Result<Option<Release>> {
+    let query = version_release
+        .filter(release::seen.eq(false))
+        .order((release::release_ts.desc(), release::version.desc()))
```
jiivan, line 94: I'm afraid that ordering by version won't work here. (`0.9.0' > '0.10.0`)

core/version/src/db/model.rs
```diff
@@ -0,0 +1,64 @@
...
+impl DBRelease {
+    pub(crate) fn current() -> anyhow::Result<Self> {
+        Ok(DBRelease {
+            version: ya_compile_time_utils::semver_str().into(),
+            name: ya_compile_time_utils::build_number_str()
+                .map(|n| format!("{} build #{}", ya_compile_time_utils::git_rev(), n))
+                .unwrap_or(ya_compile_time_utils::git_rev().into()),
+            seen: false,
+            release_ts: NaiveDateTime::parse_from_str(
+                ya_compile_time_utils::build_date(),
```
jiivan, line 29: I'm not sure about this. It's definitely not the same as github release date.

core/version/src/github.rs
```diff
@@ -0,0 +1,94 @@
...
+                rel.version,
+                e
+            )
+        })?
+    {
+        match db.as_dao::<ReleaseDAO>().save(db_rel).await {
+            Err(e) => log::error!("Storing new Yagna release {} to DB. {}", rel, e),
+            Ok(r) => {
+                counter!("version.new", 1);
+                log::warn!("{}", ReleaseMessage::Available(&r));
```
jiivan, line 42: I don't like the idea of claiming `pinger` responsibility here. It can result in ugly spamming of log more frequently than defined by interval in `pinger`.

core/version/src/db/dao.rs
tworec: good catch

core/version/src/db/model.rs
tworec: yes, I was also not sure. I will move it to the name string or introduce a new dedicated column

core/version/src/github.rs
tworec: this will be invoked:
- every 24h by `worker`
- on every start
- manually via `yagna version check` 
It will **immediately** inform the user (via logs) that there is new version. This immediate notification was my goal here.

The same message will be logged by `pinger` every 30mins.

core/version/src/db/model.rs
tworec: but it was added here to enable us to sort via this `release_ts` column (your comment above). And if we hardcode it for fake-entries it will not be possible, that's why I have added second column to sort (comment above).

tworec: I'm setting `release_ts` to now to be able to sort it properly (comment above)

core/version/src/db/dao.rs
tworec: ok, now's that `release_ts` is unique I can remove the second column from `order by`

core/version/src/github.rs
jiivan: So every 24h we will get double logs from pinger and worker?

jiivan (approved):

PR #967, opened 2021-01-26
Doc-strings to driver's functions

pnowosie:


core/payment-driver/base/src/bus.rs
```diff
@@ -111,6 +111,7 @@ pub async fn list_unlocked_identities() -> Result<Vec<NodeId>, GenericError> {
     Ok(unlocked_list)
 }
 
+/// Notifies the Payment service that the driver is ready.
```
Wiezzel, line 114: Nope. It notifies the payment service that an **account** is ready (for sending/receiving funds - see mode).

core/payment-driver/base/src/driver.rs
```diff
@@ -20,66 +20,88 @@ pub use ya_core_model::payment::local::Network;
...
     async fn exit(&self, db: DbExecutor, caller: String, msg: Exit)
         -> Result<String, GenericError>;
 
     // used by bus to bind service
     fn get_name(&self) -> String;
     fn get_default_network(&self) -> String;
     fn get_networks(&self) -> HashMap<String, Network>;
+
+    /// Tells whether funds are available for spent just after depositing
+    /// on the driver's supported network or the additional unlocking is needed.
```
Wiezzel, line 60: Nope. This flag specifies whether account initialization is needed for **receiving** payments.

core/payment-driver/base/src/driver.rs
```diff
@@ -20,66 +20,88 @@ pub use ya_core_model::payment::local::Network;
...
     fn recv_init_required(&self) -> bool;
 
+    /// Gets the balance of the funds sent from the sender to the recipient.
```
Wiezzel, line 63: Please add an annotation that this method is deprecated and drivers can just provide a stub returning some *very big* number (i.e. whole token supply).

core/payment-driver/base/src/driver.rs
```diff
@@ -20,66 +20,88 @@ pub use ya_core_model::payment::local::Network;
...
     async fn get_transaction_balance(
         &self,
         db: DbExecutor,
         caller: String,
         msg: GetTransactionBalance,
     ) -> Result<BigDecimal, GenericError>;
 
+    /// Initializes the driver service. It should call
```
Wiezzel, line 71: Initializes an **account** not the whole driver. Driver can handle multiple accounts.

Wiezzel (changes requested):

Wiezzel:
Please update to master and include `fund` method as well.

core/payment-driver/base/src/driver.rs
```diff
@@ -20,68 +20,93 @@ pub use ya_core_model::payment::local::Network;
...
 
     // used by bus to bind service
     fn get_name(&self) -> String;
     fn get_default_network(&self) -> String;
     fn get_networks(&self) -> HashMap<String, Network>;
+
+    /// Tells whether account initialization is needed for receiving payments.
     fn recv_init_required(&self) -> bool;
 
+    #[deprecated(
```
maaktweluit, line 62: nice idea! but you also need to patch the build so it allows for these deprecated warnings

core/payment-driver/base/src/driver.rs
```diff
@@ -20,68 +20,93 @@ pub use ya_core_model::payment::local::Network;
...
+    /// Transfers the funds between specified accounts. Called by CLI.
     async fn transfer(
         &self,
         db: DbExecutor,
         caller: String,
         msg: Transfer,
     ) -> Result<String, GenericError>;
 
+    /// Schedules the payment between specified accounts. Payments
+    /// are processed by the `cron` job. Payment tracking is done by
```
maaktweluit, line 90: usint the cron ( and DB ) is actually optional for new drivers, take the dummy driver as an example..

Not sure if this needs updating according to this, just FYI

core/payment-driver/base/src/driver.rs
```diff
@@ -20,68 +20,93 @@ pub use ya_core_model::payment::local::Network;
...
+    /// cron job, see: `PaymentDriverCron::confirm_payments`.
     async fn schedule_payment(
         &self,
         db: DbExecutor,
         caller: String,
         msg: SchedulePayment,
     ) -> Result<String, GenericError>;
 
+    /// Verifies the payment transaction specified by transaction's hash.
```
maaktweluit, line 99: tx_hash is an implementation detail

the verify_payment is called on the provider, with data generated on the requestor ( in zksync and erc20 cases this is tx_hash )

core/payment-driver/base/src/driver.rs
```diff
@@ -20,68 +20,93 @@ pub use ya_core_model::payment::local::Network;
...
     async fn verify_payment(
         &self,
         db: DbExecutor,
         caller: String,
         msg: VerifyPayment,
     ) -> Result<PaymentDetails, GenericError>;
 
+    /// Validates that allocated funds are still sufficient to cover
+    /// the costs of the task (including the cost of the Gas).
```
maaktweluit, line 108: `Gas` is network specific, i think "transaction fees" would be a better, more generic name

core/payment-driver/base/src/bus.rs
```diff
@@ -145,6 +147,8 @@ pub async fn sign(node_id: NodeId, payload: Vec<u8>) -> Result<Vec<u8>, GenericE
     Ok(signature)
 }
 
+/// Notifies the Payment service that the scheduled payment is processed successfully.
```
maaktweluit, line 150: maybe worth adding "calling this will call `verify_payment` in the provider driver

maaktweluit (commented):
Nice work!

Mostly nitty comments, will approve on reply / fix

core/payment-driver/base/src/driver.rs
pnowosie: Agree on 'tx_hash' as a detail. 
What I've dug in zk-sync seemed to work the way:

- schedule payment returns order-id (random guid),
- cron's process payment (exactly handle payment) links order-id with transaction's identifier (tx hash)
- verify payment accepts this identifier (not an order id)

So the cron function actually links artificial order-id with tx identifier in the DB

Wiezzel: Verify payment is called on a different node that doesn't have this 'link' at all. Driver should be able to generate `PaymentDetails` based `confirmation` and blockchain data alone.

core/payment-driver/base/src/driver.rs
pnowosie: I will made it clear in the guide. Not messing here for simplicity.

core/payment-driver/base/src/driver.rs
pnowosie: Ah in this case I think I will do with just a comment. Making the warning and then silencing it seems like solving unreal problems ;)

core/payment/examples/README.md
```diff
@@ -10,6 +10,8 @@ cargo run --example payment_api
 ```
 To use GNT instead of dummy driver us `cargo run --example payment_api -- --driver=gnt` instead.
```
Wiezzel, line 11: Could you update this line as well? It should be like this:
ERC-20 driver on rinkeby: `cargo run --example payment_api -- --driver=erc20 --platform=erc20-rinkeby-tglm`
ERC-20 driver on mainnet: `cargo run --example payment_api -- --driver=erc20 --network=mainnet --platform=erc20-mainnet-glm`
ZkSync driver on rinkeby: `cargo run --example payment_api -- --driver=zksync --platform=zksync-rinkeby-tglm`
ZkSync driver on mainnet: `cargo run --example payment_api -- --driver=zksync --network=mainnet --platform=zksync-mainnet-glm`

The `--platform` should be passed to `invoice_flow`/`debit_not_flow` as well.

Wiezzel (commented):
One more thing, and it's good to go.

Wiezzel (approved):

PR #974, opened 2021-01-27
ERC-20 payment driver: added network field to transaction

Wiezzel:
For the sender to calculate nonces properly, it is necessary to include network in transaction model as well.

Additionally:
* Added some log messages.
* Remove `unwrap()` that cause driver to panic on tx submit error.
* Extended timeout in invoice_flow and debit_note_flow (300s wasn't always enough for mainnet transactions).


maaktweluit (approved):
LGTM!

PR #975, opened 2021-01-27
Enforce developers to update lock file, otherwise test will fail

maaktweluit:
First commit will test if it actually breaks, will update lock file in the second.

Also moved `tree` and `fmt` to run before the long cache task, fail fast!


zakaprov (approved):

Wiezzel (approved):

tworec (commented):
👍

PR #978, opened 2021-01-27
ZkSync driver: cron job mutexes

Wiezzel:
If a new payment scheduling job started while a previous one had been already in progress, it would lead to database errors due to single transaction being processed twice. Introduced mutexes for payment scheduling job and payment confirmation job to avoid it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7563,6 +7563,7 @@ dependencies = [
  "diesel",
  "diesel_migrations",
  "ethereum-types 0.6.0",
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
LGTM!

One thought: maybe we should have one lock for both jobs to ensure they are not running at the same time?

Still approving!

Wiezzel:
I believe these two jobs should not interfere. They don't in zkSync driver at least.

PR #981, opened 2021-01-28
Key backup command for Identity

pnowosie:
```
  pnowosie/pay-146-key-backup-command ?  cargo run id export --help          ✔  9710  14:14:32
yagna-id-export 0.2.0
Exports given identity to a file | stdout

USAGE:
    yagna id export [FLAGS] [OPTIONS] [node-or-alias]

FLAGS:
        --json       Return results in JSON format
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --file-path <file-path>    File path where identity will be written. Defaults to `stdout`
    -d, --datadir <data-dir>       Service data dir [env: YAGNA_DATADIR]  [default:
                                   /home/pnowosie/.local/share/yagna]
    -g, --gsb-url <gsb-url>        Service Bus (aka GSB) URL [env: GSB_URL]  [default: tcp://127.0.0.1:7464]

ARGS:
    <node-or-alias>    Identity alias to export

```

### Tests

Here are identities
```
┌───────────┬──────────┬─────────┬──────────────────────────────────────────────┐
│  default  │  locked  │  alias  │  address                                     │
├───────────┼──────────┼─────────┼──────────────────────────────────────────────┤
│  X        │          │         │  0x94e1a08cee33f97374bd5dfff8ad972be70d7d31  │
│           │  X       │  bob    │  0xfdea04bd56a8813a623d31c8d7ec100677879e85  │
│           │  X       │  alice  │  0xe3d15125f931495493894e02411be28cc7c43a92  │
└───────────┴──────────┴─────────┴──────────────────────────────────────────────┘
```

- :x: writing to non-existing directory
- :heavy_check_mark: export default to stdout
- :heavy_check_mark: export default to file
- :heavy_check_mark: export bob to file
- :heavy_check_mark: export default to file
- :heavy_check_mark: export alice to stdout
- :x: export alice to already existing file
- :x: try to export non-existing alias (celine)
- :heavy_check_mark: export / restart service / import / unlock

- Diff-ed stdout and file export
```
 cargo run id export bob | diff test/bob.key -
20c20
< }
\ No newline at end of file
---
> }
```

- Retrived bob's data from `yagna.db/identity` table `key_file_json` field and compared with file. Equal but a new line char at EOF.

### DIFF
--- a/core/identity/src/cli/identity.rs
+++ b/core/identity/src/cli/identity.rs
@@ -119,7 +119,6 @@ pub enum IdentityCommand {
     /// Update given identity
     Update {
         /// Identity to update
-        #[structopt(default_value = "")]
         alias_or_id: NodeOrAlias,
         #[structopt(long)]
         alias: Option<String>,
@@ -131,6 +130,16 @@ pub enum IdentityCommand {
     Drop {
         /// Identity alias to drop
         node_or_alias: NodeOrAlias,
+    },
+
+    /// Backup given identity to a file
+    Backup {
+        /// Identity alias to backup
+        node_or_alias: Option<NodeOrAlias>,
+
+        /// File path where identity will be written. Defaults to `stdout`
+        #[structopt(long = "file-path")]
+        file_path: Option<PathBuf>,
     },
 }
 
@@ -261,6 +270,34 @@ impl IdentityCommand {
                     bus::service(identity::BUS_ID)
                         .send(identity::DropId::with_id(id.node_id))
                         .await
+                        .map_err(|e| anyhow::Error::msg(e))?,
+                )
+            }
+            IdentityCommand::Backup {
+                node_or_alias,
+                file_path,
+            } => {
+                let node_id = node_or_alias.clone().unwrap_or_default().resolve().await?;
+                let file_path = if let Some(file) = file_path {
+                    let parent = file.parent().unwrap();
+                    if !parent.is_dir() {
+                        anyhow::bail!(
+                            "Directory '{:?}' does not exists or is write protected",
+                            parent
+                        )
+                    }
+                    if file.is_file() {
+                        anyhow::bail!("File already exists")
+                    }
+                    Some(file.to_str().unwrap().to_owned())
+                } else {
+                    None
+                };
+
+                CommandOutput::object(
+                    bus::service(identity::BUS_ID)
+                        .send(identity::BackupId::new(node_id, file_path))
+                        .await?
                         .map_err(|e| anyhow::Error::msg(e))?,
                 )
             }
--- a/core/identity/src/cli/identity.rs
+++ b/core/identity/src/cli/identity.rs
@@ -278,7 +278,12 @@ impl IdentityCommand {
                 file_path,
             } => {
                 let node_id = node_or_alias.clone().unwrap_or_default().resolve().await?;
-                let file_path = if let Some(file) = file_path {
+                let key_file = bus::service(identity::BUS_ID)
+                    .send(identity::GetKeyFile::new(node_id))
+                    .await?
+                    .map_err(|e| anyhow::Error::msg(e))?;
+
+                if let Some(file) = file_path {
                     let parent = file.parent().unwrap();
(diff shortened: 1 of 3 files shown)


Wiezzel (commented):
Please change PR base to `release/v0.6` cause I'm seeing some unrelated changes.

core/model/src/identity.rs
```diff
@@ -211,6 +211,24 @@ impl RpcMessage for Subscribe {
     type Error = Error;
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetKeyFile {
```
pnowosie, line 216: Can I just alias this struct (~ `type GetKeyFile = NodeId` ) just to be able to bind... :question: 
Seems like unnecessary struct declaration.

Wiezzel: Nope, you wouldn't be able to add any `impl` for a foreign type. The most minimal solution would be:
```rust
pub struct GetKeyFile(pub NodeId);
```
Please note that implementing `new` is not necessary if fields are `pub`.

core/identity/src/cli/identity.rs
```diff
@@ -280,26 +280,26 @@ impl IdentityCommand {
...
                 let node_id = node_or_alias.clone().unwrap_or_default().resolve().await?;
                 let key_file = bus::service(identity::BUS_ID)
-                    .send(identity::GetKeyFile::new(node_id))
+                    .send(identity::GetKeyFile(node_id))
                     .await?
                     .map_err(|e| anyhow::Error::msg(e))?;
 
-                if let Some(file) = file_path {
-                    let parent = file.parent().unwrap();
-                    if !parent.is_dir() {
```
pnowosie: It is a pitty that rust doesn't allow to:
```rust
if let Some(parent) = file.parent() && parent.is_dir { ...
```

core/identity/src/cli/identity.rs
```diff
@@ -264,6 +274,34 @@ impl IdentityCommand {
...
+                let key_file = bus::service(identity::BUS_ID)
+                    .send(identity::GetKeyFile(node_id))
+                    .await?
+                    .map_err(|e| anyhow::Error::msg(e))?;
+
+                match file_path {
+                    Some(file) => {
+                        if file.parent().map_or(false, |parent| !parent.is_dir()) {
+                            anyhow::bail!(
+                                "Directory '{:?}' does not exists or is write protected",
```
Wiezzel, line 291: `Some()` in the error message is confusing and ugly. That's because `parent()` returns `Option<PathBuf>`. I believe there should be a `match` here.
```
Error: Directory 'Some(".....")' does not exists or is write protected
```

core/identity/src/cli/identity.rs
```diff
@@ -264,6 +274,34 @@ impl IdentityCommand {
...
+                                file.parent()
+                            )
+                        }
+                        if file.is_file() {
+                            anyhow::bail!("File already exists")
+                        }
+
+                        std::fs::write(file, key_file).map_err(|e| anyhow::Error::msg(e))?;
```
Wiezzel, line 299: Any error can be automatically converted into `anyhow::Error` so there's not need for explicit `map_err()`.
```suggestion
                        std::fs::write(file, key_file)?;
```

Wiezzel (changes requested):
If I just pass the filename it doesn't work correctly. :(
```
$ cargo run id export --file-path=key.json
Error: Directory 'Some("")' does not exists or is write protected
```
It should work the same as if I passed `./key.json`. Maybe [canonicalizing the path](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) could help?

core/identity/src/cli/identity.rs
pnowosie: I dropped directory checking entirely so we fallback to whatever `fs::write()` will return.

Wiezzel (approved):

maaktweluit (approved):
LGTM!

PR #982, opened 2021-01-28
Payment service: fixed computing payment amounts

Wiezzel:
1. It was previously falsely assumed that total_amount_accepted for an activity or agreement is equal to the total amount of scheduled payments. That is not true in case of accepting non-binding debit notes (i.e. without due date) which increases total_amount_accepted but doesn't trigger payment. This was fixed by introducing total_amount_scheduled field to both activity and agreement model. This field is updated when a payment is scheduled and it is used to compute the payment amount.
2. Allocation's remaining amount was being updated when a payment was confirmed not when it was scheduled, thus allowing double-spending from an allocation if a new payment was scheduled before the previous one got confirmed. Now spent & remaining amounts are updated when scheduling payment to prevent double-spending.
3. Paying zero-amount invoices was causing unnecessary transaction costs. A quick fix has been applied not to schedule payments for zero amount. Although, a proper solution that disallows issuing such invoices in the first place is required.

Fixes https://github.com/golemfactory/yagna-triage/issues/43

### DIFF
--- a/core/model/src/payment.rs
+++ b/core/model/src/payment.rs
@@ -59,8 +59,15 @@ pub mod local {
     }
 
     impl SchedulePayment {
-        pub fn from_invoice(invoice: Invoice, allocation_id: String, amount: BigDecimal) -> Self {
-            Self {
+        pub fn from_invoice(
+            invoice: Invoice,
+            allocation_id: String,
+            amount: BigDecimal,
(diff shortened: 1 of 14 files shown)


maaktweluit (approved):
LGTM!

Tested it with JS:blender example and it works smoothly!

PR #989, opened 2021-01-29
Market - fix requestor gets approval before provider

nieznanysprawiciel:
This won't fix the problem totally, but it should minimize it.


tworec (approved):
Goooodie

nieznanysprawiciel:
We've chosen different approach: https://github.com/golemfactory/yagna/pull/994

PR #1000, opened 2021-02-02
Get faucet url from envvar or resolve srv record

pnowosie:
SRV record is set (thanks to @Wiezzel who figured it out).
```
 nslookup -type=SRV _zk-faucet._tcp.dev.golem.network
Server:		127.0.0.53
Address:	127.0.0.53#53

Non-authoritative answer:
_zk-faucet._tcp.dev.golem.network	service = 10 10 5778 yacn.dev.golem.network.

Authoritative answers can be found from:

```

Testing shows that SRV record is resolved only to hostname & port
`2021-02-03T07:29:38Z INFO  faucet_srv] Resolved faucet host: "yacn.dev.golem.network:5778"`

### Assumptions

For now I'm taking the following assumptions (tech debt)
* all other zk-sync faucets we provide via SRV would end with `zk/donatex`
* environment variable `ZKSYNC_FAUCET_ADDR` value is taken as is without any changes

### Tests
Run withdrawal example with log level `debug`. Example of [[successful withdrawal](https://rinkeby.zkscan.io/explorer/transactions/42e17ba7092e914c65b725e69bd01208b589fdcadb920ecd0f4d20ce8b07f81d)].

```
[2021-02-03T08:11:17Z DEBUG ya_utils_networking::srv_resolver] Resolved address: yacn.dev.golem.network:5778
[2021-02-03T08:11:17Z DEBUG ya_zksync_driver::zksync::faucet] Faucet url: http://yacn.dev.golem.network:5778/zk/donatex/0x19a2...6df3
```

### DIFF
--- a/core/payment-driver/zksync/src/zksync/faucet.rs
+++ b/core/payment-driver/zksync/src/zksync/faucet.rs
@@ -82,6 +82,8 @@ async fn wait_for_ngnt(address: &str, network: Network) -> Result<(), GenericErr
 async fn faucet_donate(address: &str, _network: Network) -> Result<(), GenericError> {
     let client = awc::Client::new();
     let faucet_url = resolve_faucet_url().await?;
+    debug!("Faucet url: {}/{}", faucet_url, address);
+
     let response = client
         .get(format!("{}/{}", faucet_url, address))
         .send()
@@ -99,8 +101,12 @@ async fn faucet_donate(address: &str, _network: Network) -> Result<(), GenericEr
 async fn resolve_faucet_url() -> Result<String, GenericError> {
     match env::var(FAUCET_ADDR_ENVAR) {
         Ok(addr) => Ok(addr),
-        _ => srv_resolver::resolve_record(DEFAULT_FAUCET_SRV_PREFIX)
-            .await
-            .map_err(|_| GenericError::new("Faucet SRV record cannot be resolved")),
+        _ => {
+            let faucet_host = srv_resolver::resolve_yagna_record(DEFAULT_FAUCET_SRV_PREFIX)
+                .await
+                .map_err(|_| GenericError::new("Faucet SRV record cannot be resolved"))?;
+
+            Ok(format!("http://{}/zk/donatex", faucet_host))
+        }
     }
 }
--- a/core/payment-driver/zksync/src/zksync/faucet.rs
+++ b/core/payment-driver/zksync/src/zksync/faucet.rs
@@ -11,16 +11,16 @@ use tokio::time::delay_for;
 
 // Workspace uses
 use ya_payment_driver::{db::models::Network, model::GenericError};
+use ya_utils_networking::srv_resolver;
 
 // Local uses
 use crate::zksync::wallet::account_balance;
 
-const DEFAULT_FAUCET_ADDR: &str = "http://3.249.139.167:5778/zk/donatex";
+const DEFAULT_FAUCET_SRV_PREFIX: &str = "_zk-faucet._tcp.";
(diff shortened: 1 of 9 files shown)


core/payment-driver/zksync/src/zksync/faucet.rs
```diff
@@ -11,16 +11,16 @@ use tokio::time::delay_for;
 
 // Workspace uses
 use ya_payment_driver::{db::models::Network, model::GenericError};
+use ya_utils_networking::srv_resolver;
 
 // Local uses
 use crate::zksync::wallet::account_balance;
 
-const DEFAULT_FAUCET_ADDR: &str = "http://3.249.139.167:5778/zk/donatex";
```
pnowosie, line 18: Can SRV record contain not only IP/name + port, but also part of the path  (`/zk/donatex`)

core/payment-driver/zksync/src/zksync/faucet.rs
```diff
@@ -94,3 +95,12 @@ async fn faucet_donate(address: &str, _network: Network) -> Result<(), GenericEr
     // TODO: Verify tx hash
     Ok(())
 }
+
+async fn resolve_faucet_url() -> Result<String, GenericError> {
+    match env::var(FAUCET_ADDR_ENVAR) {
```
pnowosie, line 100: Is there any significant cost in `env::var(str)` call. It was previously inside `lazy_static` macro :question:

core/payment-driver/zksync/src/zksync/faucet.rs
```diff
@@ -94,3 +95,12 @@ async fn faucet_donate(address: &str, _network: Network) -> Result<(), GenericEr
...
+        Ok(addr) => Ok(addr),
```
pnowosie, line 101: Also this `match` is here because compiler complains when awaiting in the closure (lambda) passed to `or_else`.
What would be better way to write this?

core/payment-driver/zksync/src/zksync/faucet.rs
Wiezzel: Well, not very significant to be honest. I guess it's fine to leave it here since resolving the missing value requires async context.

core/payment-driver/zksync/src/zksync/faucet.rs
Wiezzel: You can try converting `env::var` result into [OptionFuture](https://docs.rs/futures/0.3.12/futures/future/struct.OptionFuture.html) and using asynchronous map function but I'm not sure if it's worth it.

core/payment-driver/zksync/src/zksync/faucet.rs
Wiezzel: I don't know if that's possible. Please verify and adjust the code if not.

utils/networking/src/srv_resolver.rs
```diff
@@ -0,0 +1,29 @@
...
+        .iter()
+        .next()
+        .ok_or_else(|| IoError::from(IoErrorKind::NotFound))?;
+    let addr = format!(
+        "{}:{}",
+        srv.target().to_string().trim_end_matches('.'),
+        srv.port()
+    );
+
+    log::debug!("Central net address: {}", addr);
```
mfranciszkiewicz, line 27: Nitpick: not necessarily "Central net"

Wiezzel (commented):
Reviewed & tested. OK :ballot_box_with_check: 
Please change PR base to `payments/beta-1`

Wiezzel (approved):

PR #1009, opened 2021-02-03
fix exposing version as a metric

etam:
`value!` is treated as a histogram and kept only for 10 seconds.

### DIFF
--- a/utils/compile-time-utils/src/lib.rs
+++ b/utils/compile-time-utils/src/lib.rs
@@ -1,5 +1,5 @@
 use git_version::git_version;
-use metrics::value;
+use metrics::gauge;
 use semver::Version;
 
 /// Returns latest tag (via `git describe --tag --abbrev=0`) or version from Cargo.toml`.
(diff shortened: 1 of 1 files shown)


utils/compile-time-utils/src/lib.rs
```diff
@@ -50,15 +50,15 @@ pub fn semver() -> std::result::Result<Version, semver::SemVerError> {
 
 pub fn report_version_to_metrics() {
     if let Ok(version) = semver() {
-        value!("yagna.version.major", version.major);
-        value!("yagna.version.minor", version.minor);
-        value!("yagna.version.patch", version.patch);
-        value!(
+        counter!("yagna.version.major", version.major);
+        counter!("yagna.version.minor", version.minor);
```
tworec, line 54: `minor` and `patch` are not monotonic.

Excerpt from the counter doc:
https://docs.rs/metrics/0.14.1/metrics/#counters
A counter is a cumulative metric that represents a monotonically increasing value which can only be increased or be reset to zero on restart. For example, you might use a counter to represent the number of operations performed, or the number of errors that have occurred.

Counters are unsigned 64-bit integers.

If you have a value that goes up and down over time, consider using a gauge.

etam: It doesn't change during the lifetime of a single execution, but OK.

tworec (approved):

PR #1014, opened 2021-02-04
Approval races - fix agreements locks

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna-triage/issues/40

I ran 100 requestor in sequence to test it.
I'm not sure if it is reliable test enough, so I added annotation on Graphana, that will warn us if this error happens again.


tworec (approved):
lgtm

Issue #432, opened 2020-07-16
Agreement - wait for approval - rejection & cancellation


nieznanysprawiciel:
- [x] Implement Requestor POST `agreements/{agreement_id}/wait` endpoint
    - [x] Wait for rejection notifications
    - [x] Wait for timeout
    - [x] Wait for cancel Agreement (Requestor can cancel Agreement as long it wasn't approved)

Issue #386, opened 2020-06-29
Cancel Agreement


nieznanysprawiciel:
Does it belong to this milestone??

- [x] Implement DELETE `agreements/{agreement_id}` endpoint
    - [x] Send cancellation to provider

Handling cancellation in wait for agreement will be handled in different task

PR #1044, opened 2021-02-15
Market - Cancel Agreement

nieznanysprawiciel:
resolves: #386
resolves: #432


core/market/src/db/model/agreement_events.rs
```diff
@@ -83,6 +84,12 @@ impl NewAgreementEvent {
                 AgreementState::Approved => AgreementEventType::Approved,
                 AgreementState::Terminated => AgreementEventType::Terminated,
             },
+            // We don't use timestamp from parameter here, because it came from other party
+            // and we use this timestamp for sorting events, when returning them to caller.
+            // On the other side, we should sign AgreementTerminated event together with timestamp,
+            // so we need to have the same value on both nodes.
+            // I don't know, how to solve this problem now, so I leave code that makes possible to
```
jiivan, line 91:
```suggestion
            // I don't know, how to solve this problem now, so I leave code that makes it possible to
```

jiivan (dismissed):

jiivan (approved):

PR #1060, opened 2021-02-18
GlmSync payment driver

Wiezzel:
A clone of the zkSync driver to be used side by side with original.
This is a temporary solution.
To see differences with `zksync` driver run `diff -r core/payment-driver/zksync/ core/payment-driver/glmsync/`

### DIFF
--- a/.env-template
+++ b/.env-template
@@ -46,10 +46,13 @@ YAGNA_DATADIR="."
 
 ## ZkSync driver
 
-# Ethereum chain: rinkeby or mainnet
-#ZKSYNC_CHAIN=rinkeby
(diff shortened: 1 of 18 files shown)


maaktweluit (approved):
LGTM!

Some thoughts:
- Should we also make it the default driver? ( or did i miss this line ? )
- Should we also ping SDK team and provider teams to check their enum values for driver / platform?

Wiezzel:
@maaktweluit 
>     * Should we also make it the default driver? ( or did i miss this line ? )
Not yet.
>     * Should we also ping SDK team and provider teams to check their enum values for driver / platform?
AFAIK provider doesn't need this but SDKs probably need an update.

PR #1071, opened 2021-02-23
Fallback for failing DNS resolution

Wiezzel:
In some setups (e.g. CI tests running on Docker network) Google DNS servers are unavailable and the resolution fails. As a fallback the default DNS resolver could be used by simply passing the original URL into the request.


maaktweluit (approved):
LGTM!

azawlocki (approved):
:+1:

PR #1072, opened 2021-02-23
ZkSync: Unlock wallet before withdrawal

Wiezzel:
Provider's wallet might not be unlocked when `payment exit` is being run. That would cause the command to fail. A call to `unlock_wallet` has been added to fix this.

Bonus: fixed glmSync transaction tracking link.

---
#### ZkSync
1. Run the service
```shell
$ cargo run service run
```

2. Get some funds
```shell
$ cargo run payment fund
Received funds from the faucet. address=0x1264aee4168286ddf962c172a53e0e89a1562e44
```

3. Withdraw
```shell
$ cargo run payment exit
Withdrawal has been accepted by the zkSync operator. It may take some time until the funds are available on Ethereum blockchain. Tracking link: https://rinkeby.zkscan.io/explorer/transactions/ebb12605a72ff4e080b8375e6b9cb77e18548bbbc1882cdb870d65c319907221
```

4. Follow the link to check if withdrawal was successful.

#### GlmSync
1. Run the service
```shell
$ export GLMSYNC_FAUCET_ADDR=http://yacn.dev.golem.network:5777/zk/donate
$ cargo run service run
```

2. Get some funds
```shell
$ cargo run payment fund --driver=glmsync
Received funds from the faucet. address=0x1264aee4168286ddf962c172a53e0e89a1562e44
```

3. Withdraw
```shell
$ cargo run payment exit --driver=glmsync
Withdrawal has been accepted by the zkSync operator. It may take some time until the funds are available on Ethereum blockchain. Tracking link: http://rinkeby.zksync.imapp.pl/explorer/transactions/f8c3d60f10dc6b1715104e28fb1f2445cfecb5617860fa34dc5cf36e95192758
```

4. Follow the link to check if withdrawal was successful.

PR #1077, opened 2021-02-24
ZkSync: Fixed token name in withdraw

Wiezzel:
Withdrawal function used hard-coded `GNT` token which was Rinkeby-specific legacy solution. Because of this withdrawals have not been working on mainnet. Usage of constant has been replaced by a call to `get_network_token` in order to use token ticker appropriate for the given network.

**GlmSync purposely omitted.**

---
**Testing instructions**

Rinkeby -> see instructions in #1072 

Mainnet -> have faith that it works (it would cost ~$30 to test it now)

### DIFF
--- a/core/payment-driver/zksync/examples/withdrawal.rs
+++ b/core/payment-driver/zksync/examples/withdrawal.rs
@@ -68,7 +68,8 @@ async fn main() -> anyhow::Result<()> {
         .parse()
         .expect("Cannot parse 'amount' parameter to BigDecimal");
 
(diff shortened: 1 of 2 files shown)


maaktweluit (approved):
LGTM!

Tested the rinkeby scenario and looks good!

Bit worried about omitting glmsync, the same bug still lives there..

Wiezzel:
@maaktweluit 
> Bit worried about omitting glmsync, the same bug still lives there..

Not need to worry. The fix for glmSync is included in #1080. We're just waiting for the operator to be ready.

PR #1082, opened 2021-02-25
fix for:  error during normal operation

prekucki:

PR #1091, opened 2021-02-28
Fix "Database is locked"

mfranciszkiewicz:
According to https://github.com/diesel-rs/diesel/issues/2365, a part of current connection setup may itself require a lock on the database. This PR sets a timeout prior to executing the remaining `PRAGMA` statements.

Tested with success on a local setup. So far no "Database is locked" errors appeared while testing on `devnet-beta.1`, though it has proven to be difficult due to bugs


mfranciszkiewicz:
Will be solved with an update to #1082

Issue #1100, opened 2021-03-01
agreement stays open after a failing task


tworec:
@maaktweluit writes:
> I had a bug in my requestor script:
https://github.com/golemfactory/hash-cracker/compare/mwu/break-it?expand=1
running this example 3x in a row blocked the whole network
```
YAGNA_APP_KEY=... ./requestor.py

# Wait for 4 subtasks failed, then hit C-c
```

PR #1102, opened 2021-03-01
Provider - fix debit notes property

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/1100

Property name in Agreement differs from property name in Offer since it must be prefixed with "/offer/properties".

### DIFF
--- a/agent/provider/src/payments/model.rs
+++ b/agent/provider/src/payments/model.rs
@@ -35,10 +35,10 @@ impl<'a> PaymentDescription<'a> {
     }
 
     pub fn get_debit_note_deadline(&self) -> Result<Option<chrono::Duration>> {
-        match self
-            .agreement
-            .pointer_typed::<i64>(DEBIT_NOTE_ACCEPT_TIMEOUT_PROPERTY)
-        {
+        match self.agreement.pointer_typed::<i64>(&format!(
+            "/offer/properties{}",
+            DEBIT_NOTE_ACCEPT_TIMEOUT_PROPERTY
+        )) {
             Ok(deadline) => Ok(Some(chrono::Duration::seconds(deadline))),
             Err(Error::NoKey(_)) => Ok(None),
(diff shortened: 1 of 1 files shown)


tworec (approved):

agent/provider/src/market/negotiator/builtin/expiration.rs
```diff
@@ -144,10 +143,12 @@ impl NegotiatorComponent for LimitExpiration {
...
-                    let deadline_prop = offer
-                        .pointer_mut(DEBIT_NOTE_ACCEPT_TIMEOUT_PROPERTY)
-                        .unwrap();
-                    *deadline_prop = serde_json::Value::Number(req_deadline.num_seconds().into());
+                    if let Ok(deadline_prop) = offer.pointer_mut(DEBIT_NOTE_ACCEPT_TIMEOUT_PROP) {
+                        *deadline_prop =
+                            serde_json::Value::Number(req_deadline.num_seconds().into());
+                    } else {
+                        bail!("Unable to set property {}", DEBIT_NOTE_ACCEPT_TIMEOUT_PROP);
+                    }
```
nieznanysprawiciel, line 151: We know that this property exists, because we checked it above (`req_deadline`)

agent/provider/src/market/negotiator/component.rs
```diff
@@ -3,12 +3,22 @@ use serde::{Deserialize, Serialize};
...
 
+impl From<&Proposal> for ProposalView {
+    fn from(p: &Proposal) -> Self {
+        ProposalView {
+            json: expand(p.properties),
+            agreement_id: p.proposal_id,
+        }
+    }
+}
+
```
nieznanysprawiciel, line 21: It should be in AgreementUtils

tworec: I wanted not to touch it this time

PR #1119, opened 2021-03-04
New payment verification mechanism

Wiezzel:
Previous payment verification method was susceptible to tx hash re-using/stealing attack. A new mechanism has been implemented that requires payment drivers to sign a message with payment details with the same key that has been used to sign the blockchain transaction.

Outdated `GetTransactionBalance` endpoint has been replaced with two new endpoints: `SignPayment` and `VerifySignature`. The new endpoints have been implemented in `dummy` and `base` driver crates.

**Task description:** [PAY-53](https://golemproject.atlassian.net/browse/PAY-53)

**Notice:** `erc-20` has been intentionally omitted because the implementation in `base` driver crate should covert it automatically once it is refactored to use that crate.

---
**Testing instructions:** Run `invoice_flow` and `debit_note_flow` with `dummy` and `zksync` drivers.


core/payment-driver/base/src/driver.rs
```diff
@@ -88,4 +83,27 @@ pub trait PaymentDriver {
...
+
+    async fn verify_signature(
+        &self,
+        _db: DbExecutor,
+        _caller: String,
+        msg: VerifySignature,
+    ) -> Result<bool, GenericError> {
+        let payload = utils::payment_hash(&msg.payment);
+        let node_id = msg.payment.payer_id;
+        let signature = bus::sign(node_id, payload).await?;
```
maaktweluit, line 106: isnt `verify_signature` executed on provider where the `payer_id` private key is not available?

maaktweluit: if above thought is correct you can use something like: https://github.com/tomusdrw/ethsign/blob/9ff18e158dda29b13558c04f77441d93c7bbce57/src/key.rs#L83

maaktweluit (changes requested):
LGTM!

One possibly "breaking" comment, please have a look, will approve on reply / fix

core/payment-driver/base/src/driver.rs
Wiezzel: :facepalm: Obviously! I'll fix that.

Wiezzel: Fixed.

maaktweluit (approved):
LGTM!

Wiezzel:
@maaktweluit Plz, approve again

maaktweluit (approved):

pnowosie (approved):
:+1:

PR #1133, opened 2021-03-09
Fix: Change metric from driver to platform

nieznanysprawiciel:


core/payment/src/processor.rs
```diff
@@ -388,7 +388,7 @@ impl PaymentProcessor {
             activity_payment.allocation_id = None;
         }
 
-        counter!("payment.amount.sent", ya_metrics::utils::cryptocurrency_to_u64(&msg.amount), "driver" => msg.driver);
+        counter!("payment.amount.sent", ya_metrics::utils::cryptocurrency_to_u64(&msg.amount), "platform" => msg.driver);
```
Wiezzel, line 391: The value should be `payment_platform` not `msg.driver` then.

Wiezzel (changes requested):

core/payment/src/processor.rs
nieznanysprawiciel: Thanks for noticing

Wiezzel: Now you're using a moved value. Compile before committing. :P

nieznanysprawiciel: Eh. I hoped it will be easy fix :P

Wiezzel (approved):

PR #1135, opened 2021-03-09
[Don't merge] Increase connection establish timeout in zksync wallet

nieznanysprawiciel:
This PR only shows solution to the problem, but I'm not sure if we should merge this as fix


mfranciszkiewicz:
@nieznanysprawiciel please re-create the PR if the changes are still needed

PR #1139, opened 2021-03-10
Filter version and pre-release tags

jiivan:


tworec (approved):

utils/compile-time-utils/src/lib.rs
```diff
@@ -2,9 +2,17 @@ use git_version::git_version;
...
-/// Returns latest tag (via `git describe --tag --abbrev=0`).
+/// Returns latest version tag.
 pub fn git_tag() -> &'static str {
-    git_version!(args = ["--tag", "--abbrev=0"], cargo_prefix = "")
+    git_version!(
+        args = [
+            "--tag",
+            "--abbrev=0",
+            "--match=v[0-9]*",
+            "--match=pre-rel-[0-9]*"
```
tworec, line 12: It should be `pre-rel-v[0-9]*`

PR #1144, opened 2021-03-10
Payment service: fixed race condition in PaymentProcessor

Wiezzel:
If `NotifyPayment` got called during processing of `SchedulePayment` (which happened when running `debit_note_flow` example with dummy driver) it would fail because the payment order had not been saved in the database yet.

To avoid this race condition (and possibly others) `PaymentProcessor` has been wrapped in a mutex. This allowed to remove the mutex from `DriverRegistry`. Also, sending `SendPayment` message needs to be done via `Arbiter::spawn` now to avoid deadlock in case it is sent to the same node.

---
Testing instructions: Run `debit_note_flow` with dummy driver. (A few times preferably)

### DIFF
--- a/core/payment/src/processor.rs
+++ b/core/payment/src/processor.rs
@@ -4,12 +4,12 @@ use crate::error::processor::{
     SchedulePaymentError, ValidateAllocationError, VerifyPaymentError,
 };
 use crate::models::order::ReadObj as DbOrder;
+use actix_web::rt::Arbiter;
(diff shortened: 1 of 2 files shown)


pnowosie (approved):
Awesome work!
...and dummy driver for the win 🎉 
``` 
... debit_note_flow]  👍🏻 Example completed successfully ❤️
```

maaktweluit (approved):
LGTM!

Tested the flows with both drivers ( erc20 is currently broken )

PR #1182, opened 2021-03-24
Provider - send less DebitNotes

nieznanysprawiciel:


tworec (approved):

agent/provider/src/payments/pricing.rs
```diff
@@ -82,7 +82,7 @@ pub struct LinearPricingOffer {
 
 impl Default for LinearPricingOffer {
     fn default() -> Self {
-        LinearPricingOffer { interval: 6.0 }
+        LinearPricingOffer { interval: 120.0 }
```
tworec, line 85: it is also hardcoded in the doc or comment somewhere... pls update all places where `6s` is mentioned

tworec (changes requested):
see my comment

tworec (dismissed):

agent/provider/src/payments/pricing.rs
nieznanysprawiciel: Yes but docs mention this only as example, so I wouldn't touch. There are tests which have this value, but I don't think they should be changes either

tworec (approved):

PR #1184, opened 2021-03-25
ZkSync: Check balance before operations

Wiezzel:
Check account balance before making a zkSync account unlock or withdrawal. This allows to display better error messages to the user and avoid panic in the `withdraw` function (`balance` and `withdraw_fee` are of `BigUint` type so `balance - withdraw_fee` would cause panic in case the fee is bigger than the account balance).

---
Testing intructions:

1. Try init in send mode without funds.
```
$ cargo run payment init --sender
Error: GenericError { inner: "Not enough balance to unlock account" }  HINT: Did you run `yagna payment fund` and follow the instructions?
```
2. Try exit without funds.
```
$ cargo run payment exit
Error: Not enough balance to exit. Minimum required balance to pay withdraw fees is 0.02957 tGLM
$ cargo run payment exit --network=mainnet
Error: Not enough balance to exit. Minimum required balance to pay withdraw fees is 55.36 GLM
```
3. Regression (init and exit **with funds**).

### DIFF
--- a/core/payment-driver/zksync/src/zksync/wallet.rs
+++ b/core/payment-driver/zksync/src/zksync/wallet.rs
@@ -9,7 +9,11 @@ use std::env;
 use std::str::FromStr;
 use zksync::operations::SyncTransactionHandle;
 use zksync::types::BlockStatus;
-use zksync::zksync_types::{tx::TxHash, Address, Nonce, TxFeeTypes};
+use zksync::zksync_types::{
+    tokens::{ChangePubKeyFeeType, ChangePubKeyFeeTypeArg},
+    tx::TxHash,
+    Address, Nonce, TxFeeTypes,
+};
 use zksync::{
     provider::get_rpc_addr,
     provider::{Provider, RpcProvider},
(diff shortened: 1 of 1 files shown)


core/payment-driver/zksync/src/zksync/wallet.rs
```diff
@@ -315,35 +345,29 @@ pub async fn withdraw<S: EthereumSigner + Clone, P: Provider + Clone>(
...
-        .map_err(GenericError::new)?
-        .total_fee;
+    let withdraw_fee = get_withdraw_fee(&wallet, &token).await?;
     info!(
         "Withdrawal transaction fee {:.5} {}",
         utils::big_uint_to_big_dec(withdraw_fee.clone()),
         token
     );
+    if withdraw_fee > balance {
+        return Err(GenericError::new("Not enough balance to withdraw"));
```
maaktweluit, line 363: maybe print balance + required amounts in this error too?

maaktweluit (approved):
LGTM!

One nitty comment, still approving!

core/payment-driver/zksync/src/zksync/wallet.rs
Wiezzel: This error should actually never be raised because it's already checked in `exit`. This is just to double-check.

Wiezzel: From exit you will get nice messages like this:
```
Error: Not enough balance to exit. Minimum required balance to pay withdraw fees is 0.02957 tGLM
```

maaktweluit: Neat 👍

maaktweluit:
Tested this PR with the payment_api examples and `yapapi`

PR #1198, opened 2021-03-30
Removed ethkey 

maaktweluit:
and used the best ethsign version to match other dependencies

Total reduction of crates installed on a `cargo build`: 
- release 0.6.3 = 738
- this branch = 668


core/payment/examples/payment_api.rs
```diff
@@ -89,9 +92,9 @@ pub async fn start_dummy_driver() -> anyhow::Result<()> {
 
 pub async fn start_erc20_driver(
     db: &DbExecutor,
-    requestor_account: Box<EthAccount>,
+    requestor_account: Box<SecretKey>,
```
Wiezzel, line 95: No need to use `Box` here anymore. The only reason for doing that was that `EthAccount::load_or_generate` returned a `Box`.

core/payment/examples/payment_api.rs
```diff
@@ -104,9 +107,9 @@ pub async fn start_erc20_driver(
 
 pub async fn start_zksync_driver(
     db: &DbExecutor,
-    requestor_account: Box<EthAccount>,
+    requestor_account: Box<SecretKey>,
```
Wiezzel, line 110: Ditto.

Wiezzel (commented):
It compiles fine with `ethsign = "0.8"`. Is it that `zksync_eth_client` is using another version and we want to avoid having to version of the same library in our dependencies?

Plz, fix formatting.

Wiezzel (approved):

PR #1211, opened 2021-04-07
small fixes to broadcasts.

### LINKED ISSUE #1205: Fix broadcasts
- [x] Use `try_lock` in broadcast received handler
- [x] Don't  send empty broadcast
- [x] Fix computing interval

maaktweluit:
- random interval 0.5-1.5x instead of 0-2x
- do not send empty broadcasts
- do not wait for the offer_bcast lock, skip handling instead

Resolves #1205

### DIFF
--- a/core/market/src/protocol/discovery.rs
+++ b/core/market/src/protocol/discovery.rs
@@ -49,6 +49,9 @@ impl Discovery {
     /// Broadcasts Offers to other nodes in network. Connected nodes will
     /// get call to function bound at `OfferBcast`.
     pub async fn bcast_offers(&self, offer_ids: Vec<SubscriptionId>) -> Result<(), DiscoveryError> {
+        if offer_ids.is_empty() {
+            return Ok(());
+        }
         let default_id = self.default_identity().await?;
         let bcast_msg = SendBroadcastMessage::new(OffersBcast { offer_ids });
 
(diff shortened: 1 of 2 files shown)


core/market/src/protocol/discovery.rs
```diff
@@ -147,8 +153,9 @@ impl Discovery {
 
     async fn on_bcast_offers(self, caller: String, msg: OffersBcast) -> Result<(), ()> {
         let num_ids_received = msg.offer_ids.len();
-        if !msg.offer_ids.is_empty() {
-            log::trace!("Received {} Offers from [{}].", num_ids_received, &caller);
```
nieznanysprawiciel, line 151: As far as I remember this log was only for non empty cases, because it was to much spam. But maybe it is not a problem as long it is on trace, so you can ignore this comment if you want.

core/market/src/protocol/discovery.rs
```diff
@@ -216,12 +229,13 @@ impl Discovery {
         msg: UnsubscribedOffersBcast,
     ) -> Result<(), ()> {
         let num_received_ids = msg.offer_ids.len();
-        if !msg.offer_ids.is_empty() {
-            log::trace!(
-                "Received {} unsubscribed Offers from [{}].",
-                num_received_ids,
-                &caller,
-            );
+        log::trace!(
```
nieznanysprawiciel, line 232: As far as I remember this log was only for non empty cases, because it was to much spam. But maybe it is not a problem as long it is on trace, so you can ignore this comment if you want.

nieznanysprawiciel (approved):

core/market/src/protocol/discovery.rs
maaktweluit: correct, but now it SHOULD not send empty bcasts again so we should see it when it does happen

tworec (approved):
Goooodie

PR #1237, opened 2021-04-15
[mkt tst] adapt test cyclic broadcasts for uncertainty 

### LINKED ISSUE #1222: reduce the number of subscribes in test_cyclic_broadcasts
We need to reduce the number of subscribes being processed, because it seems CI machines @ GH are not able to process such an amount of data within 100ms. (production code has broadcasts every 4mins which should suffice even for weak machines).

_Originally posted by @tworec in https://github.com/golemfactory/yagna/pull/1203#r611873962_

tworec:
resolves #1222

### DIFF
--- a/core/market/src/testing/mock_node.rs
+++ b/core/market/src/testing/mock_node.rs
@@ -110,6 +110,10 @@ impl MarketsNetwork {
     /// It will be used to create directories and GSB binding points,
     /// to avoid potential name clashes.
     pub async fn new(test_name: Option<&str>) -> Self {
+        // DIRTY HACK. We need it because GSB v0.4.4 causes market test suite to hang
+        // using 100% CPU trying to reconnect in the infinite loop.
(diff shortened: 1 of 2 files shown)


tworec:
DIRTY HACK can be removed when https://github.com/golemfactory/ya-service-bus/issues/22 is adressed

core/market/src/testing/mock_node.rs
```diff
@@ -110,6 +110,7 @@ impl MarketsNetwork {
     /// It will be used to create directories and GSB binding points,
     /// to avoid potential name clashes.
     pub async fn new(test_name: Option<&str>) -> Self {
+        std::env::set_var("GSB_URL", "wrong url");
```
jiivan, line 113: Could you add a comment why is this needed?

tworec: good catch! thx. Will do

jiivan (approved):

PR #1256, opened 2021-04-22
Add queue for offer and unsubscribe broadcasts

maaktweluit:
- added queue for sending broadcasts so they are rate-limited to 5 seconds per broadcast.

### DIFF
--- a/core/market/Cargo.toml
+++ b/core/market/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 [features]
 test-suite = []
 bcast-singleton = []
-testing = ["actix-http", "actix-rt", "actix-service", "env_logger"]
+testing = ["actix-http", "actix-service", "env_logger"]
 
 [dependencies]
 ya-agreement-utils = { version = "^0.2"}
@@ -55,7 +55,7 @@ uuid = { version = "0.8", features = ["v4"] }
 
 # testing
 actix-http = { version = "2.2", optional = true }
-actix-rt = { version = "1.0.0", optional = true }
+actix-rt = { version = "1.0.0" }
 actix-service = { version = "1.0", optional = true }
 env_logger = { version = "0.7", optional = true }
 regex = "1.4.2"
--- a/core/market/src/protocol/discovery.rs
+++ b/core/market/src/protocol/discovery.rs
@@ -52,10 +52,10 @@ pub struct DiscoveryImpl {
 
 impl Discovery {
     pub async fn bcast_offers(&self, offer_ids: Vec<SubscriptionId>) -> Result<(), DiscoveryError> {
-        log::trace!("bcast_offers {}", offer_ids.len());
         if offer_ids.is_empty() {
             return Ok(());
         }
+        // When there are 0 items in the queue we should schedule a send job.
         let must_schedule = {
             let mut queue = self.inner.offer_queue.lock().await;
             let result = queue.len() == 0;
@@ -64,13 +64,16 @@ impl Discovery {
             result
         };
         log::trace!(
-            "bcast_offers done appending. must_schedule? {}",
+            "bcast_offers done appending {} offers. must_schedule={}",
+            offer_ids.len(),
             must_schedule
         );
 
         if must_schedule {
             let myself = self.clone();
             let _ = Arbiter::spawn(async move {
+                // Sleep to collect multiple offers to send
+                delay_for(Duration::from_secs(5)).await;
                 myself.send_bcast_offers().await;
             });
         }
@@ -80,13 +83,15 @@ impl Discovery {
     /// Broadcasts Offers to other nodes in network. Connected nodes will
     /// get call to function bound at `OfferBcast`.
     async fn send_bcast_offers(&self) -> () {
-        delay_for(Duration::from_secs(5)).await;
+        // `...offer_queue` MUST be empty to trigger the sending again
         let offer_ids: Vec<SubscriptionId> =
             self.inner.offer_queue.lock().await.drain(0..).collect();
-        log::trace!("send_bcast_offers. {}", offer_ids.len());
+
+        // Should never happen, but just to be certain.
         if offer_ids.is_empty() {
             return ();
         }
+
         let default_id = match self.default_identity().await {
             Ok(id) => id,
             Err(e) => {
@@ -97,6 +102,7 @@ impl Discovery {
                 return;
             }
         };
+        log::debug!("Broadcasting offers. count={}", offer_ids.len());
         let bcast_msg = SendBroadcastMessage::new(OffersBcast { offer_ids });
 
         // TODO: We shouldn't use send_as. Put identity inside broadcasted message instead.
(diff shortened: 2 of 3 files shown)


core/market/src/protocol/discovery.rs
```diff
@@ -94,18 +140,57 @@ impl Discovery {
...
+        log::trace!("bcast_unsubscribes {}", offer_ids.len());
         if offer_ids.is_empty() {
             return Ok(());
         }
-        let default_id = self.default_identity().await?;
+        let must_schedule = {
+            let mut queue = self.inner.unsub_queue.lock().await;
+            let result = queue.len() == 0;
+
+            queue.append(&mut offer_ids.clone());
```
nieznanysprawiciel, line 151: Do we need clone?

core/market/src/protocol/discovery.rs
```diff
@@ -40,26 +42,70 @@ pub(super) struct OfferHandlers {
...
+        log::trace!("bcast_offers {}", offer_ids.len());
         if offer_ids.is_empty() {
             return Ok(());
         }
-        let default_id = self.default_identity().await?;
+        let must_schedule = {
+            let mut queue = self.inner.offer_queue.lock().await;
+            let result = queue.len() == 0;
+
+            queue.append(&mut offer_ids.clone());
```
nieznanysprawiciel, line 63: Do we need clone?

core/market/src/protocol/discovery.rs
```diff
@@ -94,18 +140,57 @@ impl Discovery {
...
+            result
+        };
+        log::trace!(
+            "bcast_unsubscribes done appending. must_schedule? {}",
+            must_schedule
+        );
+
+        if must_schedule {
+            let myself = self.clone();
+            let _ = Arbiter::spawn(async move {
```
nieznanysprawiciel, line 161: Is it enough to just spawn without delay?

nieznanysprawiciel (approved):

core/market/src/protocol/discovery.rs
maaktweluit: i think so to make it `&mut`

core/market/src/protocol/discovery.rs
maaktweluit: that would defeat the whole point, the bcast's are send as they are called

nieznanysprawiciel: No. I meant spawn closure with delay (very short) inside to wait for some other subscribes. Because I don't understand why it will batch broadcasts. They must be really close and we need lack with scheduling

nieznanysprawiciel (dismissed):

core/market/src/protocol/discovery.rs
maaktweluit: you are correct, the delay was missing and i was not aware of that.

Thanks for pointing it out! great catch :)

Fixed :D

nieznanysprawiciel (dismissed):

nieznanysprawiciel (approved):

core/market/Cargo.toml
```diff
@@ -55,7 +55,7 @@ uuid = { version = "0.8", features = ["v4"] }
 
 # testing
 actix-http = { version = "2.2", optional = true }
-actix-rt = { version = "1.0.0", optional = true }
+actix-rt = { version = "1.0.0" }
```
tworec, line 58: move it above pls, this is `#testing` section

core/market/src/protocol/discovery.rs
```diff
@@ -40,26 +42,76 @@ pub(super) struct OfferHandlers {
...
+            "bcast_offers done appending {} offers. must_schedule={}",
+            offer_ids.len(),
+            must_schedule
+        );
+
+        if must_schedule {
+            let myself = self.clone();
+            let _ = Arbiter::spawn(async move {
+                // Sleep to collect multiple offers to send
+                delay_for(Duration::from_secs(5)).await;
```
tworec, line 76: would be great to have this constant named, and event better to have it configurable. I'm afraid without configurability many tests from the market test suite will not pass because we use delays no longer than 5s most of the time

core/market/src/protocol/discovery.rs
```diff
@@ -40,26 +42,76 @@ pub(super) struct OfferHandlers {
...
+        }
+        Ok(())
+    }
+
+    /// Broadcasts Offers to other nodes in network. Connected nodes will
+    /// get call to function bound at `OfferBcast`.
+    async fn send_bcast_offers(&self) -> () {
+        // `...offer_queue` MUST be empty to trigger the sending again
+        let offer_ids: Vec<SubscriptionId> =
+            self.inner.offer_queue.lock().await.drain(0..).collect();
```
tworec, line 88: nitpickl ;)
```suggestion
            self.inner.offer_queue.lock().await.drain(..).collect();
```

core/market/src/protocol/discovery.rs
```diff
@@ -40,26 +42,76 @@ pub(super) struct OfferHandlers {
...
+        }
+
+        let default_id = match self.default_identity().await {
+            Ok(id) => id,
+            Err(e) => {
+                log::error!(
+                    "Error getting default identity, not sending bcast. error={:?}",
+                    e
+                );
+                return;
```
tworec, line 102: We have drained the queue, but not sending contents. Pls change the order of those commands: first, try to get default identity, and then drain the queue and then send.

core/market/src/protocol/discovery.rs
```diff
@@ -40,26 +42,76 @@ pub(super) struct OfferHandlers {
...
 
         // TODO: We shouldn't use send_as. Put identity inside broadcasted message instead.
-        let _ = bus::service(local_net::BUS_ID)
+        if let Err(e) = bus::service(local_net::BUS_ID)
             .send_as(default_id, bcast_msg) // TODO: should we send as our (default) identity?
-            .await?;
-        Ok(())
+            .await
+        {
+            log::error!("Error sending bcast, skipping... error={:?}", e);
```
tworec, line 113: shouldn't we put offer ids back into queue here?

tworec (changes requested):

core/market/src/protocol/discovery.rs
maaktweluit: then the job will not trigger again ( or we go into a recursive loop )

core/market/src/protocol/discovery.rs
maaktweluit: i like the idea! but it has the same "problem" as above

core/market/Cargo.toml
maaktweluit: good catch! done

core/market/src/protocol/discovery.rs
maaktweluit: fixed :) will push all changes together

core/market/src/protocol/discovery.rs
maaktweluit: good idea! working on it, posted a q on our team chat

maaktweluit:
Created a new PR so the market suite would trigger.. 

All comments are resolved in a separate commit there: https://github.com/golemfactory/yagna/pull/1263/commits/4649a88dbe13908a91c051b5e1cd7d44e83079bc

Issue #1132, opened 2021-03-09
Integration tests for Provider - breaking Agreements


nieznanysprawiciel:
Why:
- We need tests for Provider after introducing keep alive functionality

What:
- [x] Breaking in case of Agreement without activities
- [x] Breaking if Agreement is idle after some activities were already created
- [x] Breaking if DebitNotes aren't accepted
- [x] Breaking if Requestor yagna daemon is unreachable
- [x] Test if provider doesn't allow to create more than 1 activity at the same time

PR #1259, opened 2021-04-22
Test provider breaking Agreement

nieznanysprawiciel:
resolves: #1132


goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,260 @@
...
+from goth.runner.probe import RequestorProbe, ProviderProbe
+
+from goth_tests.helpers.activity import run_activity, wasi_exe_script, wasi_task_package
+from goth_tests.helpers.negotiation import negotiate_agreements, DemandBuilder
+from goth_tests.helpers.payment import pay_all
+
+logger = logging.getLogger(__name__)
+
+
+def _topology(assets_path: Path) -> List[YagnaContainerConfig]:
```
zakaprov, line 27: We're moving away from defining the topology in code. This can cause all sort of problems for you, especially when mixing configuration modes (i.e. using a topology and a config file in one test).
One example of an issue is CLI config overrides won't work properly (e.g. to use local `yagna` binaries instead of default ones).

Instead, you should use either the default config with some overrides (you can take a look at e.g. `test_zero_amount_txs.py`) or create a new `goth-config.yml` based on the default one.

If you need to set environment variables we have a PR which adds support for this in `goth-config.yml`: https://github.com/golemfactory/goth/pull/505

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,260 @@
...
+
+    return Runner(
+        base_log_dir=log_dir,
+        compose_config=goth_config.compose_config,
+        web_root_path=common_assets / "web-root",
+    )
+
+
+# Provider is expected to break Agreement in time configured by
+# variable: IDLE_AGREEMENT_TIMEOUT, if there are no Activities created.
```
zakaprov, line 109: Please move this comment to the docstring of this test function.

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,260 @@
...
+        # Break after 5s + 3s margin
+        await providers[0].wait_for_agreement_broken(r"No activity created", timeout=8)
+
+        await pay_all(requestor, agreement_providers)
+
+
+# Provider is expected to break Agreement, if no new Activity was created
+# after time configured by variable: IDLE_AGREEMENT_TIMEOUT.
+# This test checks case, when Requestor already computed some Activities,
+# but orphaned Agreement at some point.
```
zakaprov, line 138: Same as above, please move this to the test's docstring.

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,260 @@
...
+                requestor, provider, agreement_id, wasi_exe_script(runner)
+            )
+
+        # Break after 5s + 3s margin
+        await providers[0].wait_for_agreement_broken("No activity created", timeout=8)
+
+        await pay_all(requestor, agreement_providers)
+
+
+# Requestor is expected to accept DebitNotes in timeout negotiated in Offer.
```
zakaprov, line 173: ditto

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,260 @@
...
+            "Requestor isn't accepting DebitNotes in time",
+            timeout=12,
+        )
+
+        await pay_all(requestor, agreement_providers)
+
+
+# If Provider is unable to send DebitNotes for some period of time, he should
+# break Agreement. This is separate mechanism from DebitNotes keep alive, because
+# here we are unable to send them, so they can't timeout.
```
zakaprov, line 214: ditto

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -14,45 +16,89 @@
...
 
 
-@pytest.mark.asyncio
-async def test_provider_multi_activity(
-    common_assets: Path,
-    config_overrides: List[Override],
-    log_dir: Path,
-):
-    """Test provider handling multiple activities in single Agreement."""
+def _topology(assets_path: Path) -> List[YagnaContainerConfig]:
```
zakaprov, line 30: Like in the case of `test_provider_break_agreement.py`, please use the new way of configuring nodes. This test was already using a config override, let's go back to that.

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -14,45 +16,89 @@
...
         base_log_dir=log_dir,
         compose_config=goth_config.compose_config,
         web_root_path=common_assets / "web-root",
     )
 
-    async with runner(goth_config.containers):
+
+# Tests running multiple activities on single Provider.
+# In this case Requestor is responsible for terminating Agreement.
+# Provider should listen
```
zakaprov, line 80: Please move this to the docstring.

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -91,3 +136,65 @@ async def test_provider_multi_activity(
 
         # Payment
         await pay_all(requestor, agreement_providers)
+
+
+# Provider is expected to reject second activity, if one is already running.
```
zakaprov, line 141: ditto

goth_tests/pyproject.toml
```diff
@@ -20,9 +20,9 @@ readme = "README.md"
 python = "^3.8.0"
 pytest = "^6.2"
 pytest-asyncio = "^0.14"
-goth = "^0.3"
+#goth = "^0.3"
```
zakaprov, line 23: Don't forget to revert this once the version with required changes is released.

zakaprov (changes requested):

goth_tests/domain/ya-provider/test_provider_break_agreement.py
zakaprov: The config feature I mentioned is now merged to `goth` master, I'm going to make a release including that change today .

nieznanysprawiciel: Great. I'm working on this, but we have feature freeze and I have to postpone it.

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -91,3 +101,67 @@ async def test_provider_multi_activity(
...
+            await requestor.create_activity(agreement_id)
+
+            assert (
+                re.match(
+                    r"terminated. Reason: Only single Activity allowed,"
+                    r" message: Can't create 2 simultaneous Activities.",
+                    e.value.body,
+                )
+                is not None
+            )
```
azawlocki, line 161:
```suggestion
        with pytest.raises(ApiException) as e:
            await requestor.create_activity(agreement_id)

        assert (
            re.match(
                r"terminated. Reason: Only single Activity allowed,"
                r" message: Can't create 2 simultaneous Activities.",
                e.value.body,
            )
        )
```
If there's an error in `requestor.create_activity(agreement_id)` then the control exits the scope of `with pytest.raises(...)` and the assertion is not checked.

azawlocki (changes requested):
I have only comment, about an `assert` incorrectly placed within the scope of `with pytest.raises(...)`

goth_tests/domain/ya-provider/test_provider_multi_activity.py
nieznanysprawiciel: Thanks, good catch :)

nieznanysprawiciel: Good you noticed, because this regex was invalid and failed to match ;)

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -23,14 +20,9 @@
...
-    log_dir: Path,
-):
-    """Test provider handling multiple activities in single Agreement."""
-
+def _create_runner(
+    common_assets: Path, config_overrides: List[Override], log_dir: Path
+) -> Tuple[Runner, Configuration]:
     nodes = [
         {"name": "requestor", "type": "Requestor"},
         {"name": "provider-1", "type": "VM-Wasm-Provider", "use-proxy": True},
```
zakaprov, line 28: This override is no longer needed since you've created a custom config file (`goth_tests/domain/ya-provider/assets/goth-config.yml`). However, that file is not being used at the moment!

in the current implementation (line 32 in this revision, I cannot reach it with a PR comment)  `common_assets` is used as the base path for the config file. This `common_assets` is provided by a fixture and always resolves to `goth_tests/assets` (relative to the project root).
Instead, to use the "local" config file this should look something like this:
```
    goth_config = load_yaml(Path(__file__).parent / "assets" / "goth-config.yml", config_overrides)
```

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -91,3 +101,64 @@ async def test_provider_multi_activity(
 
         # Payment
         await pay_all(requestor, agreement_providers)
+
+
+@pytest.mark.asyncio
+async def test_provider_single_activity_at_once(
```
zakaprov, line 107:
```suggestion
async def test_provider_single_simultaneous_activity(
```

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -91,3 +101,64 @@ async def test_provider_multi_activity(
...
+        agreement_providers = await negotiate_agreements(
+            requestor,
+            demand,
+            providers,
+        )
+
+        #  Activity
+        agreement_id, provider = agreement_providers[0]
+
+        activity_id1 = await requestor.create_activity(agreement_id)
```
zakaprov, line 148:
```suggestion
        first_activity_id = await requestor.create_activity(agreement_id)
```

goth_tests/domain/ya-provider/test_provider_multi_activity.py
```diff
@@ -91,3 +101,64 @@ async def test_provider_multi_activity(
...
+        with pytest.raises(ApiException) as e:
+            await requestor.create_activity(agreement_id)
+
+        assert re.search(
+            r"terminated. Reason: Only single Activity allowed,"
+            r" message: Can't create 2 simultaneous Activities.",
+            e.value.body,
+        )
+
+        await requestor.destroy_activity(activity_id1)
```
zakaprov, line 160:
```suggestion
        await requestor.destroy_activity(first_activity_id)
```

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,230 @@
+"""End to end tests for requesting WASM tasks using goth REST API clients."""
```
zakaprov, line 1:
```suggestion
"""End to end tests for various scenarios which involve breaking agreements."""
```

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,230 @@
...
+        .property("golem.srv.caps.multi-activity", True)
+        .constraints(
+            "(&(golem.com.pricing.model=linear)\
+            (golem.srv.caps.multi-activity=true)\
+            (golem.runtime.name=wasmtime))"
+        )
+    )
+
+    if require_debit_notes:
+        demand = demand.property("golem.com.payment.debit-notes.accept-timeout?", 8)
```
zakaprov, line 51: Let's move this magical number `8` to some named "constant".

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,230 @@
...
+    return demand.build()
+
+
+def _create_runner(
+    common_assets: Path, config_overrides: List[Override], log_dir: Path
+) -> Tuple[Runner, Configuration]:
+    goth_config = load_yaml(
+        common_assets / ".." / "domain" / "ya-provider" / "assets" / "goth-config.yml",
```
zakaprov, line 59: Please take a look at the comment regarding the same function in `test_provider_multi_activity.py`.

goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -0,0 +1,230 @@
...
+        await pay_all(requestor, agreement_providers)
+
+
+@pytest.mark.asyncio
+async def test_provider_timeout_unresponsive_requestor(
+    common_assets: Path,
+    config_overrides: List[Override],
+    log_dir: Path,
+):
+    """Test provider breaking Agreement if Requestor doesn't accept DebitNotes.
```
zakaprov, line 188:
```suggestion
    """Test provider breaking Agreement if Requestor goes offline.
```

goth_tests/domain/ya-provider/test_provider_break_agreement.py
nieznanysprawiciel: It won't help too much. This property name already looks like named constant

zakaprov: meh, true :(

azawlocki (dismissed):
LGTM!

zakaprov (approved):

tworec (approved):
Didn't read but accepting since both @zakaprov and @azawlocki accepted

PR #1263, opened 2021-04-23
Add queue for offer and unsubscribe broadcasts

### LINKED ISSUE #1260: Optimise broadcast sending
why:
- bcasts are still flooding the network
- try-lock is skipping too many broadcasts

what:
- add rate-limiting queue so bcases are grouped
- make each node only send bcasts every 5 seconds

maaktweluit:
added queue for sending broadcasts so they are rate-limited to 5 seconds per broadcast.

Re-created from https://github.com/golemfactory/yagna/pull/1256 to trigger market test suite

Resolves #1260

### DIFF
--- a/core/market/Cargo.toml
+++ b/core/market/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 [features]
 test-suite = []
 bcast-singleton = []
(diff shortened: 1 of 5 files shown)


nieznanysprawiciel (dismissed):

maaktweluit:
Closing since these fixes will be included in #1265

PR #1265, opened 2021-04-24
Add queue for broadcasts + reshape market test suite config

### LINKED ISSUE #1260: Optimise broadcast sending

tworec:
Resolves #1260
This is a continuation of @maaktweluit work started in #1263 :
> Added queue for sending broadcasts so they are rate-limited to 5 seconds per broadcast.

I branched off, because branch [/market/mwu/queue-bcast](https://github.com/golemfactory/yagna/tree/market/mwu/queue-bcast) has compilation errors.
I'm proposing a fix for them here, by making `DiscoveryBuilder` API backward compatible, not to be forced to change existing code more than needed (tests mainly).

At the same time, I'm removing the `Default` impl for `DiscoveryConfig` and creating this config for market test suite specifically.
I'm fixing other tests too, which are failing because of the introduction of nondeterministic (via `try_lock` @ #1211 ) broadcast reception handling.

And btw. I'm applying build optimization for the macOS platform.


core/market/src/testing/mock_node.rs
```diff
@@ -688,3 +684,83 @@ pub mod default {
...
+        offer_rebroadcast_delay: Duration::from_millis(200),
+        unsub_rebroadcast_delay: Duration::from_millis(200),
+    };
+
+    Config {
+        discovery,
+        subscription: Default::default(),
+        events: Default::default(),
+    }
+}
```
nieznanysprawiciel, line 704: 👍🏻

core/market/tests/test_cyclic_broadcasts.rs
```diff
@@ -17,14 +16,8 @@ use ya_market::testing::{
 async fn test_startup_offers_sharing() {
     let _ = env_logger::builder().try_init();
 
-    // Change expected time of sending broadcasts.
-    let mut config = Config::default();
-    config.discovery.mean_cyclic_bcast_interval = Duration::from_millis(100);
-    config.discovery.max_bcasted_offers = 7;
-
```
nieznanysprawiciel, line 24: This was for purpose, please don't remove here and in any tests that creates config.
If config isn't default, that means, that test depends on values that are set here. You shouldn't use default config, that's values can change at any point in time

Cargo.toml
```diff
@@ -201,3 +201,8 @@ ya-utils-process = { path = "utils/process"}
 ya-diesel-utils = { path = "utils/diesel-utils"}
 ya-metrics = { path = "core/metrics" }
 ya-provider = { path = "agent/provider"}
+
+# Speed up builds on macOS (will be default in next rust version probably)
+# https://jakedeichert.com/blog/reducing-rust-incremental-compilation-times-on-macos-by-70-percent/
+[profile.dev]
+split-debuginfo = "unpacked"
```
maaktweluit, line 208: missing whiteline at EoF is a bad practice in git.

You can configure your editor to automagically add those on `git` tracked files

core/market/src/config.rs
```diff
@@ -22,17 +22,18 @@ pub struct DiscoveryConfig {
     #[structopt(env, parse(try_from_str = humantime::parse_duration), default_value = "4min")]
     pub mean_cyclic_unsubscribes_interval: Duration,
     #[structopt(env, parse(try_from_str = humantime::parse_duration), default_value = "5sec")]
-    pub offer_broadcast_time: Duration,
+    pub offer_rebroadcast_delay: Duration,
```
maaktweluit, line 25: its not only re-broadcast, but all broadcasts

+1 for time -> delay

core/market/tests/test_cyclic_broadcasts.rs
tworec: As you saw, I've created a dedicated config for **all** market suite tests (vide `create_market_config_for_test`).
It is crafted in a way, which does not require special overwrites within any particular test case.

Is that ok?

maaktweluit (changes requested):
LGTM!

2 nitty comments, will approve on reply :)

core/market/tests/test_cyclic_broadcasts.rs
tworec: I've also moved test helper code which is waiting for broadcasts near afromentioned config creation
See `fn assert_offers_broadcasted`

Cargo.toml
tworec: done

core/market/src/config.rs
tworec: done

maaktweluit (dismissed):

core/market/tests/test_cyclic_broadcasts.rs
nieznanysprawiciel: > As you saw, I've created a dedicated config for **all market suite tests** (vide `create_market_config_for_test`).
> It is crafted in a way, which does not require special overwrites within any particular test case.
> 
> Is that ok?

Most of the tests can work with any value from this config. These were prepared specifically for the value that were set here. When you change this general config in future, you are not aware, that you are breaking these tests. So I prefer to set them explicite here

tworec: We agreed on that during video call

maaktweluit (approved):

nieznanysprawiciel (approved):

PR #1266, opened 2021-04-26
Add network disconnect step to goth Actions workflow

zakaprov:
This is an attempt to avoid the issue described here: https://github.com/moby/moby/issues/19261

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -126,6 +126,13 @@ jobs:
       - name: Install dependencies
         run: poetry install --no-root
 
+      - name: Disconnect Docker containers from default network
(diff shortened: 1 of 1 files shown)


azawlocki (approved):
This solution won't work in all cases of https://github.com/moby/moby/issues/19261, it relies on the containers attached to the network being listed in the `Containers` section of `docker network inspect`. Sometimes the `Containers` list is empty but `docker network rm` thinks some entrypoints are attached to the network and refuses to remove it.

It happened in a CI run for this very PR: https://github.com/golemfactory/yagna/runs/2438464930#step:6:1

Still, it's better than nothing.

PR #1278, opened 2021-04-27
Fix scope for config overrides in goth tests

zakaprov:
Resolves: https://github.com/golemfactory/goth/issues/498

### DIFF
--- a/goth_tests/conftest.py
+++ b/goth_tests/conftest.py
@@ -37,7 +37,7 @@ def log_dir() -> Path:
     return log_dir
 
 
-@pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


azawlocki (dismissed):
Looks good (although making `config_overrides` a fixture with a "function" scope would probably work as well and require only changing in a single line of code).

zakaprov:
@azawlocki Agreed, I'm going to update this PR once `goth` version `0.2.4` is released (so that integration tests here use a stable release version). The relevant PR in `goth` is here: https://github.com/golemfactory/goth/pull/500

azawlocki (approved):
LGTM!

PR #1304, opened 2021-05-07
[mkt] Restore 1h as default Offer/Demand TTL

tworec:
Why:
- we see Offers @ `mainnet` with TTL of 50s which are expiring during negotiation

What:
- revert unintentionally (accidentally?) shortened config value (vide #1002)
- remove double default values definition (impl Default removed, StructOpt kept)

### DIFF
--- a/core/market/src/config.rs
+++ b/core/market/src/config.rs
@@ -50,7 +50,7 @@ impl Config {
 impl Default for SubscriptionConfig {
     fn default() -> Self {
         SubscriptionConfig {
-            default_ttl: chrono::Duration::seconds(50),
+            default_ttl: chrono::Duration::hours(1),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
LGTM!

core/market/src/config.rs
```diff
@@ -33,37 +33,40 @@ pub struct SubscriptionConfig {
...
-        // Mock command line arguments, because we want to use ENV fallback
-        // or default values if ENV variables don't exist.
-        Ok(Config::from_iter_safe(vec!["yagna"].iter())?)
+        // Empty command line arguments, because we want to use ENV fallback
+        // or default values if ENV variables are not set.
+        Ok(Config::from_iter_safe(&[""])?)
     }
 }
 
-impl Default for SubscriptionConfig {
```
nieznanysprawiciel, line 50: 👍🏻

core/market/src/config.rs
```diff
@@ -33,37 +33,40 @@ pub struct SubscriptionConfig {
...
 }
 
 impl Config {
     pub fn from_env() -> Result<Config, structopt::clap::Error> {
-        // Mock command line arguments, because we want to use ENV fallback
-        // or default values if ENV variables don't exist.
-        Ok(Config::from_iter_safe(vec!["yagna"].iter())?)
+        // Empty command line arguments, because we want to use ENV fallback
+        // or default values if ENV variables are not set.
+        Ok(Config::from_iter_safe(&[""])?)
```
nieznanysprawiciel, line 48: Does it work? First argument is expected to be binary name like in regular command line args

nieznanysprawiciel (approved):

core/market/src/config.rs
tworec: yes, I've added test

Issue #1129, opened 2021-03-09
Restart negotiations on Provider after Agreement was terminated


nieznanysprawiciel:
Why:
Provider approves one Agreement and rejects all others. After Agreement is finished he can't negotiate with Requestors, that he rejected. Current workaround is to resubscribe Offer after each Agreement, but this solution is far from optimal.

What:
- [x] Collect Proposals, that were rejected, but rejection wasn't final
- [ ] Teach negotiators about `golem.proposal.rejection.is-final`
- [x] Show collected Proposals to Negotiators after Agreement termination for new decision
- [x] Remove code resubscribing Offers after each Agreement termination
- [ ] Enable respawning negotiations in market

Should follow specification: https://github.com/golemfactory/golem-internal/wiki/Yagna-Capability-List

Needs #1150

PR #1306, opened 2021-05-07
Restart negotiations

jiivan:
fixes: #1129

### DIFF
--- a/agent/provider/src/market/negotiator/builtin/expiration.rs
+++ b/agent/provider/src/market/negotiator/builtin/expiration.rs
@@ -2,9 +2,9 @@ use anyhow::{anyhow, Result};
 use chrono::{DateTime, Duration, TimeZone, Utc};
 
 use ya_agreement_utils::{Error, OfferDefinition};
-use ya_client_model::market::Reason;
 
 use crate::display::EnableDisplay;
+use crate::market::negotiator::common::reason_with_extra;
 use crate::market::negotiator::factory::AgreementExpirationNegotiatorConfig;
 use crate::market::negotiator::{
     AgreementResult, NegotiationResult, NegotiatorComponent, ProposalView,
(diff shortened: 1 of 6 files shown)


agent/provider/src/market/negotiator/builtin/expiration.rs
```diff
@@ -136,10 +145,10 @@ impl NegotiatorComponent for LimitExpiration {
                     // Below certain timeout it is impossible for Requestor to accept DebitNotes.
                     if req_deadline.num_seconds() < self.min_deadline {
                         return Ok(NegotiationResult::Reject {
-                            reason: Some(Reason::new(format!(
-                                "To low DebitNotes timeout: {}",
-                                req_deadline.display()
-                            ))),
+                            reason: Some(reason_with_extra(
+                                format!("To low DebitNotes timeout: {}", req_deadline.display()),
+                                serde_json::json!({"golem.proposal.rejection.is-final": true}),
```
nieznanysprawiciel, line 150: It would be much better, if `NegotiationResult` would have `isFinal` field. It would be responsibility of external layer to add this entry to Reason struct.
Note that this is field has special meaning for negotiation framework

agent/provider/src/market/negotiator/builtin/expiration.rs
```diff
@@ -102,18 +102,24 @@ impl NegotiatorComponent for LimitExpiration {
...
-                    max_expiration_delta.display()
-                ))),
+                reason: Some(reason_with_extra(
+                    format!(
+                        "Proposal expires at: {} which is less than {} or more than {} from now",
+                        req_expiration,
+                        self.min_expiration.display(),
+                        max_expiration_delta.display()
+                    ),
+                    serde_json::json!({ "golem.proposal.rejection.is-final": too_late }), // when it's too soon we could try later
```
nieznanysprawiciel, line 121: Is this comment true?
If expiration is to close to `now`, Provider will never pickup this Proposal

agent/provider/src/market/provider_market.rs
```diff
@@ -736,11 +788,12 @@ impl Handler<AgreementFinalized> for ProviderMarket {
...
             ctx.spawn(terminate_agreement(myself.api.clone(), msg).into_actor(myself));
 
-            log::info!("Re-subscribing all active offers to get fresh proposals from the Market");
+            log::info!("Re-negotiating all demands");
 
-            let subscriptions = std::mem::replace(&mut myself.subscriptions, HashMap::new());
+            let demands = std::mem::replace(&mut myself.postponed_demands, Vec::new());
             ctx.spawn(
-                resubscribe_offers(ctx.address(), myself.api.clone(), subscriptions)
+                //renegotiate_demands(self.async_context(ctx), myself.api.clone(), demands).into_actor(myself),
```
nieznanysprawiciel, line 795: Leftovers?

agent/provider/src/market/provider_market.rs
```diff
@@ -136,6 +142,7 @@ impl ProviderMarket {
             negotiator: Arc::new(NegotiatorAddr::default()),
             config: Arc::new(config),
             subscriptions: HashMap::new(),
+            postponed_demands: Vec::new(),
```
nieznanysprawiciel, line 145: I would store this in `CompositeNegotiator`, but as I see there are reasons to have this here? Some would be lacking in negotiator?

agent/provider/src/market/negotiator/builtin/expiration.rs
jiivan: If demand came too_soon, then in the future it might be inside the acceptance window

nieznanysprawiciel: But it doesn't matter when Demand came. Expiration timestamp will always be closer

jiivan: https://github.com/golemfactory/yagna/blob/9928a2f44fe69aabb3142c5139e423e03668a169/agent/provider/src/market/negotiator/builtin/expiration.rs#L100-L105

`too_soon` means that requestor expiration is bigger than max expiration. As far as I understand it will become smaller and smaller with passing time. Am I missing something?

nieznanysprawiciel (approved):

PR #1309, opened 2021-05-10
Logging guidelines: Timestamp format

stranger80:
Updated Logging Guidelines to specify timestamp format - Local time, millisecond precision, with offset from UTC, eg: 

2020-08-27T07:56:22.348+02:00


maaktweluit (changes requested):
LGTM!

Some small comments:
- Why is `yagna.db` committed?
- Why are there 22 commits? i can help clean up this branch and only keep the last if you want ( please ping on discord if you want to sync on this )

PR #1310, opened 2021-05-11
Added test checker on pull-request waiting for goth

maaktweluit:


.github/workflows/goth.yml
```diff
@@ -18,12 +31,15 @@ jobs:
...
     name: Integration Tests
     runs-on: goth
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    needs: wait-for-build
     defaults:
       run:
         working-directory: './goth_tests'
 
     steps:
+      - run: echo '${{ toJson(github) }}'
```
tworec, line 40: Pls make sure we do not reveal `github.token` by echoing whole `github` context.
See warning here:
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Wait for build to succeed ( ubuntu )
+        uses: tomchv/wait-my-workflow@v1.1.0
+        id: wait-for-build-ubu
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: "Tests & Build (ubuntu-latest)"
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          timeoutSeconds: 7200
+
+      - name: Make sure the build was a success ( ubuntu )
```
tworec, line 23: nitpick: this step is marked as `skipped` (gray icon 🕧) when target workflow has succeded, so we can consider changing this step name:

```suggestion
      - name: Fail if the build was not a success ( ubuntu )
```

But it would be even more readable when we can check the condition and mark step as success (🟢 ).

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Wait for build to succeed ( macos )
+        uses: tomchv/wait-my-workflow@v1.1.0
+        id: wait-for-build-mac
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: "Tests & Build (macos-latest)"
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          timeoutSeconds: 7200
+
+      - name: Make sure the build was a success ( macos )
```
tworec, line 49: nitpick: macos is usually faster than windows, so we can swap order

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+        id: wait-for-build-ubu
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: "Tests & Build (ubuntu-latest)"
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          timeoutSeconds: 7200
+
+      - name: Make sure the build was a success ( ubuntu )
+        if: steps.wait-for-build-ubu.outputs.conclusion != 'success'
+        run: echo buils not success && false
```
tworec, line 25:
```suggestion
        run: echo "Ubuntu build has failed" && false
```

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: "Tests & Build (windows-latest)"
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          timeoutSeconds: 7200
+
+      - name: Make sure the build was a success ( windows )
+        if: steps.wait-for-build-win.outputs.conclusion != 'success'
+
+        run: echo buils not success && false
```
tworec, line 39:
```suggestion
        run: echo "Windows build has failed" && false

```

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+        id: wait-for-build-mac
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: "Tests & Build (macos-latest)"
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          timeoutSeconds: 7200
+
+      - name: Make sure the build was a success ( macos )
+        if: steps.wait-for-build-mac.outputs.conclusion != 'success'
+        run: echo buils not success && false
```
tworec, line 51:
```suggestion
        run: echo "macOS build has failed" && false
```

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Sleep to wait for the build to trigger
+        run: sleep 1m
+
+      - name: Wait for test to succeed
+        uses: tomchv/wait-my-workflow@v1.1.0
+        id: wait-for-test
+        if: steps.wait-for-build.outputs.conclusion == 'success'
```
tworec, line 59: this conditional marks this step skipped (gray 🕧 ) when it should be failed (🔴 )

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          checkName: "Integration Tests"
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          timeoutSeconds: 1200
+
+      - run: echo '${{ toJson(steps.wait-for-test) }}'
```
tworec, line 66: This is probably some leftover or development stage of this workflow. Sorry if reviewing too early.

tworec (changes requested):

.github/workflows/test-check.yml
zakaprov: Somewhat related: @maaktweluit is the order here relevant? In other words: if a build is finished before we get to this step will the result be reported correctly?

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Fail if test did not launch
+        if: steps.wait-for-test.outputs.conclusion == 'does not exist'
+        run: echo job does not exist && false
```
zakaprov, line 70:
```suggestion
        run: echo "Integration tests job does not exist" && false
```

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Fail if test failed
+        if: steps.wait-for-test.outputs.conclusion == 'failure'
+        run: echo fail && false
```
zakaprov, line 74:
```suggestion
        run: echo "Integration tests job failed" && false
```

.github/workflows/test-check.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Fail if test timed out
+        if: steps.wait-for-test.outputs.conclusion == 'timed_out'
+        run: echo Timeout && false
```
zakaprov, line 78:
```suggestion
        run: echo "Timed out waiting for integration tests to finish" && false
```

.github/workflows/goth.yml
maaktweluit: for sure! is already removed

this was just for debugging.. will keep it in mind next time

zakaprov:
Also: have you looked at having the test checker print a link to the `goth` tests job it's waiting for? It's a tiny detail, but it would be very convenient!

.github/workflows/test-check.yml
maaktweluit: Good one! fixed it :)

.github/workflows/test-check.yml
maaktweluit: 1. fixed
2. the order does not matter at all, if the build already succeeded ( for example on rebuild ) it returns green right away.

.github/workflows/test-check.yml
maaktweluit: this is outdated, the name of the OS is in the title and the status is now always printed.

.github/workflows/test-check.yml
maaktweluit: idem

.github/workflows/test-check.yml
maaktweluit: idem

.github/workflows/test-check.yml
maaktweluit: good catch! this no longer applies...

.github/workflows/test-check.yml
maaktweluit: you are correct, its removed now!

No worries, i'm very glad with the review comments!

.github/workflows/test-check.yml
maaktweluit: idem

.github/workflows/test-check.yml
maaktweluit: idem

maaktweluit:
> Also: have you looked at having the test checker print a link to the `goth` tests job it's waiting for? It's a tiny detail, but it would be very convenient!

Great idea! was one of the things i wanted to add in cleanup..

Unfortunately we had to pick another route: Waiting inside goth.yaml.
The benefit is that the PR is now neatly linked from the start.

maaktweluit:
Replied / resolved all comments :)

@zakaprov @tworec please review again

.github/workflows/goth.yml
```diff
@@ -1,24 +1,59 @@
...
+  pull_request:
+    branches:
+      - master
+      - release/*
 
 jobs:
-  triggering-workflow-info:
-    name: Triggerin Workflow Info
+  test_check:
+    name: Check test results
```
tworec, line 14:
```suggestion
    name: Check Rust CI test results
```

tworec (approved):
lgtm

PR #1312, opened 2021-05-11
Market/config upgrade

tworec:
Why:
- sync config changes from `master` to `release/v0.6`
- enable users to configure cleanup intervals (will help @cryptobench)

What:
- cherry-pick commit from #1304 which was targeting `master`
- make cleanup interval configurable
- move two other env vars to the config

Relates to #453 and #460

### DIFF
--- a/core/market/src/config.rs
+++ b/core/market/src/config.rs
@@ -5,9 +5,9 @@ use structopt::StructOpt;
 pub struct Config {
     #[structopt(flatten)]
     pub discovery: DiscoveryConfig,
-    #[structopt(skip)]
+    #[structopt(flatten)]
     pub subscription: SubscriptionConfig,
-    #[structopt(skip)]
+    #[structopt(flatten)]
     pub events: EventsConfig,
 }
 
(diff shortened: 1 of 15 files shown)


core/market/src/config.rs
```diff
@@ -33,37 +35,61 @@ pub struct SubscriptionConfig {
     pub default_ttl: chrono::Duration,
 }
 
-#[derive(Clone)]
+#[derive(StructOpt, Clone)]
 pub struct EventsConfig {
+    #[structopt(env = "MAX_MARKET_EVENTS_DEFAULT", default_value = "20")]
```
jiivan, line 40:
```suggestion
    #[structopt(env = "MARKET_MAX_EVENTS_DEFAULT", default_value = "20")]
```

core/market/src/config.rs
```diff
@@ -33,37 +35,61 @@ pub struct SubscriptionConfig {
...
     pub max_events_default: i32,
+    #[structopt(env = "MAX_MARKET_EVENTS_MAX", default_value = "100")]
```
jiivan, line 42:
```suggestion
    #[structopt(env = "MARKET_MAX_EVENTS_MAX", default_value = "100")]
```

core/market/src/config.rs
```diff
@@ -33,37 +35,61 @@ pub struct SubscriptionConfig {
...
     pub max_events_max: i32,
 }
 
+#[derive(StructOpt, Clone)]
+pub struct DbConfig {
+    /// Interval in which Market cleaner will be invoked
+    #[structopt(env = "MARKET_DB_CLEANUP_INTERVAL", parse(try_from_str = humantime::parse_duration), default_value = "24h")]
+    pub cleanup_interval: Duration,
+    /// Number of days to persist Agreements and related Agreement Events
+    #[structopt(env = "MARKET_AGREEMENT_STORE_DAYS", default_value = "90")]
```
jiivan, line 52: Why drop the `YAGNA_` prefix?

jiivan: Besides the default, there was also the `min` value (30). Will it be ignored now?

tworec:
I've cherrypicked #1292 to fix flaky tests.

core/market/src/config.rs
tworec: > Why drop the `YAGNA_` prefix?

To be consistent with the rest of env vars here.

> Besides the default, there was also the `min` value (30). Will it be ignored now?

I was struggling with this decision but finally decided to drop this feature.

jiivan (approved):

PR #1314, opened 2021-05-12
Runtime API service: refactor run_async's factory to FnOnce

mfranciszkiewicz:
Improves the runtime factory function by allowing to capture outer variables


tworec (approved):
lgtm

PR #1327, opened 2021-05-19
Bump ya-sb-router and actix to 0.10

### LINKED ISSUE #1325: bump actix to 0.10 (uses old tokio) within yagna
Why:
- service bus is already on actix 0.10
- we need to catch up with that
- we want to release it sooner than beta.2 to fix the memory problem and stabilize it as soon as possible in the wild 

What:
- [ ] bump dep within yagna only

mfranciszkiewicz:
Resolves #1325

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -172,6 +172,11 @@ ya-service-api-interfaces = { path = "core/serv-api/interfaces" }
 ya-service-api-web = { path = "core/serv-api/web" }
 
 ## SERVICE BUS
+ya-service-bus = { path = "../ya-service-bus" }
+ya-sb-proto = { path = "../ya-service-bus/crates/proto" }
+ya-sb-router = { path = "../ya-service-bus/crates/router" }
+ya-sb-util = { path = "../ya-service-bus/crates/util" }
+
 #ya-service-bus = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "b59a415ca00aecf83cf3df78fe032ee067344620"}
 #ya-sb-proto = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "b59a415ca00aecf83cf3df78fe032ee067344620"}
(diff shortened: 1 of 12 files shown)


exe-unit/runtime-api/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ya-runtime-api"
-version = "0.2.1"
+version = "0.3.0"
```
nieznanysprawiciel, line 3: As far as I know, we use `cargo release` to bump versions. You should probably use use patch in top level cargo.toml

Cargo.toml
```diff
@@ -172,10 +172,10 @@ ya-service-api-interfaces = { path = "core/serv-api/interfaces" }
 ya-service-api-web = { path = "core/serv-api/web" }
 
 ## SERVICE BUS
-#ya-service-bus = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "b59a415ca00aecf83cf3df78fe032ee067344620"}
-#ya-sb-proto = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "b59a415ca00aecf83cf3df78fe032ee067344620"}
-#ya-sb-router = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "b59a415ca00aecf83cf3df78fe032ee067344620"}
-#ya-sb-util = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "b59a415ca00aecf83cf3df78fe032ee067344620"}
+ya-service-bus = { git = "https://github.com/golemfactory/ya-service-bus.git", rev = "fe78d2f49986ff45152555866cca46d212d437f6"}
```
tworec, line 175: we need to release ya-service-bus to crates.io before Patch.7 release

tworec: We need to decrease verbosity of the new GSB

tworec (approved):

PR #1338, opened 2021-05-21
Changed date format.

### LINKED ISSUE #1301: Improve log format
Why:
- to have unified logs across the project

What:
- [x] add milliseconds 
- [x] add `--debug` flag
https://www.notion.so/golemnetwork/ACTION-Unify-all-the-logs-8a55189679e74f69870f1261a3c6f39a
- [x] change time format according to
https://www.notion.so/golemnetwork/ACTION-unify-the-logs-between-yagna-and-yaXapis-time-format-62ba173d2fd6419981d836ab50ae90d8

scx1332:
Resolves #1301


agent/provider/src/provider_agent.rs
```diff
@@ -172,14 +172,33 @@ impl GlobalsState {
...
+        //if --debug option is provided override RUST_LOG flag with debug defaults
+        //if you want to more detailed control over logs use RUST_LOG variable and do not use --debug flag
+        if args.debug {
+            std::env::set_var(
+                "RUST_LOG",
+                "debug,tokio_core=info,tokio_reactor=info,hyper=info,reqwest=info",
+            );
+        }
+        //start_logger is using env var RUST_LOG internally
+        let log_handler = start_logger("info", Some(&log_dir), &vec![])?;
```
nieznanysprawiciel, line 192: You shouldn't use RUST_LOG variable. Check what `start_logger` function is able to do.
In first parameter you can pass log level. Last parameter is list of modules to filter

core/serv/src/main.rs
```diff
@@ -332,7 +336,17 @@ impl ServiceCommand {
...
             }) => {
+                //if --debug option is provided override RUST_LOG flag with debug defaults
+                //if you want to more detailed control over logs use RUST_LOG variable and do not use --debug flag
+                if *debug {
+                    env::set_var(
+                        "RUST_LOG",
+                        "debug,tokio_core=info,tokio_reactor=info,hyper=info,reqwest=info",
+                    );
+                }
+
```
nieznanysprawiciel, line 349: The same here. Below you have example how to filter out spamming modules

nieznanysprawiciel (changes requested):

core/serv/src/main.rs
```diff
@@ -332,7 +336,17 @@ impl ServiceCommand {
...
                 max_rest_timeout,
                 log_dir,
+                debug,
             }) => {
+                //if --debug option is provided override RUST_LOG flag with debug defaults
+                //if you want to more detailed control over logs use RUST_LOG variable and do not use --debug flag
+                if *debug {
+                    env::set_var(
+                        "RUST_LOG",
+                        "debug,tokio_core=info,tokio_reactor=info,hyper=info,reqwest=info",
```
maaktweluit, line 346: `trust_dns_resolver` and `trust_dns_proto` can still be put on info as well

agent/provider/src/provider_agent.rs
scx1332: start_logger is using RUST_LOG internally and it is ignoring everything else if it set. If someone uses --debug flag I think we should override it.

nieznanysprawiciel (dismissed):

scx1332 (commented):
Looks fine to me. I don't see anything weird.
There is change in Windows runner that was not there originally, but I think it is fine. Still there is Goth issue (maybe related to output format change)

tworec (approved):
lgtm

PR #1340, opened 2021-05-21
ExeUnit cache: cleaner cache/tmp, less intermediary files

### LINKED ISSUE #1319: multiple copies of the gvmi image in their exe-unit/cache/tmp directory
A provider machine can run out of disk space from having multiple copies of a gvmi image in their exe-unit/cache/tmp directory.
`(/root/.local/share/ya-provider/exe-unit/cache/tmp)` — potentially the solution could be to clear exe-unit/cache/tmp directory before starting a new task.

Relevant to Bug Bounty report #2: https://docs.google.com/document/d/1-ZyOOZ_f8g4qfLmcw0G1kEBxQCzALcZvTDc_6VUQSs4/edit?usp=sharing

mfranciszkiewicz:
- hash is now included in target image name
- skip an intermediary cache file
- fix: remove temporary files on transfer failure

Resolves #1319

### DIFF
--- a/exe-unit/src/service/transfer.rs
+++ b/exe-unit/src/service/transfer.rs
@@ -276,7 +276,6 @@ impl Handler<DeployImage> for TransferService {
         {
             let from_provider = actor_try!(self.provider(&source_url));
             let to_provider: FileTransferProvider = Default::default();
-            let cache_path = self.cache.to_cache_path(&cache_name);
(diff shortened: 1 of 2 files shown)


maaktweluit (approved):
LGTM!

Not the best at this part of the code yet, but as far as i can see its a neat improvement!

PR #1343, opened 2021-05-21
Major dependency upgrade: actix 0.11, tokio 1.6

mfranciszkiewicz:
- non-core actix crates are in beta and version-pinned (due cross-crate compatibility issues)
- zksync crate is supported via `tokio-compat-02`
- web3 is pinned at a git repository revision
- ethereum-tx-sign doesn't seem to be actively maintained and was forked to upgrade its dependencies
- serial_test forked with actix-rt 2.2 compat


tworec:
pls retarget this to `master`

exe-unit/src/network/inet.rs
```diff
@@ -516,7 +514,6 @@ async fn inet_tcp_proxy<'a>(ip: IpAddr, port: u16) -> Result<(TransportSender, T
     log::debug!("[inet] connecting TCP to {}:{}", ip, port);
 
     let tcp_stream = TcpStream::connect((ip, port)).await?;
-    let _ = tcp_stream.set_keepalive(Some(TCP_KEEP_ALIVE));
```
nieznanysprawiciel, line 519: @mfranciszkiewicz I had to remove this line, because new tokio::TcpStream doesn't have equivalent. is this dangerous?

nieznanysprawiciel:
- serial_test upgraded with feature flag to enable new `actix-rt`
- removed problem with `parity-crypto` library, when doing `cargo update`

exe-unit/src/network/inet.rs
mfranciszkiewicz: Connections may die due to lack of activity. We can use a workaround similar to the one implemented in hyper
https://github.com/hyperium/hyper/blob/055b4e7ea6bd22859c20d60776b0c8f20d27498e/src/client/connect/http.rs#L588-L673

wkargul (approved):

wkargul (approved):

Issue #1087, opened 2021-02-25
Failed to save Offer - expired. investigate


nieznanysprawiciel:
```
[2021-05-24T17:13:31Z WARN  ya_market::matcher::handlers] Failed to save Offer [618609ab0f074041adcd0ca383c471a1-a1c634796e24bf525baa12a69ff5f630d5d7c183b16e3f30501f2ceb6601941d]. Error: Offer [618609ab0f074041adcd0ca383c471a1-a1c634796e24bf525baa12a69ff5f630d5d7c183b16e3f30501f2ceb6601941d] expired.
```

What:
- [x] Decrease log level (done via #1367)
- [ ] Investigate why Nodes are sending expired Offers

scx1332:
Log level decreased. Investigate part is left open.

maaktweluit:
I have looked into the possible reasons for sending a expired offer and found this:
https://github.com/golemfactory/yagna/blob/master/core/market/src/db/dao/offer.rs#L74

I think it should not filter the unsubscribes by timestamp, all un-subscribes should be removed from the list.

maaktweluit:
I have finished my investigation and was not able to find the root cause of this issue.

Re-assigned to 2rec with more market knowledge so we can check it further.

tworec:
Upon starting my node (requestor, occasionally connected) I've got
```
[2021-06-12T01:07:55.314+0200 INFO  actix_server::builder] Starting "actix-web-service-127.0.0.1:7465" service on 127.0.0.1:7465
[2021-06-12T01:08:24.230+0200 INFO  ya_market::matcher::handlers] Failed to save Offer [5eb3954...].  Error: Offer [5eb395...] expired.
``` 
But on the second node (provider, which is running all the time) this offer was saved (insertion ts witnesses it was just 6s after created on the originating node).
```
select * from market_offer where id like '5eb395%';
           id = 5eb3954fb6694cd5a9a56333ea038723-7f16a68906a257f9d0dfb8112b0fc5659129d6e958a127eb62e2bb2d43cdeb96
[...]

PR #1367, opened 2021-05-27
Lowered warn to inf on failed to save offer

scx1332:
Please note that this is hiding deeper issue, that these offers shouldn't be sent. 

https://github.com/golemfactory/yagna/issues/1087


tworec:
I've cherrypicked it also to `release/v0.7` via 79d9d27f

Issue #1364, opened 2021-05-26
galatea task hangs on second run command call


nieznanysprawiciel:
Repos:
https://github.com/golemfactory/galatea/blob/pnowosie/pass-requests-to-classificators-APPS-83/provider/test/yagna_batch_test.py

https://github.com/golemfactory/galatea/blob/pnowosie/pass-requests-to-classificators-APPS-83/provider/classifier/classifier.py

Image hash: c6b743459d3428fb860582e556ceba1c76dbc8a1d599a55dcf73e437

Wiezzel:
This error occurs with provider in version `v0.6.7-beta.1` but does not occur when provider is in version `v0.6.5`.

mfranciszkiewicz:
https://github.com/golemfactory/ya-runtime-sdk/pull/11 should fix this issue

Wiezzel:
> [golemfactory/ya-runtime-sdk#11](https://github.com/golemfactory/ya-runtime-sdk/pull/11) should fix this issue

:crossed_fingers:

mfranciszkiewicz:
> ~[golemfactory/ya-runtime-sdk#11](https://github.com/golemfactory/ya-runtime-sdk/pull/11) should fix this issue~

This issue most probably needs to be fixed on the Supervisor level

nieznanysprawiciel:
The problem is first run command, that runs in the background. Execution finishes immediately after spawning process, so ExeUnit stops tracking this process, but output is still generated and Supervisor doesn't consume output events from the queue. The second spawned process finishes correctly, but event about this finish is stacked in the queue and never reaches Supervisor.

I'm working on fix on Supervisor level

Wiezzel:
@nieznanysprawiciel Let me know if you find some way to handle this in supervisor.

mfranciszkiewicz:
@Wiezzel `galatea` test is passing successfully with ExeUnit built from https://github.com/golemfactory/yagna/pull/1384. Could you confirm that the issue is resolved?

mfranciszkiewicz:
Resolved by #1384

PR #1384, opened 2021-06-01
ExeUnit Supervisor: ProcessStatus handling tweaks (incl. fixes)

mfranciszkiewicz:
- handle output from the start command in server (blocking) mode
- drop events from pids we're not listening the status of
- change the way runtime binary args are built and handled (cleanup)


Wiezzel:
Confirmed that this indeed fixes #1364

nieznanysprawiciel (approved):

PR #1388, opened 2021-06-07
Market - supress regenerate proposal warning

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna-triage/issues/120


jiivan:
Maybe we should leave at least some trace logs on known errors?

nieznanysprawiciel:
> Maybe we should leave at least some trace logs on known errors?

I didn't leave logs, because these "known errors" aren't errors at all. In case of expired and unsubscribed Demands/Offers Porposal shouldn't be generated

jiivan (approved):
lgtm

PR #1392, opened 2021-06-08
[mkt] cease rebroadcast

tworec:
partly resolves https://github.com/golemfactory/golem-internal/issues/288 
resolves https://github.com/golemfactory/golem-internal/issues/296


mfranciszkiewicz (commented):
The commented code could become an optional feature but I'm fine with this solution.

Note for the future: if disabling re-broadcasting is not enough, we may have to look into cyclic broadcasts.

jiivan (approved):

PR #1401, opened 2021-06-10
Fix Mkt Bcast metrics

tworec:
Why:
- initialization of the metric `market.offers.broadcasts.size` as `counter` conflicts with its usage, which causes `400 Bad Request` from Prometheus and prevents metrics retention for such nodes.
- bcast size has potentially fastly changing value, so histogram is better than gauge
- in order to asses how many bcats are skipped network-wide, we need this to be reported as a metric

What:
- restore bcast size metric to use `value!` rather than `gauge!`(revert change from #1392)
- adding a missing metric to count skipped bcasts

Bonus:
- Fixes message for #1087

### DIFF
--- a/core/market/src/matcher.rs
+++ b/core/market/src/matcher.rs
@@ -87,6 +87,7 @@ impl Matcher {
         counter!("market.offers.incoming", 0);
         counter!("market.offers.broadcasts", 0);
         counter!("market.offers.broadcasts.size", 0);
+        counter!("market.offers.broadcasts.skip", 0);
         counter!("market.offers.broadcasts.net", 0);
         counter!("market.offers.broadcasts.net_errors", 0);
         counter!("market.offers.unsubscribes.incoming", 0);
--- a/core/market/src/matcher.rs
+++ b/core/market/src/matcher.rs
@@ -86,13 +86,11 @@ impl Matcher {
         // until first change to value will be made.
         counter!("market.offers.incoming", 0);
         counter!("market.offers.broadcasts", 0);
-        counter!("market.offers.broadcasts.size", 0);
         counter!("market.offers.broadcasts.skip", 0);
         counter!("market.offers.broadcasts.net", 0);
         counter!("market.offers.broadcasts.net_errors", 0);
         counter!("market.offers.unsubscribes.incoming", 0);
         counter!("market.offers.unsubscribes.broadcasts", 0);
-        counter!("market.offers.unsubscribes.broadcasts.size", 0);
         counter!("market.offers.unsubscribes.broadcasts.net", 0);
         counter!("market.offers.unsubscribes.broadcasts.net_errors", 0);
 
--- a/core/market/src/matcher/handlers.rs
+++ b/core/market/src/matcher/handlers.rs
@@ -36,7 +36,6 @@ pub(super) async fn receive_remote_offers(
     let added_offers_ids = futures::stream::iter(msg.offers.into_iter())
         .filter_map(|offer| {
             let resolver = resolver.clone();
-            let offer_id = offer.id.clone();
             async move {
                 resolver
                     .store
@@ -46,10 +45,7 @@ pub(super) async fn receive_remote_offers(
                         resolver.receive(&offer);
                         offer.id
                     })
-                    .map_err(|e| {
-                        log::info!("Failed to save Offer [{}]. Error: {}", &offer_id, &e);
-                        e
-                    })
+                    .map_err(|e| log::info!("Skipping foreign Offer: {}", e))
                     .ok()
             }
         })
--- a/core/market/src/protocol/discovery.rs
+++ b/core/market/src/protocol/discovery.rs
@@ -1,6 +1,6 @@
 //! Discovery protocol interface
 use actix_rt::Arbiter;
-use metrics::{counter, gauge, timing};
+use metrics::{counter, timing, value};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Instant;
@@ -112,7 +112,7 @@ impl Discovery {
         let bcast_msg = SendBroadcastMessage::new(OffersBcast { offer_ids });
 
         counter!("market.offers.broadcasts.net", 1);
-        gauge!("market.offers.broadcasts.size", size as i64);
+        value!("market.offers.broadcasts.size", size as u64);
 
         // TODO: We shouldn't use send_as. Put identity inside broadcasted message instead.
         if let Err(e) = bus::service(local_net::BUS_ID)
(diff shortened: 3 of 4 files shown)


jiivan (dismissed):

jiivan (approved):

core/metrics/src/pusher.rs
```diff
@@ -44,7 +43,8 @@ pub async fn push_forever(host_url: &str) {
         }
     };
 
-    let mut push_interval = time::interval(Duration::from_secs(60));
+    let start = Instant::now() + Duration::from_secs(5);
```
tworec, line 46: This is to wait until Services are initialized. 
Without this initial metrics push returns error, as metrics are empty and server does not like it.

core/market/src/matcher/handlers.rs
```diff
@@ -46,10 +45,7 @@ pub(super) async fn receive_remote_offers(
                         resolver.receive(&offer);
                         offer.id
                     })
-                    .map_err(|e| {
-                        log::info!("Failed to save Offer [{}]. Error: {}", &offer_id, &e);
-                        e
-                    })
+                    .map_err(|e| log::info!("Skipping foreign Offer: {}", e))
```
tworec, line 48: I'm only changing wording here, to change sentiment. This is not a failure, just a notification. This is completely all right to skip some offers.

core/market/src/matcher.rs
```diff
@@ -86,12 +86,11 @@ impl Matcher {
         // until first change to value will be made.
         counter!("market.offers.incoming", 0);
         counter!("market.offers.broadcasts", 0);
-        counter!("market.offers.broadcasts.size", 0);
+        counter!("market.offers.broadcasts.skip", 0);
```
tworec, line 89: removing `size` metrics here and adding a new one, to measure `skip`ed ones, which is crucial for core team to decide how to deal with bcasts in the future.

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
 pub async fn push(client: &Client, push_url: String) {
-    let current_metrics = crate::service::export_metrics().await;
+    let metrics = crate::service::export_metrics().await;
     let res = client
         .put(push_url.as_str())
-        .send_body(current_metrics)
+        .send_body(metrics.clone())
         .await;
-    log::trace!("Pushed current metrics {:#?}", res);
+    match res {
```
tworec, line 62: Here I'm improving what we are communicating to the users through the logs.
Until now, noting was communicated (only trace message, which was not displayed most of the times).
That's why we missed it stopped to work.

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
+        Ok(r) if r.status().is_success() => {
+            log::trace!("Metrics pushed: {}", r.status())
```
tworec, line 64: In case of success, just stay at `trace` level. No reason to bother users.

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
+        }
+        Ok(r) if r.status().is_server_error() => {
+            log::debug!("Metrics server error: {:#?}", r);
```
tworec, line 67: In case of server errror just use `debug`. It is our (Golem Factory) fault, so nothin that should bother users.

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
+            log::trace!("Url: {}\nMetrics:\n{}", push_url, metrics);
+        }
+        Ok(mut r) => {
+            let body = r.body().await.unwrap_or_default().to_vec();
+            let msg = String::from_utf8_lossy(&body);
+            log::warn!("Pushing metrics failed: `{}`: {:#?}", msg, r);
```
tworec, line 73: in case of other errors we want to `warn` users, especially developers while they are testing their changes, to not allow brokerage of the metrics reporting again.

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
+            log::debug!("Url: {}", push_url);
+            log::trace!("Metrics:\n{}", metrics);
+        }
+        Err(SendRequestError::Timeout) | Err(SendRequestError::Connect(ConnectError::Timeout)) => {
+            log::debug!("Pushing metrics timed out");
```
tworec, line 78: Timeout is frequent currently, so lets not bother users (`debug` is enough).

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
+            log::trace!("Url: {}\nMetrics:\n{}", push_url, metrics);
+        }
+        Err(e) => {
+            log::info!("Pushing metrics failed: {}", e);
```
tworec, line 82: In case of other errors I'm proposing just to `info`rm users.

core/market/src/matcher/handlers.rs
tworec: This is related to #1087

core/metrics/src/pusher.rs
mfranciszkiewicz: Can we skip pushing empty metrics instead?

tworec: yes, but it is more tricky, since string with metrics is not empty (have some comment).
For the sake of simplicity I've dealt with it like that (not to employ flaky heuristics)

core/market/src/protocol/discovery.rs
```diff
@@ -112,7 +112,7 @@ impl Discovery {
         let bcast_msg = SendBroadcastMessage::new(OffersBcast { offer_ids });
 
         counter!("market.offers.broadcasts.net", 1);
-        gauge!("market.offers.broadcasts.size", size as i64);
+        value!("market.offers.broadcasts.len", size as u64);
```
jiivan, line 115: :+1: `.len` is more understandable for me than `.size`

jiivan (approved):
I like the clarifications of log messages! :+1:

mfranciszkiewicz (approved):

mateuszsrebrny (approved):

ederenn (approved):

core/metrics/src/pusher.rs
```diff
@@ -54,12 +54,36 @@ pub async fn push_forever(host_url: &str) {
...
+        Ok(r) if r.status().is_server_error() => {
+            log::debug!("Metrics server error: {:#?}", r);
+            log::trace!("Url: {}\nMetrics:\n{}", push_url, metrics);
+        }
+        Ok(mut r) => {
+            let body = r.body().await.unwrap_or_default().to_vec();
+            let msg = String::from_utf8_lossy(&body);
+            log::warn!("Pushing metrics failed: `{}`: {:#?}", msg, r);
+            log::debug!("Url: {}", push_url);
+            log::trace!("Metrics:\n{}", metrics);
```
maaktweluit, line 75: newlines in logs are bad for readability

maaktweluit (approved):
LGTM!

Just not fond (understatement) of newlines in logs..

core/metrics/src/pusher.rs
tworec: yes, I know, but having all of those is even worse, Url is pretty long (I've tried it in one line) and metrics are multi-page which fits only `trace`.

core/market/src/protocol/discovery.rs
tworec: I had to change the name, because metrics server (Prometheus) has remembered `...broadcast.size` has `counter` type.

Issue #1400, opened 2021-06-09
Re-negotiation seems not working properly


tworec:
Renegotiation feature seems not working:

Scenario:
- 1 Provider : `golemsp run --subnet jun9 --payment-network rinkeby`
- 1 Requestor daemon `yagna service run`
- 2 Requestor agents demanding same subnet ` python blender.py --subnet-tag jun9`

First requestor is able to sign agreement and finish job + pay. Second is waiting until timeout (15 mins) and fails.

First requestor
```
[2021-06-09 16:58:48,704 INFO yapapi] Using log file `blender-yapapi-2021-06-09_16.58.48.log`; in case of errors look for additional information there
yapapi version: 0.6.0a0
Using subnet: jun9, payment driver: zksync, and network: rinkeby

[2021-06-09 16:58:51,870 INFO yapapi.summary] Received proposals from 1 provider so far
[2021-06-09 16:58:53,505 INFO yapapi.summary] Agreement proposed to provider 'friendly-winter'
[2021-06-09 16:58:53,901 INFO yapapi.summary] Agreement confirmed by provider 'friendly-winter'
[2021-06-09 16:58:55,573 INFO yapapi.summary] Task started on provider 'friendly-winter', task data: 0
Task computed: Task(id=1, data=0), result: output_0.png, time: 0:00:06.934393
```

Second requestor
```
python blender.py --subnet-tag jun9
[2021-06-09 16:58:49,379 INFO yapapi] Using log file `blender-yapapi-2021-06-09_16.58.49.log`; in case of errors look for additional information there
yapapi version: 0.6.0a0
Using subnet: jun9, payment driver: zksync, and network: rinkeby

[2021-06-09 16:59:14,096 WARNING yapapi.summary] 1 offer has been collected from the market, but no provider has responded for 20s. Make sure you're using the latest released versions of yagna and yapapi, and the correct subnet. Also make sure that the timeout for computing all tasks is within the 5 min to 30 min range.
[2021-06-09 16:59:34,102 WARNING yapapi.summary] 1 offer has been collected from the market, but no provider has responded for 40s. Make sure you're using the latest released versions of yagna and yapapi, and the correct subnet. Also make sure that the timeout for computing all tasks is within the 5 min to 30 min range.
```

Even though provider did logged that it is re-negotiating in the meantime
```
[...]

PR #1406, opened 2021-06-10
Provider - fix proposal renegotiations

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/1400

Please do not merge it before `Beta.2` release. It is targeted to `Beta.2.Patch.1`

### DIFF
--- a/agent/provider/src/market/negotiator/common.rs
+++ b/agent/provider/src/market/negotiator/common.rs
@@ -7,6 +7,7 @@ use ya_agreement_utils::{AgreementView, OfferDefinition};
 use ya_client::model::market::Reason;
 use ya_client_model::market::{NewOffer, Proposal};
 
+use crate::display::EnableDisplay;
 use crate::market::termination_reason::BreakReason;
 
 /// Response for requestor proposals.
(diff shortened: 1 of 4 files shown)


jiivan (approved):

agent/provider/src/market/negotiator/common.rs
```diff
@@ -18,12 +19,10 @@ pub enum ProposalResponse {
         offer: NewOffer,
     },
     AcceptProposal,
-    #[display(
-        fmt = "RejectProposal{}",
-        "reason.as_ref().map(|r| format!(\" (reason: {})\", r)).unwrap_or(\"\".into())"
-    )]
+    #[display(fmt = "RejectProposal (reason: {})", "reason.display()")]
```
jiivan, line 22: Some concise indication whether rejection is final would be beneficial:
`RejectProposal (F reason: None)

jiivan (approved):

agent/provider/src/market/negotiator/common.rs
nieznanysprawiciel: I only wanted to preserve previous version. We can do this in future improvements

Issue #1393, opened 2021-06-09
VPN


tworec:
Link PRs which are addressing VPN feature

mfranciszkiewicz:
All VPN branches are named `mf/vpn`

- yagna https://github.com/golemfactory/yagna/tree/mf/vpn
- https://github.com/golemfactory/ya-runtime-vm/pull/78
- https://github.com/golemfactory/ya-client/pull/98

Basic VPN support introduced for demo purposes:
- https://github.com/golemfactory/yapapi/pull/444
- https://github.com/golemfactory/ya-py-aioclient/pull/17

PR #1410, opened 2021-06-14
vpn

mfranciszkiewicz:
See https://github.com/golemfactory/yagna/issues/1393 for a dependency list


core/vpn/src/device.rs
```diff
@@ -0,0 +1,81 @@
...
+    queue: &'a mut VecDeque<Vec<u8>>,
+}
+
+impl<'a> phy::TxToken for TxToken<'a> {
+    fn consume<R, F>(self, _timestamp: time::Instant, len: usize, f: F) -> smoltcp::Result<R>
+    where
+        F: FnOnce(&mut [u8]) -> smoltcp::Result<R>,
+    {
+        let mut buffer = Vec::with_capacity(len);
+        buffer.resize(len, 0);
```
maaktweluit, line 74: this line seems to do nothing: `buffer` was just initialised with the same`len` as capacity.

I propose to:
- Remove the line
- Init the `buffer` => `with_capacity(0)`

maaktweluit (commented):
Nice work!

Will continue the review tomorrow morning :)

core/vpn/src/device.rs
mfranciszkiewicz: It fills the buffer of capacity `len` with zeroes. Before that, `buffer.len()` is equal to 0.

Playground
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7492e3da91e3adc5a7fcfc4e63c89c23

Wiezzel: How about this?
```rust
let mut buffer = vec![0; len];
```

maaktweluit: Ahhhhh, i was thinking about this option later in the evening yesterday.. TIL!

Thank you! Disregard my comment :)

I like @Wiezzel 's proposal here too, but its not as "broken" ( read sub-obtimal ) as i thought ;)

maaktweluit (approved):
LGTM!

nieznanysprawiciel (dismissed):

PR #1433, opened 2021-06-17
Use dynamic stable branch in goth nightly

zakaprov:
This adds a job which obtains the latest stable branch name and uses that name to generate a run matrix definition for the proper testing job.
Doing so removes the need for manually updating the workflow file whenever there's a new stable branch created in the repo.

While testing this I noticed there are two branches which both refer to the `0.7` release, these are: `release/v0.7` and `release/v0.7.0`. As far as I can tell the second one should be removed.


.github/workflows/nightly-goth.yml
```diff
@@ -12,11 +12,35 @@ on:
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      # finds branches with names matching 'release\/v[0-9](\.[0-9])+' (e.g. release/v0.7) and returns one with highest version
+      - name: Get latest stable branch
+        id: latest-stable
+        # sed removes leading whitespaces and '*' characters (git uses it to indicate the current branch)
+        run: |
+          branch=$(git branch -a | sed 's/remotes\/origin\///' | sed 's/^[ \t*]*//' | grep -E '^release\/v[0-9](\.[0-9])+$' | sort -Vr | head -1)
```
azawlocki, line 31: In case `yagna` gets `v0.7.10`, `v1.23.4` or `v20.0.0` release :)
```suggestion
          branch=$(git branch -a | sed 's/remotes\/origin\///' | sed 's/^[ \t*]*//' | grep -E '^release\/v[0-9]+(\.[0-9]+)+$' | sort -Vr | head -1)
```

azawlocki: BTW, you can also replace `sed 's/remotes\/origin\///'` with slightly simpler `sed 's:remotes/origin::'`.

azawlocki (changes requested):
See my suggestion on improving a regex for `grep`

azawlocki (approved):
LGTM!

Issue #1462, opened 2021-06-29
Lazy subscribing to broadcasts after first Demand


nieznanysprawiciel:
Why:
- Broadcasts are overwhelming network. This would reduce traffic since most of the nodes are Providers

What:
- Subscribe to broadcasts after first Demand was created
- Don't unsubscribe when all Demands disappear - it is complicated and won't give to much benefit

PR #1474, opened 2021-07-02
LazyBroadcastBinder (for requestor only)

jiivan:
fixes: #1462

### DIFF
--- a/core/market/src/protocol/discovery.rs
+++ b/core/market/src/protocol/discovery.rs
@@ -47,7 +47,7 @@ pub struct DiscoveryImpl {
 
     offer_queue: Mutex<Vec<SubscriptionId>>,
     unsub_queue: Mutex<Vec<SubscriptionId>>,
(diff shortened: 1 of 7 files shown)


core/market/src/protocol/discovery.rs
```diff
@@ -47,6 +47,7 @@ pub struct DiscoveryImpl {
 
     offer_queue: Mutex<Vec<SubscriptionId>>,
     unsub_queue: Mutex<Vec<SubscriptionId>>,
+    lazy_binder_queue: Mutex<Vec<String>>,
```
nieznanysprawiciel, line 50: This could be Option not Vec probably. Am I right?

jiivan: There is no limit on the times that `bind_gsb()` could be invoked, that's why I chose Vec.

nieznanysprawiciel (dismissed):

nieznanysprawiciel (approved):

PR #1477, opened 2021-07-05
ExeUnit batch results: if cmd idx is set, return output for that cmd only

mfranciszkiewicz:
Default behavior (include output of all commands) remains unchanged


maaktweluit (approved):
LGTM!

nieznanysprawiciel (approved):

PR #1478, opened 2021-07-06
PaymentProcessor: assert `orders` size before calling `unwrap`

mfranciszkiewicz:


core/payment/src/processor.rs
```diff
@@ -344,14 +344,17 @@ impl PaymentProcessor {
...
             return Err(OrderValidationError::new("order_ids is empty").into());
         }
         let orders = self
             .db_executor
             .as_dao::<OrderDao>()
             .get_many(msg.order_ids, driver.clone())
             .await?;
+        if orders.is_empty() {
+            return Err(OrderValidationError::new("orders not found in the database").into());
+        }
```
Wiezzel, line 357: I'd suggest moving this check into `validate_orders`. That's exactly what the function is for.

mfranciszkiewicz: Well, so far it validated only existing orders, so I kept the current flow. Moving it as you wish.

maaktweluit (approved):
LGTM!

Wiezzel (approved):

PR #1501, opened 2021-07-10
v0.7.3: ya-sb-router 0.4.5, ya-service-bus 0.4.10

mfranciszkiewicz:
Dependency update.

Fixes:
- missing `UnregisterRequest` handler in router
- disables forced disconnection on `unmatched call reply`

Which resolves the following issues:
- immediate disconnections after reconnecting due to `unmatched call reply`
- `gftp` server close command due to the missing handler

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7007,9 +7007,9 @@ dependencies = [
 
 [[package]]
 name = "ya-sb-router"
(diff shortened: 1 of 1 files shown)


maaktweluit (approved):
LGTM!

PR #1516, opened 2021-07-15
Update README.md

paulperegud:
Update example to reflect ProviderLogMixin being dropped here: https://github.com/golemfactory/goth/commit/8db82de23d044e9006cce99bca0accaabc079f04


jiivan:
@paulperegud thanks for your contribution!

Since this README is related to `yagna/goth_tests`, not the **goth** package itself, the actual fix would be to use `from goth_tests.helpers.probe import ProviderProbe` instead of `from goth.runner.probe import ProviderProbe`

PR #1520, opened 2021-07-16
Change goth version to golemfactory/goth:az/ya-aioclient-0.6

azawlocki:
The purpose of this PR is to make sure that updating from `ya-aioclient-0.5.1` to `ya-acioclient-0.6.2` in `goth` does not break `goth` tests in `yagna`.


azawlocki:
All checks have passed

PR #1537, opened 2021-07-22
Id/drop

### LINKED ISSUE #1096: command "yagna id drop" not working
yagna: 0.6.0

```
yagna id drop 0x...
Error: Bad request: No service registered under given address '/local/identity/DropId'.
```

tworec:
Resolves #1096 

~~This PR is not backward compatible in terms of yagna CLI (`yagna id show` to be exact).~~
It is also experimentally adding a test suite in `golemsp` crate which is invoking `yanga` binary from `target/debug` folder. This test suite is switched off by default and controlled via compilation feature, because those tests need to be invoked after `cargo build`. I've added a separate step in CI for it also.


core/identity/src/cli/identity.rs
```diff
@@ -248,30 +225,16 @@ impl IdentityCommand {
...
-                    Err(e) => return CommandOutput::object(Err::<(), _>(e)),
-                    Ok(None) => anyhow::bail!("identity not found"),
-                };
+                let id = node_or_alias.resolve().await?;
 
-                if id.is_default {
-                    anyhow::bail!("default identity")
-                }
                 CommandOutput::object(
                     bus::service(identity::BUS_ID)
```
tworec, line 235: TODO: add here a warning that if this identity was used as a default identity it might have related entries in service databases (eg. in `payments.db`).

core/identity/src/cli/identity.rs
```diff
@@ -248,30 +225,16 @@ impl IdentityCommand {
...
-                    .await
-                    .map_err(|e| anyhow::Error::msg(e))?;
-                let id = match id {
-                    Ok(Some(v)) => v,
-                    Err(e) => return CommandOutput::object(Err::<(), _>(e)),
-                    Ok(None) => anyhow::bail!("identity not found"),
-                };
+                let id = node_or_alias.resolve().await?;
 
-                if id.is_default {
```
tworec, line 267: We should also check if this `id` is still used as a network id. If yes, do not allow to delete it, or unregister.

golem_cli/src/command/yagna.rs
```diff
@@ -354,3 +347,56 @@ mod tracker {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
```
tworec, line 352: below tests will pass only after `cargo build` was invoked before and `yagna` is built in debug mode.

core/identity/src/service/identity.rs
```diff
@@ -392,6 +396,12 @@ impl IdentityService {
             let this = this.clone();
             async move { this.lock().await.update_identity(update).await }
         });
+        let this = me.clone();
+        let _ = bus::bind(model::BUS_ID, move |drop: model::Drop| {
+            let this = this.clone();
+            async move { this.lock().await.drop_identity(drop).await }
```
tworec, line 402: We can use the mechanism of Identity Events introduced for informing subscribed services about identity (Un)Locking.

mfranciszkiewicz:
Implementation is incomplete and outdated (targeting v0.7 codebase)

PR #1566, opened 2021-08-12
Retry transaction in case of invalid nonce

nieznanysprawiciel:


core/payment-driver/zksync/src/dao.rs
```diff
@@ -181,6 +181,21 @@ impl ZksyncDao {
...
     }
 
+    pub async fn retry_transaction(&self, tx_id: &str) {
+        if let Err(e) = self
+            .transaction()
+            .update_tx_status(tx_id.to_string(), TransactionStatus::Created.into())
+            .await
+        {
+            log::error!(
+                "Failed to update transaction state in `transaction` {:?} : {:?}",
```
tworec, line 191:
```suggestion
                "Failed to set status to Created for tx {:?} : {:?}",
```

core/payment-driver/zksync/src/driver.rs
```diff
@@ -471,14 +471,31 @@ impl PaymentDriverCron for ZksyncDriver {
...
-                        tx,
-                        err
-                    );
-                    self.dao.transaction_failed(&tx.tx_id).await;
-                    for order_id in order_ids.iter() {
-                        self.dao.payment_failed(order_id).await;
+                    // In case of invalid nonce error we can retry sending transaction.
+                    // Reset payment and transaction state to 'not sent', so cron job will pickup
+                    // transaction again.
+                    if err.contains("nonce") {
```
tworec, line 477: I'd say that the safest would be to check some error code, but not sure if there are error codes standardized.
To be sure, I'd recommend more strict check here
```suggestion
                    if err.contains("Nonce mismatch") {
```

core/payment-driver/zksync/src/driver.rs
```diff
@@ -471,14 +471,31 @@ impl PaymentDriverCron for ZksyncDriver {
...
+                        log::warn!(
+                            "ZkSync transaction verification failed. Nonce mismatch. Transaction \
+                            will be retied. tx_details={:?} error={}",
```
tworec, line 480:
```suggestion
                            "Scheduling retry for tx {:?} because of nonce mismatch. ZkSync error: {}",
```

core/payment-driver/zksync/src/dao.rs
```diff
@@ -196,6 +211,21 @@ impl ZksyncDao {
...
     }
 
+    pub async fn retry_payment(&self, order_id: &str) {
+        if let Err(e) = self
+            .payment()
+            .update_status(order_id.to_string(), PAYMENT_STATUS_NOT_YET)
+            .await
+        {
+            log::error!(
+                "Failed to retry transaction in `payment` {:?} : {:?}",
```
tworec, line 221:
```suggestion
                "Failed to set status of the `payment` {:?} to be retried : {:?}",
```

tworec (changes requested):
I think we should check zkSync nonce error msg more strictly.
+ minor err msgs tweaks proposed.

core/payment-driver/zksync/src/driver.rs
```diff
@@ -471,14 +471,31 @@ impl PaymentDriverCron for ZksyncDriver {
...
+                            tx,
+                            err
+                        );
+
+                        self.dao.retry_transaction(&tx.tx_id).await;
```
maaktweluit, line 485: There is no need to re-try the transaction, since the current transaction just failed.

Only the payments need the state updated to `_NOT_YET`.

maaktweluit (commented):
LGTM!

One small comment about tx-status, will approve after that is resolved :)

core/payment-driver/zksync/src/driver.rs
```diff
@@ -471,14 +471,29 @@ impl PaymentDriverCron for ZksyncDriver {
...
+                        for order_id in order_ids.iter() {
+                            self.dao.retry_payment(order_id).await;
+                        }
+                    } else {
+                        log::error!(
+                            "ZkSync transaction verification failed. tx_details={:?} error={}",
+                            tx,
+                            err
+                        );
+                        self.dao.transaction_failed(&tx.tx_id).await;
```
maaktweluit, line 493: this line should still be moved out of the if/else, now the transaction will remain in pending state on nonce failure

tworec (approved):
lgtm

tworec (changes requested):
Retarget to `release/v0.7`

tworec (dismissed):

tworec (approved):

mfranciszkiewicz (approved):

PR #1573, opened 2021-08-27
[CI] Next attempt to stabilise market test suite

tworec:


tworec:
Closing this, since it has not helped to fix market test suite

PR #1621, opened 2021-09-30
Wait for last debitnote to be send and received

maaktweluit:
Added some waiting for logs to avoid race conditions.

This way the "last debit note" is always the one after the exe-unit is finished.


nieznanysprawiciel (approved):

zakaprov (approved):
Looks good!

PR #1625, opened 2021-10-03
VPN - handle multiple calls to `add node`

mfranciszkiewicz:
Propagates the node information in the network despite that the IP address is already registered on requestor's side


maaktweluit (approved):
LGTM!

PR #1629, opened 2021-10-04
ExeUnit: fix run -> term -> PID response race condition

mfranciszkiewicz:
In case of processes that finish almost instantaneously, the process ID response might be received after the termination event. Process IDs were used for tracking process completion and thus process termination was never registered.

### DIFF
--- a/exe-unit/src/runtime/event.rs
+++ b/exe-unit/src/runtime/event.rs
@@ -3,11 +3,11 @@ use std::future::Future;
 use std::ops::Not;
 use std::pin::Pin;
 use std::sync::{Arc, Mutex};
-use std::task::{Context, Poll};
+use std::task::{Context, Poll, Waker};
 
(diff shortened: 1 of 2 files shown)


maaktweluit (approved):
LGTM!

So many `.unwrap()`'s in this code.. maybe we should add more error handling :P

mfranciszkiewicz:
@maaktweluit There are 2 kinds of unwraps here:
1. Mutex lock unwraps
    Will only fail if a Mutex is poisoned (i.e. the thread panics while the lock has been acquired). There is no benefit to handling these since the application will no longer be working properly. Additionally, mutexes are internal to each struct / enum variant and cannot be acquired outside of this module.
2. unwrap on the `done` channel
    `Handle` can only be constructed within this module and `Handle::new` is where the unwrap resides.

PR #1630, opened 2021-10-04
Add 30 seconds ping to net client

maaktweluit:


core/net/src/hybrid/service.rs
```diff
@@ -145,6 +146,16 @@ pub async fn start_network(default_id: NodeId, ids: Vec<NodeId>) -> anyhow::Resu
         }
     }));
 
+    // Keep server connection alive by pinging every 30 seconds.
+    let ping_client = client.clone();
+    tokio::task::spawn_local(async move {
+        let mut interval = time::interval(Duration::new(30, 0));
+        loop {
+            interval.tick().await;
+            ping_client.server_session().await.unwrap().ping().await.unwrap();
```
mfranciszkiewicz, line 155: Maybe something like:
```rust
if let Ok(session) = ping_client.server_session().await {
    if let Err(err) = session.ping().await {
        log::warn!("unable to ping the relay server: {}", err);
    }
}
```

PR #1650, opened 2021-10-15
Make erc20 the default driver. Enable polygon.

zakaprov:
Changes the default driver to erc20 & enables polygon networks in golemsp.
This is related to Polygon support changes in `goth`: https://github.com/golemfactory/goth/pull/553


shadeofblue:
@zakaprov according to the consensus within the team, we would like the default to be ERC20 / rinkeby (i.e. NOT Polygon / mumbai) ...

jiivan (commented):
Thanks for implementing this.

Have you checked whether golemsp's `get_payment_network()` and `run()` work? They have explicit references to ZKSYNC.

https://github.com/golemfactory/yagna/blob/8f2be73e15edc6ac945da9853b3a3a317036d7b5/golem_cli/src/status.rs#L215-L241

https://github.com/golemfactory/yagna/blob/8f2be73e15edc6ac945da9853b3a3a317036d7b5/golem_cli/src/service.rs#L111-L126

core/payment/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub mod migrations {
     struct _Dummy;
 }
 
-pub const DEFAULT_PAYMENT_PLATFORM: &str = "zksync-rinkeby-tglm"; // TODO: remove
+pub const DEFAULT_PAYMENT_PLATFORM: &str = "erc20-rinkeby-tglm"; // TODO: remove
```
shadeofblue, line 31: btw, why is there a "TODO: remove" ?

jiivan (dismissed):

Wiezzel (commented):
Shouldn't [this line](https://github.com/golemfactory/yagna/blob/master/core/model/src/payment.rs#L510) be updated ?

PR #1666, opened 2021-10-28
Choose net implementation (Hybrid/Central) using ENV variable

nieznanysprawiciel:
Use ENV vars:
`YA_NET_TYPE=hybrid`
`YA_NET_TYPE=central`

Central Net is used by default


maaktweluit (dismissed):
I though we would stick with the feature flag, now a lot of my work on goth is redundant.

core/net/src/central/mod.rs
```diff
@@ -1,11 +1,8 @@
 mod api;
-#[cfg(any(feature = "service", test))]
```
mfranciszkiewicz, line 2: Modules gated by the `service` feature are not needed  by `core::{activity, market, payment, vpn}` and `ya-transfer` (imported by ExeUnit)

core/net/src/hybrid/mod.rs
```diff
@@ -1,10 +1,7 @@
 mod api;
 mod codec;
-#[cfg(any(feature = "service", test))]
```
mfranciszkiewicz, line 3: Same as `central/mod.rs`

mfranciszkiewicz (approved):

PR #1669, opened 2021-10-29
Testing proposal timeout branch in goth

maaktweluit:
Extend negotiation ( wait for proposal ) timeout to 30 seconds because braodcasts can collide and get ignored.

This ensures 8-10 broadcast can go through compared to the 1-2 with the lower timeout

### DIFF
--- a/goth_tests/pyproject.toml
+++ b/goth_tests/pyproject.toml
@@ -23,7 +23,7 @@ pytest-asyncio = "^0.14"
 # goth = "^0.7"
 # to use development goth version uncomment below
 # goth = { git = "https://github.com/golemfactory/goth.git", branch = "fix-flaky-tests-wait_for_proposals" }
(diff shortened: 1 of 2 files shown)


goth_tests/pyproject.toml
```diff
@@ -23,7 +23,7 @@ pytest-asyncio = "^0.14"
 # goth = "^0.7"
 # to use development goth version uncomment below
 # goth = { git = "https://github.com/golemfactory/goth.git", branch = "fix-flaky-tests-wait_for_proposals" }
-goth = { git = "https://github.com/golemfactory/goth.git", rev = "b5b4ad3c4156899781888f12c9e614ad7f734af9" }
+goth = { git = "https://github.com/golemfactory/goth.git", branch = "master" }
```
nieznanysprawiciel, line 26: I think we should reference specific commit on master

maaktweluit: the lock file already references a specific commit, but can update this :)

nieznanysprawiciel (approved):

PR #1712, opened 2021-11-23
ExeUnit: pass extra arguments down to runtime

### LINKED ISSUE #1702: ExeUnit: pass extra flags from runtime descriptor to binary
**What:**

Users can add custom flags / arguments in a runtime descriptor, which will be passed down from the Provider Agent, through ExeUnit, to the Runtime binary.

**Why:**

A single Runtime binary can support multiple operating modes that could be selected via a custom flag or argument.

Currently, the Provider passes down all extra flags and arguments set in a descriptor, but that feature is not implemented in ExeUnit.

**Use case:**

Erigon runtime. Currently the expected behaviour (multiple operating modes) cannot be implemented properly.

mfranciszkiewicz:
Introduces a `--runtime-arg` / `-r` argument to ExeUnit CLI for passing extra arguments to executed runtimes. The arguments are specified in a runtime descriptor.

Example runtime descriptor:

```json
[
  {
    "name": "custom-runtime",
    "version": "0.1.0",
    "supervisor-path": "exe-unit",
    "runtime-path": "custom-runtime/custom-runtime",
    "description": "My runtime",
    "extra-args": [
      "--runtime-managed-image",
      "--runtime-arg", "first",
      "--runtime-arg", "second"
    ]
  }
]
```

Example resulting invocation:

`custom-runtime <common_args> first second start <start_args>`

Example runtime implementation: https://github.com/golemfactory/ya-runtime-sdk/blob/d2bf17f9f96c7521df769f8a69b4b02c08c78de6/examples/environment/src/main.rs

Resolves #1702

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6826,7 +6826,7 @@ dependencies = [
 
 [[package]]
 name = "ya-exe-unit"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "actix",
  "actix-files",
--- a/exe-unit/Cargo.toml
+++ b/exe-unit/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ya-exe-unit"
-version = "0.1.4"
+version = "0.1.5"
 authors = ["Golem Factory <contact@golem.network>"]
 edition = "2018"
 
--- a/exe-unit/examples/transfer.rs
+++ b/exe-unit/examples/transfer.rs
@@ -214,6 +214,7 @@ async fn main() -> anyhow::Result<()> {
         agreement,
         work_dir: work_dir.clone(),
         cache_dir,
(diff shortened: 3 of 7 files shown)


nieznanysprawiciel (approved):
Everything is fine. I'm just wondering, if it wasn't more convenient, to have additional key (`runtime-args` for example) instead of using `extra-args` and being forced to type `--runtime-arg` for each argument

mfranciszkiewicz:
> Everything is fine. I'm just wondering, if it wasn't more convenient, to have additional key (`runtime-args` for example) instead of using `extra-args` and being forced to type `--runtime-arg` for each argument

I was thinking of that too, but since the expected use-cases would require a single extra argument, I limited the scope to `ya-exe-unit`. The provider can be extended at any time, if needed

PR #1723, opened 2021-11-28
improve json output for CommandOutput::Table

### LINKED ISSUE #906: better json output from commands
Current structure example:
`yagna --json id list`
```json
{
  "headers": [
    "default",
    "locked",
    "alias",
    "address"
  ],
  "values": [
    [
      "X",
      "",
      null,
      "0x9657978563b470cecef3822bf7a103cdf0aed1e6"
    ]
  ]
}
```

Better format proposal:
```
[
  {
    "default": "X",
    "locked": "",
    "alias": null,
    "address": "0x9657978563b470cecef3822bf7a103cdf0aed1e6"
  }
]
```

Dependencies:
- GOTH
- checkers

How this apply to compatibility policy??

biryukovmaxim:
{"headers": [...], "values": [[...]]} -> [{"header": "value"}]
resolve #906

### DIFF
--- a/core/serv-api/src/lib.rs
+++ b/core/serv-api/src/lib.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use prettytable::{color, format, format::TableFormat, Attr, Cell, Row, Table};
 use serde::Serialize;
+use std::collections::HashMap;
 use std::path::PathBuf;
 
 #[derive(Clone, Debug, Default)]
@@ -55,14 +56,40 @@ impl CommandOutput {
                 header,
             } => {
                 if json_output {
-                    println!(
-                        "{}",
-                        serde_json::to_string_pretty(&serde_json::json!({
-                            "headers": columns,
-                            "values": values
-                        }))
-                        .unwrap()
-                    )
+                    let columns_size_eq_values_size = values.iter().all(|row| match row {
+                        serde_json::Value::Array(values) => values.len() == columns.len(),
+                        _ => false,
+                    });
+                    if columns_size_eq_values_size {
+                        let kvs: Vec<HashMap<&String, &serde_json::Value>> = values
+                            .iter()
+                            .map(|row| match row {
+                                serde_json::Value::Array(row_values)
+                                    if columns.len() == row_values.len() =>
+                                {
+                                    columns
+                                        .into_iter()
+                                        .enumerate()
+                                        .map(|(idx, key)| (key, &row_values[idx]))
+                                        .collect()
+                                }
+                                _ => unreachable!(),
+                            })
+                            .collect();
+                        println!(
+                            "{}",
+                            serde_json::to_string_pretty(&serde_json::json!(kvs)).unwrap()
+                        )
+                    } else {
+                        println!(
+                            "{}",
+                            serde_json::to_string_pretty(&serde_json::json!({
+                                "headers": columns,
+                                "values": values
+                            }))
+                            .unwrap()
+                        )
+                    }
                 } else {
                     if let Some(txt) = header {
                         println!("{}", txt);


core/serv-api/src/lib.rs
```diff
@@ -55,14 +56,49 @@ impl CommandOutput {
...
-                            "headers": columns,
-                            "values": values
-                        }))
-                        .unwrap()
-                    )
+                    let columns_size_eq_values_size = values
+                        .iter()
+                        .map(|high_value| {
+                            if let serde_json::Value::Array(values) = high_value {
+                                values.len()
```
filipgolem, line 63:
```suggestion
                        .map(|row| {
                            if let serde_json::Value::Array(array) = row {
                                array.len()
```

core/serv-api/src/lib.rs
```diff
@@ -55,14 +56,49 @@ impl CommandOutput {
...
+                        let kvs: Vec<HashMap<&String, &serde_json::Value>> = values
+                            .iter()
+                            .map(|values| match values {
+                                serde_json::Value::Array(values)
+                                    if columns.len() == values.len() =>
+                                {
+                                    columns
+                                        .into_iter()
+                                        .enumerate()
+                                        .map(|(idx, key)| (key, &values[idx]))
```
filipgolem, line 82:
```suggestion
                            .map(|row| match row {
                                serde_json::Value::Array(row_values)
                                    if columns.len() == row_values.len() =>
                                {
                                    columns
                                        .into_iter()
                                        .enumerate()
                                        .map(|(idx, key)| (key, &row_values[idx]))
```

filipgolem (approved):
Good job. I would just rename some variables (see suggestions).

core/serv-api/src/lib.rs
```diff
@@ -55,14 +56,49 @@ impl CommandOutput {
...
+                            if let serde_json::Value::Array(values) = high_value {
+                                values.len()
+                            } else {
+                                0
+                            }
+                        })
+                        .filter(|size| *size != columns.len())
+                        .collect::<Vec<usize>>()
+                        .len()
+                        == 0;
```
mfranciszkiewicz, line 71: shorter version without `collect`:

```rust
let columns_size_eq_values_size = !values.iter().any(|row| match row {
    serde_json::Value::Array(values) => values.len() != columns.len(),
     _ => true,
});
```

biryukovmaxim:
Should I add new commits to resolve discussion? Or need only one commit(ammend/rebase) and force-push to remote?

core/serv-api/src/lib.rs
filipgolem: Or maybe this to avoid negating twice?
```rust
let columns_size_eq_values_size = values.iter().all(|row| match row {
    serde_json::Value::Array(values) => values.len() == columns.len(),
     _ => false,
});

```

filipgolem (approved):
Approved. Consider changing "not any not equal" to "all equal": https://github.com/golemfactory/yagna/pull/1723#discussion_r759106236

core/serv-api/src/lib.rs
```diff
@@ -55,14 +56,40 @@ impl CommandOutput {
...
-                        "{}",
-                        serde_json::to_string_pretty(&serde_json::json!({
-                            "headers": columns,
-                            "values": values
-                        }))
-                        .unwrap()
-                    )
+                    let columns_size_eq_values_size = !values.iter().any(|row| match row {
+                        serde_json::Value::Array(values) => values.len() != columns.len(),
+                        _ => true,
```
filipgolem, line 61:
```suggestion
                    let columns_size_eq_values_size = values.iter().all(|row| match row {
                        serde_json::Value::Array(values) => values.len() == columns.len(),
                        _ => false,
```

biryukovmaxim: yep, it looks better!

filipgolem (approved):

maaktweluit:
Hey @biryukovmaxim 

Thanks for your contribution!
The PR is approved but we need 3 more steps before we can merge this:
- Updating to master
- Checking if the automated test `goth` works properly after merging
- Check if there is any other tooling dependent on this

The first 2 should be simple, after updating to master we can see the goth build run and make an issue in the `goth` repo when it fails.

For the last one i made a note on our board, we will look into this and update you here :)

etam:
I would suggest a further refactoring (may be as a next step after this PR), to change `ResponseTable` in `core/serv-api/src/lib.rs` from
```rust
pub struct ResponseTable {
    pub columns: Vec<String>,
    pub values: Vec<serde_json::Value>,
}
```
to
```rust
pub struct ResponseTable<T: serde::Serialize>(Vec<T>);
```
or something like that.

nieznanysprawiciel (approved):

nieznanysprawiciel:
We updated all tools depending on this code.
PR will be merged after tests will pass

Issue #1735, opened 2021-11-30
[ya-provider] Enable multiple networks


jiivan:
Why: Operators want to use both polygon and mainnet networks

What: Enable selecting multiple networks

Issue #1734, opened 2021-11-30
[golemsp] Don't use YA_PAYMENT_NETWORK


jiivan:
Why: It's used by `ya-provider` that has different meaning since -rc4

What: Use different env variable name

etam:
Or you can use this env variable, but unset before launching ya-provider

PR #1736, opened 2021-11-30
[ya-provider] Allow multiple payment networks

jiivan:
fixes: #1735 
fixes: https://github.com/golemfactory/yagna-triage/issues/167
fixes: #1734


maaktweluit (approved):
LGTM!

Tested it locally and it is putting both mainnet and polygon driver address properly in the demand.

PR #1739, opened 2021-12-01
Net - p2p direct messaging

nieznanysprawiciel:
Resolves: https://github.com/golemfactory/ya-relay/issues/13
required: https://github.com/golemfactory/ya-relay/pull/48

This PR doesn't solve session expiration NET problems:
https://github.com/golemfactory/ya-relay/issues/64


.env-template
```diff
@@ -100,3 +97,19 @@ NODE_NAME=__YOUR_NODE_NAME_GOES_HERE__
...
+## Central Net configuration.
+#CENTRAL_NET_HOST=3.249.139.167:7464
+
+## Hybrid NET configuration
+
+# ya-net p2p client will listen on this address.
+YA_NET_BIND_URL=udp://0.0.0.0:11500
+
+# Address of relay server.
+YA_NET_RELAY_HOST=127.0.0.1:7464
```
maaktweluit, line 115: no newline at end of file is bad for git

core/serv/src/main.rs
```diff
@@ -390,14 +390,17 @@ impl ServiceCommand {
...
                         ("hyper", log::LevelFilter::Info),
                         ("reqwest", log::LevelFilter::Info),
                         ("tokio_core", log::LevelFilter::Info),
                         ("tokio_reactor", log::LevelFilter::Info),
                         ("trust_dns_resolver", log::LevelFilter::Info),
                         ("trust_dns_proto", log::LevelFilter::Info),
                         ("web3", log::LevelFilter::Info),
+                        ("tokio_util", log::LevelFilter::Off),
+                        ("mio", log::LevelFilter::Off),
+                        ("ya_persistence::executor", log::LevelFilter::Debug),
```
maaktweluit, line 403: always on debug?

.env-template
```diff
@@ -100,3 +97,19 @@ NODE_NAME=__YOUR_NODE_NAME_GOES_HERE__
 
 # Yagna service REST API application key token.
 YAGNA_APPKEY=__GENERATED_APP_KEY__
+
+# Choose NET type to use. Central NET is used by default, but will be removed
+# in future version to use more decentralized solutions.
+#YA_NET_TYPE=hybrid
```
maaktweluit, line 103: Why are the env variable changes also mixed into the p2p direct messaging changes?

This makes the PR a lot less clean.

Did you also update GOTH to be compatible with the env change?

maaktweluit (commented):
LGTM! will approve after the updates on ya-relay side ( and answers about goth )

.env-template
nieznanysprawiciel: There are no changes to ENV themselves. I added new variables, but didn't change any. Here I added the documentation only.

There is a change in resolving relay server address and it was necessary for yagna to work.
Goth passes as you probably saw ;)

core/serv/src/main.rs
nieznanysprawiciel: They were really annoying, but maybe I will remove this

maaktweluit: shouldnt they be on info / warn then?

.env-template
maaktweluit: ahh, i was a bit too quick on the comment there, sorry.

Yes, goth seems to work so its good :)

core/serv/src/main.rs
maaktweluit: i see you removed it :) will resolve the comment

PR #1758, opened 2021-12-08
Quotation marks in properties resolver

biryukovmaxim:
added unit tests
new type in enum PropertyValue and Literal
added special case for parsing string included escaped chars

### DIFF
--- a/core/market/resolver/src/resolver/prop_parser.rs
+++ b/core/market/resolver/src/resolver/prop_parser.rs
@@ -2,7 +2,6 @@ use std::str;
 use std::string::String;
 
 use nom::digit;
-use nom::AsBytes;
 use nom::IResult;
 
 named!(prop_def <&str, &str>,
@@ -187,7 +186,7 @@ pub enum Literal<'a> {
 named!(
     val_literal<Literal>,
     alt!(
-        str_literal
+        string_literal
             | version_literal
             | datetime_literal
             | true_literal
@@ -195,7 +194,6 @@ named!(
             | decimal_literal
             | number_literal
             | list_literal
-            | string_literal
     )
 );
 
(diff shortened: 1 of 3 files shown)


Wiezzel (commented):
Renaming `val` to `other` adds a lot of noise here and seems unrelated to the issue.

biryukovmaxim:
@Wiezzel,
I see, agree that this is not related to the issue. but without this renaming it is difficult to read/analyze/navigate the code that implements the comparison. the difference between val and value is not obvious. can I save changes in a separate commit of the same pr? Or should do another pr?

Wiezzel:
@biryukovmaxim Another PR would be the best option from reviewer's point of view but making another commit in this PR seems acceptable for me.

core/market/resolver/src/resolver/prop_parser.rs
```diff
@@ -209,14 +213,39 @@ named!(
...
+                ),
+                String::new(),
+                |acc, fragment| acc + str::from_utf8(fragment).unwrap()
+            ),
+            |str| Literal::String(str)
+        ),
+        char!('"')
+    ))
+);
+
```
mfranciszkiewicz, line 248: Maybe I'm missing something, but I don't see how `Literal::String` would be chosen over `Literal::Str` while parsing.

If we skip the unescaping part, the `Literal::String` type wouldn't be needed and parsing could be done via e.g.:

```rust
named!(
    string_literal<Literal>,
    ws!(do_parse!(
        val: delimited!(
            tag!("\""),
            escaped!(none_of!("\\\""), '\\', one_of!("\\\"")),
            tag!("\"")
        ) >> (Literal::Str(str::from_utf8(val).unwrap()))
    ))
);
```

biryukovmaxim: string_literal used here:
https://github.com/golemfactory/yagna/pull/1758/files#diff-b14d1e21873636609b0a7924e3830facdd507a38ad770f5f318c053774ca74acR446

I didn't find the way how I can parse string including escaped chars with nom crate version 2.*
I think it's not the best option, but acceptable workaround.

Another solution is upgrade nom and use something like this: https://docs.rs/nom/5.1.2/nom/combinator/fn.all_consuming.html in str_literal func

stranger80: I must admit I'd prefer going in the direction pointed by @mfranciszkiewicz above. 
I get the need of introducing `Literal::String()` but when it comes to parsing escaped quotes, use the simpler macro written by Marek, just "unescape" the result (by replacing \" with ").

stranger80 (changes requested):
I must admit I'd prefer going in the direction pointed by @mfranciszkiewicz above. 
I get the need of introducing `Literal::String()` but when it comes to parsing escaped quotes, use the simpler macro written by Marek, just "unescape" the result (by replacing \" with ").

Also, add relevant unit test in core/market/resolver/tests/prop-parser.rs, eg:

```
#[test]
fn parse_prop_value_from_literal_string_with_escaped_quotes() {
    assert_eq!(
        parse_prop_value_literal("\"\\\"king\\\" \\\"node\\\"\""),
        Ok(Literal::String("\"king\" \"node\""))
    );
}
```

core/market/resolver/src/resolver/prop_parser.rs
biryukovmaxim: I have made the parser simpler. Workaround is removed. Unit test is added

stranger80 (approved):
Yep, looks good, cheers!

mfranciszkiewicz (approved):

PR #1769, opened 2021-12-19
Mid-agreement payments

mfranciszkiewicz:


mfranciszkiewicz:
note: `hybrid-net` integration tests are using an outdated build of `ya-provider`

```
[2021-12-21T09:37:59.472+0000 INFO  ya_provider::provider_agent] Starting ya-provider. Version: 0.2.0 (1887661 2021-12-20 build #189).
```

agent/provider/src/payments/payments.rs
```diff
@@ -912,14 +954,14 @@ impl Actor for Payments {
...
-            let debit_checker = provider_ctx.debit_checker.clone();
             provider_ctx
                 .debit_checker
                 .send(Subscribe(payment_addr.recipient()))
                 .await
                 .map_err(|_| log::error!("Subscribing to DebitNotes deadline checker failed."))
                 .ok();
-            check_debit_notes_events(provider_ctx, debit_checker).await;
+
+            check_debit_notes_events(provider_ctx).await;
```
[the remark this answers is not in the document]
nieznanysprawiciel: And the second thing: Provider should terminate Agreement with `DebitNoteNotPaid` `Reason`, if he isn't paid.

agent/provider/src/provider_agent.rs
```diff
@@ -186,7 +186,10 @@ impl ProviderAgent {
         for preset in presets {
             let pricing_model: Box<dyn PricingOffer> = match preset.pricing_model.as_str() {
                 "linear" => match std::env::var("DEBIT_NOTE_INTERVAL") {
-                    Ok(val) => Box::new(LinearPricingOffer::default().interval(val.parse()?)),
+                    Ok(val) => {
+                        let interval = humantime::parse_duration(&val)?;
+                        Box::new(LinearPricingOffer::default().interval(interval.as_secs() as f64))
+                    }
```
nieznanysprawiciel, line 192: I don't thing `LinearPricingOffer` needs to store interval. We are adding it to Offer in `DebitNoteInterval` negotiator, so we can remove it from here

agent/provider/src/payments/payments.rs
```diff
@@ -209,7 +211,7 @@ async fn send_debit_note(
         activity_id: debit_note_info.activity_id.clone(),
         total_amount_due: cost_info.cost,
         usage_counter_vector: Some(json!(cost_info.usage)),
-        payment_due_date: None,
+        payment_due_date: debit_note_info.payment_timeout.map(|dur| Utc::now() + dur),
```
nieznanysprawiciel, line 214: Spec says, that DebitNotes should be sent relative to Agreement timestamp `approved_ts`. This complicates things, but we need to have absolute reference point. Current Provider implementation isn't caring about this, so it requires changes

agent/provider/src/market/termination_reason.rs
```diff
@@ -29,11 +32,37 @@ pub enum BreakReason {
     )]
     #[strum(message = "DebitNotesDeadline")]
     DebitNotesDeadline(chrono::Duration),
+    #[display(fmt = "Requestor rejected the debit note: {:?}", _0)]
```
nieznanysprawiciel, line 35: How will this be printed? This is debug formatting so I don't expect anything pleasant..

agent/provider/src/payments/payments.rs
```diff
@@ -579,12 +679,22 @@ impl Handler<UpdateCost> for Payments {
...
+                    let delta = Utc::now() - started;
+                    let delay = if update_interval > delta {
+                        // schedule the next debit note as close to `update_interval` as possible
+                        (update_interval - delta).to_std()?
+                    } else {
+                        // the process lasted longer than `update_interval`, continue immediately
+                        Duration::from_secs(0)
+                    };
+
+                    ctx.notify_later(msg, delay);
```
nieznanysprawiciel, line 697: It seems for me, that if `ctx.notify_later(msg, delay);` won't be exact enough, the error will accumulate, because we don't depend on value stored in `agreement.last_send_debit_note`. 

Am I right?

nieznanysprawiciel (commented):
Please check my comments

agent/provider/src/market/termination_reason.rs
mfranciszkiewicz: https://github.com/golemfactory/ya-client/blob/4c304836de47be0633f13246da96023979a313f8/model/src/payment/rejection.rs

goth_tests/domain/payments/test_mid_payments.py
```diff
@@ -0,0 +1,99 @@
...
+            payments = await provider.api.payment.get_payments(after_timestamp=ts)
+            for payment in payments:
+                amount += float(payment.amount)
+                ts = payment.timestamp if payment.timestamp > ts else ts
+            # prevent new debit notes in the last iteration
+            if i == ITERATION_COUNT - 2:
+                await requestor.destroy_activity(activity_id)
+                await provider.wait_for_exeunit_finished()
+
+        assert round(stats.amount, 12) == round(amount, 12)
```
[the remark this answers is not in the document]
mfranciszkiewicz: Could you elaborate? A missing payment will terminate the activity, and the debit amount will not match the amount paid

agent/provider/src/payments/payments.rs
mfranciszkiewicz: Is it going to surpass the average network delay though? Assuming the previous solution was good enough, this one is bit more accurate. Using `agreement.last_send_debit_note` is more complex - e.g. when the `ctx` is blocked for time `t` we need to calculate the n-th next timestamp depending on `t`.

This piece of code could have been refactored into `actix::utils::IntervalFunc`; cancellation is a bit more tricky there but we would be free from tracking and calculating timestamps. I didn't want to do that in this PR, could be done in a follow up if you wish.

nieznanysprawiciel: Previous code was good enough, because we didn't track these intervals in relation to Agreement creation timestamp. I'm wondering how big error could accumulate after month of work for example.

My concern is: Is it possible that Agreement will be broken by someone due to this error? If not, we can ignore this problem

goth_tests/domain/payments/test_mid_payments.py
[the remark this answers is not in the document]
nieznanysprawiciel: In each loop iteration we should get only one event notifying us about payment, that has been made. But if you for example in first iteration receive 0 events and in second 2 events, than sum will still be correct.

Provider should break Agreement, but I assume, that this is what we are testing - if he will really break it. If not, than I'm not sure what this test is testing..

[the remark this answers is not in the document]
mfranciszkiewicz: The time of payment event receipt is irrelevant, the payment one is. This test verifies that in fact all debit notes have been paid and the underlying mechanism is working correctly. The test was never meant to guarantee that we’ll receive events one by one and it’s still possible to receive none or multiple events, especially in the first / second iteration.

If that makes any sense, I can verify that each event was paid in time. I see little value in it though, only an additional maintenance cost.

agent/provider/src/payments/payments.rs
mfranciszkiewicz: It’s tightly connected with the debit note acceptance timeout. Previously it was equally likely for that timestamp to drift off.

Let’s fix it in a follow-up PR

agent/provider/src/market/termination_reason.rs
nieznanysprawiciel: Please display only information that are really useful in this context and avoid debug formatting. This message will be passed to Agreement termination reason, so should be pretty

agent/provider/src/payments/payments.rs
nieznanysprawiciel: Ok, please make issue for this

agent/provider/src/market/termination_reason.rs
mfranciszkiewicz: I've changed this line to log only the `rejection_reason` from `Rejection`. `RejectionReason` is a plain enum (so will only result in variant names) but doesn't implement `Display`. 

This enum originates from `ya_client` - I'd rather not implement a custom `to_string` but modify the target crate instead. As of now, the message shout be looking ok

nieznanysprawiciel (approved):

Issue #1755, opened 2021-12-08
0.9.0 with hybrid net broken, without it works


maaktweluit:
why
- we want working hybrid net

what
- debug the broken providers and provide a fix

jiivan:
In 0.9 `relay_addr` is definitely broken as it trats `config.host` as an env key name.
https://github.com/golemfactory/yagna/blob/a78ed7ee52ca5bd3e6005d840249add607eed6d1/core/net/src/hybrid/service.rs#L57-L58

In master it's much better:

https://github.com/golemfactory/yagna/blob/master/core/net/src/hybrid/service.rs#L57-L58

PR #1794, opened 2022-01-12
[fix] 0.9.0 with hybrid net broken, without it works

jiivan:
fixes: #1755

### DIFF
--- a/core/net/src/hybrid/service.rs
+++ b/core/net/src/hybrid/service.rs
@@ -66,7 +66,7 @@ async fn relay_addr(config: &Config) -> anyhow::Result<SocketAddr> {
     let (host, port) = &host_port
         .split_once(":")
         .context("Please use host:port format")?;
-    let ip = resolver::resolve_dns_record_host(&host.to_string()).await?;
+    let ip = resolver::resolve_dns_record_host(&host).await?;
     let ip_port = format!("{}:{}", ip, port);
     let socket = ip_port
         .to_socket_addrs()?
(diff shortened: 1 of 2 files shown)


core/net/src/hybrid/service.rs
```diff
@@ -55,16 +55,24 @@ thread_local! {
...
             .unwrap_or_else(|_| DEFAULT_NET_RELAY_HOST.to_string()),
-    }
-    .to_socket_addrs()?
-    .next()
-    .expect("relay address needed"))
+    };
+    log::trace!("resolving host_port: {}", host_port);
+    let (host, port) = &host_port
+        .split_once(":")
+        .context("Please use host:port format")?;
```
nieznanysprawiciel, line 68: Maybe it could be easier to use https://docs.rs/url/latest/url/struct.Url.html

utils/networking/src/resolver.rs
```diff
@@ -44,20 +44,21 @@ pub async fn resolve_dns_record(request_url: &str) -> anyhow::Result<String> {
         .ok_or(anyhow::anyhow!("Invalid url: {}", request_url))?
         .to_string();
 
+    let address = resolve_dns_record_host(&request_host).await?;
+    Ok(request_url.replace(&request_host, &address))
+}
+
+pub async fn resolve_dns_record_host(host: &String) -> anyhow::Result<String> {
```
nieznanysprawiciel, line 51:
```suggestion
pub async fn resolve_dns_record_host(host: &str) -> anyhow::Result<String> {
```

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

PR #1801, opened 2022-01-24
Better DebitNote interval precision relative to agreement approval time

mfranciszkiewicz:
This PR replaces the current `DebitNote` scheduling implementation (with interval skew allowed) with an approval-time-centric one, where the interval is strict. When the Provider is unable to keep up with the interval, DebitNotes are scheduled for the next valid interval tick.

The implementation is not  based on `IntervalFunc`. Send invocations, which take longer than the send interval, would require additional synchronization to detect overlapping calls.

Context:
Requestors will reject providers who send `DebitNotes` more often that the agreed update interval.

### DIFF
--- a/agent/provider/src/payments/payments.rs
+++ b/agent/provider/src/payments/payments.rs
@@ -27,6 +27,7 @@ use ya_utils_actix::deadline_checker::{
 use ya_utils_actix::{actix_signal_handler, forward_actix_handler};
 
 use crate::execution::{ActivityDestroyed, CreateActivity};
+use crate::interval::RelativeInterval;
 use crate::market::provider_market::NewAgreement;
 use crate::market::termination_reason::BreakReason;
 use crate::tasks::{AgreementBroken, AgreementClosed, BreakAgreement};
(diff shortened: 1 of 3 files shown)


agent/provider/src/payments/payments.rs
```diff
@@ -99,6 +100,47 @@ struct CostsSummary {
...
+        self.iteration = i;
+
+        Ok(delay.to_std()?)
+    }
+
+    pub fn current(&self) -> DateTime<Utc> {
+        self.base + (self.interval * self.iteration)
+    }
+}
+
```
nieznanysprawiciel, line 143: I like the fact, that this is separate structure. Could you move it to separate file?

agent/provider/src/payments/payments.rs
```diff
@@ -632,71 +678,63 @@ impl Handler<UpdateCost> for Payments {
...
-                            .map(|agreement| agreement.last_send_debit_note = last_debit_note + update_interval);
-                    }
 
-                    // Don't bother, if previous debit note was sent successfully or not.
-                    // Schedule UpdateCost for later.
+                        // A note regarding short debit note intervals:
+                        // If sending a DebitNote note takes longer than the interval duration,
+                        // the next DebitNote will be scheduled at the next possible interval,
+                        // relative to agreement approval date, and based on current time.
+                        let delay = msg.interval_ctx.advance()?;
```
nieznanysprawiciel, line 730: And this is fully correct behavior for me

nieznanysprawiciel (dismissed):

PR #1809, opened 2022-01-31
Handling closed session part1

nieznanysprawiciel:
partially
resolves: https://github.com/golemfactory/ya-relay/issues/64

### DIFF
--- a/core/net/Cargo.toml
+++ b/core/net/Cargo.toml
@@ -12,9 +12,14 @@ hybrid-net = []
 
 [dependencies]
 ya-core-model = { version = "^0.5", features=["net", "identity"] }
-ya-relay-client = { git = "https://github.com/golemfactory/ya-relay.git", rev = "9458a94a49d16686174c8e8fd7a1f6dee2e8f5c9" }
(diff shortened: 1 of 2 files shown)


maaktweluit (approved):

core/net/Cargo.toml
```diff
@@ -12,9 +12,14 @@ hybrid-net = []
 
 [dependencies]
 ya-core-model = { version = "^0.5", features=["net", "identity"] }
-ya-relay-client = { git = "https://github.com/golemfactory/ya-relay.git", rev = "9458a94a49d16686174c8e8fd7a1f6dee2e8f5c9" }
-ya-relay-core = { git = "https://github.com/golemfactory/ya-relay.git", rev = "9458a94a49d16686174c8e8fd7a1f6dee2e8f5c9" }
-ya-relay-proto = { git = "https://github.com/golemfactory/ya-relay.git", rev = "9458a94a49d16686174c8e8fd7a1f6dee2e8f5c9", features = ["codec"] }
+
+ya-relay-client = { git = "https://github.com/golemfactory/ya-relay.git", rev = "eeb2ff2e482166e4715434eac6235bad4886de72" }
```
maaktweluit, line 16: would be nice if we only need the client import here, it can re-expose the other parts it needs

nieznanysprawiciel: Good idea. Let's do it in next update to `ya-relay`

PR #1811, opened 2022-02-01
Make RUST_LOG override all internal config

maaktweluit:
and made default config simpler with a RUST_LOG string


nieznanysprawiciel (approved):

mfranciszkiewicz:
This implementation works only partially and requires a different approach

Issue #1820, opened 2022-02-04
VPN doesn't work correctly using the newest 0.10.0-rc3


shadeofblue:
running the http and ssh examples from yapapi on the newest yagna `0.10.0-rc3` results in:

```
[<HttpService running on q53 [ 0x44df807cd832393b254378ec33efb65adee837b8 ]>, <HttpService running on sycamore [ 0xa0cf793e2bb1764067cd4126cba444f901fd1b41 ]>]
local HTTP request: {}
sending a remote request '/' to <HttpService running on sycamore [ 0xa0cf793e2bb1764067cd4126cba444f901fd1b41 ]>
Error handling request
Traceback (most recent call last):
  File "/home/blue/.envs/yapapi/lib/python3.9/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
    resp = await self._request_handler(request)
  File "/home/blue/yapapi/examples/http-proxy/http_proxy.py", line 54, in request_handler
    response, status = await instance.handle_request(request.path_qs)
  File "/home/blue/yapapi/examples/http-proxy/http_proxy.py", line 130, in handle_request
    async with ws_session.ws_connect(
  File "/home/blue/.envs/yapapi/lib/python3.9/site-packages/aiohttp/client.py", line 1117, in __aenter__
    self._resp = await self._coro
  File "/home/blue/.envs/yapapi/lib/python3.9/site-packages/aiohttp/client.py", line 769, in _ws_connect
    raise WSServerHandshakeError(
aiohttp.client_exceptions.WSServerHandshakeError: 504, message='Invalid response status', url=URL('ws://127.0.0.1:7465/net-api/v1/net/f238ee56b77449d6b24371e2c411a395/tcp/192.168.0.3/80')
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f4328eb5400>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7f4328dcd520>, 253729.431425161)]']
connector: <aiohttp.connector.TCPConnector object at 0x7f4328eb5c10>
```

ssh:

```
[...]

PR #1825, opened 2022-02-04
VPN compatibility fix for central NET

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/1820


jiivan (approved):
This should do it for the migration period.
The `/u` prefix was short for __unreliable__, but I like UDP better. :)

PR #1823, opened 2022-02-04
Remove outdated nodes info ya-relay-client revision

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/1807

PR #1827, opened 2022-02-08
Update ya-relay for smoltcp update, lock rand_core version to match

maaktweluit:


nieznanysprawiciel (approved):

maaktweluit:
Will be fixed in #1823

Issue #1832, opened 2022-02-09
NET - VpnControl over TCP


jiivan:
Use TCP for VpnControl packets (stream goes thru UDP)

PR #1833, opened 2022-02-09
VpnControl over TCP

jiivan:
fixes: #1832


nieznanysprawiciel (approved):
Everything looks fine. Only `DuoEndpoint` name could be better, but I have no idea, so approving

jiivan:
> Everything looks fine. Only `DuoEndpoint` name could be better, but I have no idea, so approving

Yes, I thought about TwinEndpoint or DualEndpoint but decided that Duo sounds good enough.

Issue #1831, opened 2022-02-09
Port payment batching from `feature/win10` into `master`


mfranciszkiewicz:
What:
- port the account configuration and batching mechanism from the `feature/win10` branch
- update account configuration for goth tests

Why:
Batching is a required optimization for mid-agreement payments. Debit note payments need to be delayed and grouped within an acceptable time window, so that the sum of payment fees is lowered. Currently, accepting a debit note (max 4 min) causes an instant payment submission, which is non-optimal in terms of costs.

PR #1841, opened 2022-02-15
Payment batching

mfranciszkiewicz:
Payments are now grouped by the sender and payee addresses for the chosen payment platform. Batches are sent within a time window of the nearest `payment_due_date` (minus some constant payment processing overhead).

Payment batching is enabled by default when using `yagna payment init --sender` and can be disabled via the `--send-immediately` flag. Within an `accounts.json` file it's enough to specify a `"send": "auto"` mode; use the legacy value of `"send": true` to disable batching.

The original code was ported from branch `feature/win10`


nieznanysprawiciel (commented):
We decided to give up batching at this point

mfranciszkiewicz:
https://github.com/golemfactory/yagna/issues/1831#issuecomment-1056669808

PR #1846, opened 2022-02-17
Implementation of automatic alloc release (due to timeouts)

wkargul:
This draft PR consists of implementation of automatic allocations release (after the specified timeout has passed). Goth tests creation for this impl is still in progress.

### DIFF
--- a/core/payment/src/api/allocations.rs
+++ b/core/payment/src/api/allocations.rs
@@ -1,7 +1,10 @@
-// Extrnal crates
+use std::time::Duration;
+// External crates
 use actix_web::web::{delete, get, post, put, Data, Json, Path, Query};
 use actix_web::{HttpResponse, Scope};
+use chrono::Utc;
 use serde_json::value::Value::Null;
+use ya_client_model::NodeId;
 
 // Workspace uses
 use ya_agreement_utils::{ClauseOperator, ConstraintKey, Constraints};
@@ -61,12 +64,27 @@ async fn create_allocation(
         Err(e) => return response::server_error(&e),
     }
 
-    let dao: AllocationDao = db.as_dao();
+    let db_ = db.clone();
+    let dao = db.as_dao::<AllocationDao>();
+
     match async move {
         let allocation_id = dao
             .create(allocation, node_id, payment_platform, address)
             .await?;
-        Ok(dao.get(allocation_id, node_id).await?)
+        let allocation_id_ = allocation_id.clone();
+        match dao.get(allocation_id, node_id).await? {
+            None => Ok(None),
+            Some(allocation) => {
+                let deadline = allocation.timeout.clone().unwrap_or(Utc::now());
+                tokio::task::spawn_local(async move {
+                    let deadline = deadline - Utc::now();
+                    tokio::time::delay_for(deadline.to_std().unwrap_or(Duration::from_secs(0)))
+                        .await;
+                    release_alloc(db_, allocation_id_, node_id.clone()).await;
+                });
+                Ok(Some(allocation))
+            }
+        }
     }
     .await
     {
@@ -122,12 +140,8 @@ async fn release_allocation(
 ) -> HttpResponse {
     let allocation_id = path.allocation_id.clone();
     let node_id = id.identity;
-    let dao: AllocationDao = db.as_dao();
-    match dao.release(allocation_id, node_id).await {
-        Ok(true) => response::ok(Null),
-        Ok(false) => response::not_found(),
-        Err(e) => response::server_error(&e),
-    }
+
+    release_alloc(db, allocation_id, node_id).await
 }
 
 async fn get_demand_decorations(
@@ -165,4 +179,17 @@ async fn get_demand_decorations(
         properties,
         constraints,
     })
+}
+
+pub async fn release_alloc(
+    db: Data<DbExecutor>,
+    allocation_id: String,
+    node_id: NodeId,
+) -> HttpResponse {
+    let dao = db.as_dao::<AllocationDao>();
+    match dao.release(allocation_id, node_id).await {
+        Ok(true) => response::ok(Null),
+        Ok(false) => response::not_found(),
+        Err(e) => response::server_error(&e),
+    }
 }
--- a/core/payment/src/lib.rs
+++ b/core/payment/src/lib.rs
@@ -1,6 +1,7 @@
 #![allow(dead_code)] // Crate under development
 #![allow(unused_variables)] // Crate under development
 use crate::processor::PaymentProcessor;
+use chrono::Utc;
 use futures::FutureExt;
 use std::time::Duration;
 use ya_core_model::payment::local as pay_local;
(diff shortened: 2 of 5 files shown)


core/payment/src/lib.rs
```diff
@@ -73,3 +75,30 @@ impl PaymentService {
...
+        .get_filtered(None, None, None, None, None)
+        .await
+    {
+        allocations.iter().for_each(|a| {
+            let allocation = a.clone();
+            let db = db.clone();
+            let deadline = allocation.timeout.clone().unwrap_or(Utc::now());
+            tokio::task::spawn_local(async move {
+                let deadline = deadline - Utc::now();
+                tokio::time::delay_for(deadline.to_std().unwrap_or(Duration::from_secs(0))).await;
```
nieznanysprawiciel, line 91: Please propose solutions, what should we do in case of conversion failure

core/payment/src/lib.rs
```diff
@@ -73,3 +75,30 @@ impl PaymentService {
...
+        .as_dao::<dao::AllocationDao>()
+        .get_filtered(None, None, None, None, None)
+        .await
+    {
+        allocations.iter().for_each(|a| {
+            let allocation = a.clone();
+            let db = db.clone();
+            let deadline = allocation.timeout.clone().unwrap_or(Utc::now());
+            tokio::task::spawn_local(async move {
+                let deadline = deadline - Utc::now();
```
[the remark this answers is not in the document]
wkargul: for `tokio::time::delay_for()` we use `Duration`. `duration` is fetched with this block of code :

`deadline.to_std().unwrap_or(Duration::from_secs(0))` 

`to_std()` -> This function errors when duration is less than zero. As standard library implementation is limited to non-negative values.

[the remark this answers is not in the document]
wkargul: Entire function `release_allocation_after()` is now in `core/payment/src/api/allocations.rs`.
As you can see below I decided to rewrite code to simplify and make it reading-friendly. In case of having deadline in the past calculated timestamp is negative and deadline stays being equal to zero. So the thread is not delayed and allocation release is exectuted instantly.

```rust
            let timestamp = timeout.timestamp() - Utc::now().timestamp();
            let mut deadline = 0u64;

            if timestamp.is_positive() {
                deadline = timestamp as u64;
            }

            tokio::time::delay_for(Duration::from_secs(deadline)).await;
```

core/payment/src/lib.rs
```diff
@@ -73,3 +75,30 @@ impl PaymentService {
...
+async fn release_allocation(db: DbExecutor) {
+    if let Ok(allocations) = db
+        .as_dao::<dao::AllocationDao>()
+        .get_filtered(None, None, None, None, None)
+        .await
+    {
+        allocations.iter().for_each(|a| {
+            let allocation = a.clone();
+            let db = db.clone();
+            let deadline = allocation.timeout.clone().unwrap_or(Utc::now());
```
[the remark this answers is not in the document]
wkargul: In case of timeout is set to `None` allocation is not released automatically at any point in time. 

The only way to release ALL the allocations is to use CLI command `yagna payment release-allocations` (already implemented). We can think about adding the option of releasing allocation by ID. Currently CLI command clear all the allocations.

core/payment/src/lib.rs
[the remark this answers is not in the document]
wkargul: Resolved with timestamp calculation mechanism above.

core/payment/src/lib.rs
wkargul: Other mechanism for deadline calculation has been implemented. This should be now bug-resistant.

wkargul:
I simplified allocation deadline calculation and implemented new yagna command for clearance of all existing allocations. Also a few cosmetic changes made.

This should solve  issues linked to the task. Please review and let me know if you need any clarification.

core/payment/src/api/allocations.rs
```diff
@@ -54,19 +57,27 @@ async fn create_allocation(
...
+
     match async move {
         let allocation_id = dao
             .create(allocation, node_id, payment_platform, address)
             .await?;
-        Ok(dao.get(allocation_id, node_id).await?)
+        match dao.get(allocation_id, node_id).await? {
+            None => Ok(None),
+            Some(allocation) => {
+                release_allocation_after(db_, allocation.clone(), Some(node_id.clone())).await;
```
maaktweluit, line 77: Why is this function not on the "Data Access Object"? this would remove the ugly db_ clone

```suggestion
                dao.release_allocation_after(allocation.clone(), Some(node_id.clone())).await;
```

core/payment/src/lib.rs
```diff
@@ -48,7 +50,9 @@ impl Service for PaymentService {
 
 impl PaymentService {
     pub async fn gsb<Context: Provider<Self, DbExecutor>>(context: &Context) -> anyhow::Result<()> {
-        let db: DbExecutor = context.component();
+        let db = Data::new(context.clone().component());
+        api::allocations::release_allocations(db.clone(), false).await;
```
maaktweluit, line 54: i think its better to call this in or after `PaymentProcessor::new`

Also this struct already has the db context as `self.db_executor` so no need to clone again

core/payment/src/processor.rs
```diff
@@ -604,6 +606,10 @@ impl PaymentProcessor {
         Ok(result)
     }
 
+    pub async fn release_allocations(&self, db: Data<DbExecutor>) {
+        allocations::release_allocations(db, true).await
```
maaktweluit, line 610: looking a few lines above the database is accessed like this:
```
let existing_allocations = self
            .db_executor
            .as_dao::<AllocationDao>()
            .get_for_address(platform.clone(), address.clone())
            .await?;
```

would keep this structure to make all code consistent

maaktweluit (changes requested):
Great start! please check my comments.

Would try to use `.clone()` less, some times the objects you clone are already present, other times you can pass by reference.

core/payment/src/processor.rs
```diff
@@ -604,6 +607,48 @@ impl PaymentProcessor {
         Ok(result)
     }
 
+    /// This function releases allocations.
+    /// When `bool` is `true` all existing allocations are released immediately.
+    /// For `false` each allocation timestamp is respected.
+    pub async fn release_allocations(&self, force: bool) {
+        let existing_allocations = self
```
maaktweluit, line 614: You are calling `as_dao()` 3x in this function, maybe start with:
```
let dao = self.db_executor.as_dao::<AllocationDao>()
```

core/payment/src/processor.rs
```diff
@@ -1,3 +1,4 @@
+use crate::dao;
```
maaktweluit, line 1: I dont think this import is needed, it is actually confusing to me:
- all other dao's are imported by name
- in this file `let dao` is used pretty often

Would remove this import and use (already imported) `AllocationDao` instead

maaktweluit (changes requested):
LGTM!

One small comment left, then i'll approve :)

maaktweluit (dismissed):

maaktweluit (approved):

wkargul:
This PR consists of automatic allocations release mechanism.

Summary:
- This implementation guarantees backward compatibility. If allocation's timeout has been set to `None`, it stays unreleased (the only way to release it is use to CLI command). 
- Allocation's timeout is no longer ignored if set to `Some`. yagna automatically release allocations after the specified timeout has passed.
- Works with deamon restart as well.
- CLI command to release **ALL** allocations (regardless set timeouts) -> `yagna payment release-allocations`

PR #1851, opened 2022-02-24
Significantly increase the _default_ max agreement expiration

mfranciszkiewicz:
Default max agreement expiration has been increased to 100 years. The check was not removed since the value should be configurable by the provider


nieznanysprawiciel:
Will this value be used only in case of using `mid-agreement` payments or the only requirement is to have DebitNotes keep alive?

mfranciszkiewicz:
> Will this value be used only in case of using `mid-agreement` payments or the only requirement is to have DebitNotes keep alive?

Mid-agreement payments are, by default, enforced for computations longer than 10h. Below that limit, we simply apply a duration filter

nieznanysprawiciel (approved):

Issue #1819, opened 2022-02-04
"new release" notification despite running the latest official release


shadeofblue:
```
[2022-02-04T09:07:16.164+0100 WARN  ya_version::notifier] New Yagna Version 0.9.2 'v0.9.2 Gentle Sun' released 2022-01-10 is available! Update via `curl -sSf https://join.golem.network/as-provider | bash -`

^C[2022-02-04T08:27:24Z INFO  golemsp::service] Got ctrl+c. Bye!
```
[ ... ]

```
$ yagna --version
yagna 0.9.2 (d4a24786 2022-01-10 build #191)
```

johny-b:
Looks like this might be related to #1781

shadeofblue:
more observations: when running `yagna 0.9.2 (d4a24786 2022-01-10 build #191)`, I'm now getting:

```
[2022-02-24T10:24:58.688+0100 WARN  ya_version::notifier] New Yagna Version 0.10.0-rc5 'v0.10.0-rc5 #196' released 2022-02-10 is available! Update via `curl -sSf https://join.golem.network/as-provider | bash -`
```

question 1:

* why it is reporting an RC?

question 2:

* why is it reporting this particular RC? and not the newest one (rc10) - or one in between?

wkargul:
Hi @shadeofblue 

The linked PR should solve the issue. Please check my comment there for a bit more information.

Answer 1:
- Every new release information was fetched from github and put into local data base. yagna was comparing versions to the latest input in db. That's way yagna was reading every new release. Now it is filtered to skip reading "%rc%" versions as well as stop fetching versions without tag `latest` on github.

Answer 2:
- To be honest, I was not able to reproduce this situation. At my end it was always taking the latest RC. As we no longer fetch rc to yagna db I would not worry about it.  On other hand, if there is something what triggers faulty versions comparison it might cause issues in the future. But as far as I checked I see no bugs anymore. What is your opinion about it, @shadeofblue ?

PR #1854, opened 2022-02-28
Wkargul/new release notification

wkargul:
This PR consists of fix for "new release" notification. The main problem was with "rc" versions being reported as available for update. The second issue was redundant notification at yagna start.

Main issue solved with filter impl for "%rc%" versions fetched from local DB. What is more, I sealed github client reading the latest released versions of yagna. It is now reading only those with tag -> `latest`.

Second issue is solved with a slightly change in interval. `tick()` is replaced with `delay_for()` as we do not need this level of accuracy when counting the interval. `tick()` was also problematic as caused redundant logs at yagna startup (if newest yagna's version was avaialble).

It fixes https://github.com/golemfactory/yagna/issues/1819

### DIFF
--- a/core/version/src/db/dao.rs
+++ b/core/version/src/db/dao.rs
@@ -91,6 +91,7 @@ fn get_release(conn: &ConnType, ver: &str) -> anyhow::Result<Option<Release>> {
 fn get_pending_release(conn: &ConnType, include_seen: bool) -> anyhow::Result<Option<Release>> {
     let mut query = version_release
         // insertion_ts is to distinguish among fake-entries of `DBRelease::current`
+        .filter(release::version.not_like("%rc%"))
         .order((release::release_ts.desc(), release::insertion_ts.desc()))
         .into_boxed();
     if !include_seen {
--- a/core/version/src/github.rs
+++ b/core/version/src/github.rs
@@ -74,7 +74,6 @@ pub(crate) async fn check_running_release(db: &DbExecutor) -> anyhow::Result<Rel
             .repo_name(REPO_NAME)
             .bin_name("") // seems required by builder but unused
             .current_version("") // similar as above
-            .target_version_tag("latest")
             .build()?
             .get_release_version(running_tag)
     })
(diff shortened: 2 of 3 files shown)


maaktweluit (commented):
LGTM!

One thought: it seems we still store the current version in BD if its an RC, maybe we should also filter before inserting?

wkargul:
Hey @maaktweluit 
I would keep storing current running version. The only "rc" releases can be inserted into DB if we run "rc" intentionally. It might be helpful to have such record in db. What do you think? Suggestion about update for "rc" is no longer hinted by yagna, so the regular user is no longer spammed with irrelevant hints.

With commit `[c5456ed](https://github.com/golemfactory/yagna/pull/1854/commits/c5456ed4a4b20646a8b93bf505a5c3f01cd5509a)` I fixed fetching any other release than those having tag `latest` when checking running release. So at this point the only way to insert "rc" into DB is when running such version intentionally.

maaktweluit:
> I would keep storing current running version. The only "rc" releases can be inserted into DB if we run "rc" intentionally. It might be helpful to have such record in db. What do you think?
The database table is there to notify users about the new released versions, storing the "rc" releases is not helping this goal in any way, only add complexity without functionality.
Unless we properly implement pre-release vs full release.

> Suggestion about update for "rc" is no longer hinted by yagna, so the regular user is no longer spammed with irrelevant hints.

This was the main goal of this change 👍  :)
> 
> With commit [c5456ed](https://github.com/golemfactory/yagna/pull/1854/commits/c5456ed4a4b20646a8b93bf505a5c3f01cd5509a) I fixed fetching any other releases than those having tag `latest` when checking running release. So at this point the only way to insert "rc" into DB is when running such version intentionally.
What do you mean by "running such version intentionaly"? when a normal user runs an RC this should not be store in the database, IMHO.

wkargul:
If the database table is there to notify users about the new released versions it indeed does not make sense to store `-rc` releases there. I implemented `if` statement to omit DB update then.

I resigned from filtering `latest` releases when checking running release. We are fetching details of the release matching the specified version there, so it makes no sense to filter it by `latest` tag.

PR #1890, opened 2022-03-16
Handle a version check on RC

jiivan:
fixes: https://github.com/golemfactory/yagna-triage/issues/192


wkargul (approved):
LGTM!
I was wondering if we can just simply `unwrap()` but I assume that is safe enough here (no risk of panic either way).

PR #1892, opened 2022-03-20
erc20 driver: shared web3 client (hyper HTTP client connection pool)

mfranciszkiewicz:
This PR prevents a TCP connection leak in the erc20 driver by using a single HTTP connection pool per geth server address. Each TCP connection is counted in into the user's open file limit. The fix is mainly targeting active requestor nodes in the network


jiivan (approved):
LGTM, this should improve the situation.

PR #1896, opened 2022-03-23
Hybrid net: chunk broadcasted data

mfranciszkiewicz:
Chunks broadcasted data to fit within the UDP packet size limit. 

This PR **does not** try to resolve https://github.com/golemfactory/yagna/issues/1898 but implements a workaround for messages broadcasted by the market service


nieznanysprawiciel (approved):
Everything seems correct here.

PR #1901, opened 2022-03-28
Fix for releasing already released allocations

wkargul:


wkargul:
When allocation has timeout set to `Some` yagna spawns a background task and count the time till release. During runtime allocation can be consumed/release when used but it does not have any effect on the background task. So this is the root cause why "vague log" can be seen (yagna tries to release already released allocation). 

To prevent double release I implemented checking function which fetch info from DB about release status of the allocation.


The other solution (simpler one) might be just deletion of line of code from _core/payment/src/dao/allocation.rs_
```
log::warn!("Allocation {} not found. Release failed.", allocation_id);
```

etam:
I think this should be done in a single database query. You may add an argument to `fn release` saying whether missing allocation is an error or not and use it in `Ok(false) =>`.

core/payment/src/api/allocations.rs
```diff
@@ -182,3 +187,41 @@ async fn get_demand_decorations(
...
+            .release(allocation_id.clone(), node_id)
+            .await
+        {
+            Ok(true) => {
+                log::info!("Allocation {} released.", allocation_id);
+            }
+            Ok(false) => (),
+            Err(e) => {
+                log::warn!(
+                    "Allocation {} release failed. Db error occurred: {}",
```
nieznanysprawiciel, line 220:
```suggestion
                    "Releasing allocation {} failed. Db error occurred: {}",
```

core/payment/src/processor.rs
```diff
@@ -620,10 +625,16 @@ impl PaymentProcessor {
...
                         if force {
-                            dao.forced_release_allocation(allocation.allocation_id, None)
-                                .await
+                            release_allocation_after(
+                                db.clone(),
+                                allocation.allocation_id,
+                                None,
+                                None,
+                            )
+                            .await
```
nieznanysprawiciel, line 634: Problem with this solution is, that after this function returns, we can still have allocations that we wanted to force release. But it could remain like this for now

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

Issue #1895, opened 2022-03-23
Adjust default provider pricing to 0.025 GLM


nieznanysprawiciel:
Why:
- To high prices

What:
- Change defaults to 0.025 GLM CPU per hour

PR #1908, opened 2022-03-30
Adjust default provider pricing

jiivan:
fixes: #1895


cryptobench (commented):
This is a start but we might risk in the future that it needs a change again because the GLM token price fluctuates. I think we really should increase the UX by having a API lookup with a default value like 0.025 it can drop back to , IF it failed to look it up.

nieznanysprawiciel (approved):

Issue #1876, opened 2022-03-10
NET - error resolving hostname (connecting with relay)


nieznanysprawiciel:
What:
- Error message appears, when starting yagna with hybrid NET
`[2022-03-10T12:14:01.221+0100 WARN  ya_utils_networking::resolver] Error resolving hostname: relative URL without a base url=yacn2a.dev.golem.network`

It seems that it doesn't cause any problems later

PR #1917, opened 2022-03-30
Fix hostname resolving in hybrid net

jiivan:
fixes: #1876


core/net/src/hybrid/service.rs
```diff
@@ -73,7 +73,7 @@ async fn relay_addr(config: &Config) -> anyhow::Result<SocketAddr> {
     let (host, port) = &host_port
         .split_once(":")
         .context("Please use host:port format")?;
-    let ip = resolver::try_resolve_dns_record(&host).await;
+    let ip = resolver::try_resolve_dns_record(format!("udp://{}", host).as_str()).await;
```
mfranciszkiewicz, line 76: we should probably support `host`s with and without the scheme and apply the prefix conditionally

utils/networking/src/resolver.rs
```diff
@@ -61,13 +63,27 @@ pub async fn resolve_dns_record_host(host: &str) -> anyhow::Result<String> {
...
 }
 
-/// Try resolving hostname with `resolve_dns_record`. Return the original URL if it fails
-pub async fn try_resolve_dns_record(request_url: &str) -> String {
-    match resolve_dns_record(request_url).await {
+/// Try resolving hostname with `resolve_dns_record` or `resolve_dns_record_host`.
+/// Returns the original URL if it fails.
+pub async fn try_resolve_dns_record(request_url_or_host: &str) -> String {
+    lazy_static! {
+        static ref SCHEME_RE: Regex = Regex::new("^[a-z0-9\\-\\.]+?:").unwrap();
```
mfranciszkiewicz, line 70:
```suggestion
        static ref SCHEME_RE: Regex = Regex::new("(?i)^[a-z0-9\\-\\.]+?:").unwrap();
```

is case-insensitive

mfranciszkiewicz (approved):

PR #1919, opened 2022-03-30
NET - Target node filtering and unique session locks

mfranciszkiewicz:


goth_tests/domain/ya-provider/test_provider_break_agreement.py
```diff
@@ -227,7 +227,7 @@ async def test_provider_timeout_unresponsive_requestor(
         # First DebitNote will be sent after 15s. Let's wait with some margin.
         await providers[0].wait_for_agreement_broken(
             "Requestor is unreachable more than",
-            timeout=18,
+            timeout=60,
```
nieznanysprawiciel, line 230: It seems we have very big margin.
I'm not sure about this change....

mfranciszkiewicz: What _harm_ does this change bring? In some edge cases (bugs), the suite will run slower. In other scenarios, the change will save time needed for debugging the issue and tweaking the short time margin to another value.

E.g., the first debit note could theoretically be sent with a delay, in another send-out interval (relative to agreement approval time), pessimistically ~ 30s. We can also count in minor delays caused by new the network implementation, which still is a work in progress. But this does not mean that the implementation is not behaving correctly, i.e. if providers break the agreement if requestor goes offline

nieznanysprawiciel: Previous implementation should terminate Agreement after exactly 15s.
If new implementation isn't so strict, than there is no harm

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

PR #1920, opened 2022-03-30
add fiat value display to status screen

jedbrooke:
- [x] fetch coin price data from coingecko 
 - [x] print error message in the event api is down
 - [x] add display of current fiat value of GLM
 - [x] add display of user's current balance of GLM in fiat
 - [ ] add `golemsp settings set --currency` to select which currency the user would wish to see their balance in

example output:
example output

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,7 +115,7 @@ dependencies = [
  "http",
  "httparse",
  "indexmap",
- "itoa",
+ "itoa 0.4.7",
  "language-tags",
  "lazy_static",
  "log",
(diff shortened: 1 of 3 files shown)


jedbrooke:
some ideas I had for improvements:
 - cache price data for 15 min to reduce api calls
 - store user's price preference for currency in `golemsp settings set` (currently it is hardcoded to usd)
 - add coingecko attribution (we might be required to by their terms)
 
let me know if you have any other ideas for this feature

nieznanysprawiciel:
Thanks for this contribution, this is really great idea.
We would like to keep `golemsp` as simple as possible and provide more advanced features as extensions.
That's why we are going to make simple extensions system for `golemsp`: https://github.com/golemfactory/yagna/issues/1932

We think your code could be a great extension, when this feature will be ready.

tworec:
This code was included in yagna extension here https://github.com/golemfactory/yagna-usd

Issue #1931, opened 2022-04-06
Calling REST release-allocation should result in HTTP GONE if allocation was already released


nieznanysprawiciel:
Why:
- Returning Not Found is misleading

What:
- Calling REST release-allocation should result in HTTP GONE if allocation was already released

related to: https://github.com/golemfactory/yapapi/issues/923

Issue #1930, opened 2022-04-06
Cherry-pick release allocations fix to release/v0.10


nieznanysprawiciel:

wkargul:
This cherry-pick regards https://github.com/golemfactory/yagna/issues/1885

PR #1938, opened 2022-04-08
Wkargul/allocation release bug tmp

wkargul:
This PR consists of:
- HTTP GONE (410) implementation for getting already released allocations. When trying to accept debit note/invoice http error 404 is no longer served for already released allocations. It fixes https://github.com/golemfactory/yagna/issues/1931
- Fix for vague logs appearing when allocation timeout is passing by. It fixes  https://github.com/golemfactory/yagna/issues/1930

### DIFF
--- a/core/payment/src/dao/allocation.rs
+++ b/core/payment/src/dao/allocation.rs
@@ -2,9 +2,8 @@ use crate::error::{DbError, DbResult};
 use crate::models::allocation::{ReadObj, WriteObj};
 use crate::schema::pay_allocation::dsl;
 use bigdecimal::BigDecimal;
-use chrono::{DateTime, NaiveDateTime, Utc};
+use chrono::NaiveDateTime;
 use diesel::{self, ExpressionMethods, OptionalExtension, QueryDsl, RunQueryDsl};
-use std::time::Duration;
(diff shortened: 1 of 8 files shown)


nieznanysprawiciel (commented):
Everything is ok, except you don't return `Gone` in case of release allocation REST endpoint

core/payment/src/dao/allocation.rs
```diff
@@ -152,12 +152,31 @@ impl<'c> AllocationDao<'c> {
...
 
-    pub async fn release(&self, allocation_id: String, owner_id: Option<NodeId>) -> DbResult<bool> {
+    pub async fn release(
+        &self,
+        allocation_id: String,
+        owner_id: Option<NodeId>,
+    ) -> DbResult<AllocationReleaseStatus> {
         let id = allocation_id.clone();
         do_with_transaction(self.pool, move |conn| {
+            let mut query = dsl::pay_allocation
```
nieznanysprawiciel, line 162: I think this code could be much simpler. First you can get allocation without filtering, than check conditions (owner_id and if allocation is released). Than you can update released field, knowing that it will succeed

nieznanysprawiciel (approved):

PR #1939, opened 2022-04-11
Added gas info in payment status table

szymek156:
Shows gas amount together with the currency name.
Displays N/A if there is no gas for given configuration,
like usage of zksync driver.


scx1332 (dismissed):

mfranciszkiewicz (approved):
looks fine to me

Issue #1798, opened 2022-01-24
Goth failure: 24.01


nieznanysprawiciel:
https://github.com/golemfactory/yagna/actions/runs/1737844657

wkargul:
I would like to extend this task with logs:
- goth test [#125](https://github.com/golemfactory/yagna/actions/runs/1737844657) -> test_custom_runtime_counter FAIL -> [goth-logs_master-125.zip](https://github.com/golemfactory/yagna/files/8464909/goth-logs_master-125.zip)
- goth test [#199](https://github.com/golemfactory/yagna/actions/runs/2112846002) -> test_provider_renegotiate_proposal FAIL -> [goth-logs_master-199.zip](https://github.com/golemfactory/yagna/files/8464972/goth-logs_master-199.zip)
- goth test [#201](https://github.com/golemfactory/yagna/actions/runs/2142655603) -> test_mid_agreement_payments FAIL -> [goth-logs_release-v0.10-201.zip](https://github.com/golemfactory/yagna/files/8464928/goth-logs_release-v0.10-201.zip)


I see mutual error for both runs run no. 199 and run no. 201 : 

`[2022-04-XXTXX:XXXX.XXXX+0000 DEBUG ya_service_bus::typed] Call to XXXXX : Transport error: hyper::Error(IncompleteMessage)
` 

It seems like server has troubles with connection so the payment request can not be handled in the proper way. Request is not retried once it fails. Thus that might be a reason for various test fails.

I assume that also https://github.com/golemfactory/gnt2 could be updated with newe version of `ganache` as this also may be a source of issues.

PR #1942, opened 2022-04-13
Fix panics in verify_tx

scx1332:
This PR fix panics in verify_tx function. Some more logs have been added as this might possibly help to understand issue causing bug as mentioned in https://github.com/golemfactory/yagna/issues/1798


wkargul:
The changes are included with this PR https://github.com/golemfactory/yagna/pull/1970. I am closing PR

Issue #1911, opened 2022-03-30
Add workaround to certificates issue to `yagna`


nieznanysprawiciel:
Why:
- Due to problem with certificates, some nodes aren't sending metrics

What:
- Set `SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt` variable in `yagna` if it doesn't exist
- Check if this file exists

solves: https://github.com/golemfactory/yagna-triage/issues/87

mfranciszkiewicz:
> Extra note: Adding SSL_CERT_FILE=/etc/pki/tls/certs/ca-bundle.crt doesn't solve the issue for everyone.

We should conditionally set either `SSL_CERT_FILE` or `SSL_CERT_DIR` depending on the distribution, e.g. as in https://manuals.gfi.com/en/kerio/connect/content/server-configuration/ssl-certificates/adding-trusted-root-certificates-to-the-server-1605.html

mfranciszkiewicz:
These actions are normally executed by `openssl-probe`, which was not included in the linux binary release build (resolved in https://github.com/golemfactory/yagna/pull/1944)

PR #1944, opened 2022-04-13
Fix linux binary release workflow

mfranciszkiewicz:
`openssl-probe` was never included in and executed by the linux release binary.

The binary build command featured only `openssl/vendored` whereas `static-openssl` implies both `openssl/vendored` and `openssl-probe`

Resolves https://github.com/golemfactory/yagna-triage/issues/87
Resolves https://github.com/golemfactory/yagna/issues/1911


jiivan (approved):
LGTM

Issue #1965, opened 2022-04-25
NET - measure ping between nodes


nieznanysprawiciel:
Why:
- We need to investigate network problems

What:
- Measure direct ping on session level
- Measure ping on GSB
  - TCP messages
  - UDP messages

PR #1966, opened 2022-04-25
Net - gsb pings measure

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/1965

### DIFF
--- a/core/net/src/hybrid/cli.rs
+++ b/core/net/src/hybrid/cli.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use futures::future::join_all;
 use futures::TryFutureExt;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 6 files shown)


core/net/src/hybrid/cli.rs
```diff
@@ -85,3 +97,74 @@ fn to_status_metrics(metrics: &mut ChannelMetrics) -> model::StatusMetrics {
...
+                    udp_ping,
+                })
+            })
+            .map(|future| future.map_err(|e| StatusError::RuntimeException(e.to_string())))
+            .collect::<Vec<_>>(),
+    )
+    .await
+    .into_iter()
+    .enumerate()
+    .filter_map(|(idx, result)| match result {
```
mfranciszkiewicz, line 156: `.map`

core/net/src/hybrid/cli.rs
```diff
@@ -85,3 +97,74 @@ fn to_status_metrics(metrics: &mut ChannelMetrics) -> model::StatusMetrics {
...
+                let udp_ping = udp_before.elapsed();
+
+                anyhow::Ok(GsbPingResponse {
+                    node_id: *id,
+                    node_alias: alias.clone(),
+                    tcp_ping,
+                    udp_ping,
+                })
+            })
+            .map(|future| future.map_err(|e| StatusError::RuntimeException(e.to_string())))
```
mfranciszkiewicz, line 150: It may be useful to know which flavor of ping failed - whether it was udp or a virtual tcp connection. Pings would have to be reordered though

core/net/src/hybrid/cli.rs
```diff
@@ -85,3 +97,74 @@ fn to_status_metrics(metrics: &mut ChannelMetrics) -> model::StatusMetrics {
...
+                let tcp_before = Instant::now();
+
+                ya_net::from(our_node_id)
+                    .to(*id)
+                    .service(ya_net::DIAGNOSTIC)
+                    .send(GsbRemotePing {})
+                    .timeout(Some(ping_timeout))
+                    .await???;
+
+                let tcp_ping = tcp_before.elapsed();
```
mfranciszkiewicz, line 131: Can we assume that for each new session there exists an established vtcp connection? It's not a problem currently, but might be one if we introduce e.g. a `connect` command

nieznanysprawiciel: Even if we can't, I wouldn't like to handle this. What would we achieve? That someone avoids confusion, that ping takes longer due to TCP handshake?

mfranciszkiewicz: We would always print the actual response time. Handling it might be just as easy as sending the tcp ping message twice or making sure that a TCP connection is always established for each application session

nieznanysprawiciel: I would choose the second solution. But still at this point we don't have this problem

mfranciszkiewicz (approved):

PR #1992, opened 2022-05-20
Scx1332/Small fix to identity

scx1332:
Small fix to identity implementation by reqc


core/identity/src/dao/identity.rs
```diff
@@ -77,6 +77,11 @@ impl<'c> IdentityDao<'c> {
                 diesel::insert_into(s::identity::table)
                     .values(&preconfigured_identity)
                     .execute(conn)?;
+                diesel::update(s::identity::table)
+                    .set(id_dsl::is_default.eq(false))
```
mfranciszkiewicz, line 81: This is not currently the case, although `preconfigured_identity.is_default` equal to `false` is legal. Since we're already assuming that this will be the default identity, we can enforce `preconfigured_identity.is_default = true;`

mfranciszkiewicz (approved):

core/identity/src/dao/identity.rs
scx1332: I don't understand this comment

scx1332 (commented):
legit

nieznanysprawiciel:
What problem is this PR trying to solve?

PR #2056, opened 2022-06-28
Net API v2. Yagna /net/status endpoint. Public ip info in Offers.

pwalski:


core/net/Cargo.toml
```diff
@@ -11,6 +11,8 @@ service = []
 hybrid-net = []
 
 [dependencies]
+ya-client = "0.6"
```
nieznanysprawiciel, line 14: Do you need `ya-client`? I suspect `ya-client-model` should be enough?

pwalski:
Besides tests (which fail partially because it is a fork) there are following things to do:
- merging v1 and v2 scopes for VPN api
- tests of /net/info
- I have not debugged/tested yet (at the moment of writing comment it debugging in progress) how Offers gets serialized now
- /net API Errors handling - right now only rudimentary handling of `anyhow` errors

agent/provider/src/provider_agent.rs
```diff
@@ -75,6 +76,7 @@ pub struct ProviderAgent {
     log_handler: LoggerHandle,
     networks: Vec<NetworkName>,
     keystore_monitor: FileMonitor,
+    net_api: NetApi,
```
pwalski, line 79: I placed `net_api` directly instead of turning it into `Actor` because it is used only in one async call [here ](https://github.com/pwalski/yagna/blob/839af3cd3e2074189e468dc5cf5f71dc530cb3f8/agent/provider/src/provider_agent.rs#L541)

utils/agreement-utils/src/typed_props.rs
```diff
@@ -62,6 +64,7 @@ impl NodeInfo {
         if let Some(subnet) = self.subnet {
             let _ = node.insert("debug".into(), serde_json::json!({ "subnet": subnet }));
         }
+        let _ = node.insert("is_public".into(), serde_json::json!({"is_public": self.is_public}));
```
nieznanysprawiciel, line 67: Hmmm maybe `net.is_public`

core/net/src/error.rs
```diff
@@ -0,0 +1,23 @@
+use actix_web::{ResponseError, HttpResponse};
+use ya_client_model::ErrorMessage;
+
+pub type Result<T> = std::result::Result<T, NetError>;
+
+#[derive(thiserror::Error, Debug)]
+pub enum NetError {
+    //TODO handle more specific cases
+    #[error("Some error")]
```
nieznanysprawiciel, line 9: Please be more specific ;)

core/net/src/central/rest_api.rs
```diff
@@ -0,0 +1,12 @@
...
+
+pub fn web_scope() -> Scope {
+    actix_web::web::scope(NET_API_V2_NET_PATH).service(get_info)
+}
+
+#[actix_web::get("/info")]
+async fn get_info() -> impl Responder {
+    let info = Info { public_ip: None };
+    HttpResponse::Ok().json(info)
+}
```
nieznanysprawiciel, line 12: As I commented in `ya-client` we could make `status` endpoint which returns more information

core/net/src/hybrid/rest_api.rs
```diff
@@ -0,0 +1,20 @@
...
+use super::client::ClientProxy;
+
+pub fn web_scope() -> Scope {
+    actix_web::web::scope(NET_API_V2_NET_PATH).service(get_info)
+}
+
+#[actix_web::get("/info")]
+async fn get_info() -> Result<impl Responder> {
+    //TODO do not create client every time
+    let client = ClientProxy::new().await?;
```
nieznanysprawiciel, line 15: This doesn't create Proxy every time, because internally there is `lazy_static` which returns the same instance.
But it would be better to pass it as `app_data`

pwalski: The split between hybrid and a central is a problem.
Above fn `pub fn web_scope() -> Scope` cannot be `async` which is required for `ClientProxy::new().await?`.
Going higher in call hierarchy I am ending outside of `hybrid` module so I cannot create net service `rest` method like this:
 `rest<CONTEXT: Provider<Self, ClientActor>>(ctx: &CONTEXT) -> actix_web::Scope`
 and also `ClientProxy` is neither public nor an Actor.

core/net/src/hybrid/rest_api.rs
```diff
@@ -0,0 +1,28 @@
...
+use ya_client_model::net::{Status, NET_API_V2_NET_PATH};
+
+use crate::error::Result;
+
+use super::client::ClientProxy;
+
+pub fn web_scope() -> Scope {
+    let scope = actix_web::web::scope(NET_API_V2_NET_PATH);
+    match ClientProxy::new() {
+        Ok(client) => scope.app_data(Data::new(client)).service(get_info),
```
mfranciszkiewicz, line 11: You can also create instances of `ClientProxy` in handlers (it's not expensive)

core/net/src/hybrid/client.rs
```diff
@@ -57,7 +57,7 @@ impl Actor for ClientActor {
         let addr = ctx.address();
         ctx.wait(
             async move {
-                ADDRESS.write().await.replace(addr);
+                ADDRESS.write().unwrap().replace(addr);
```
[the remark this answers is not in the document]
pwalski: Just moved it to [ClientActor::init](https://github.com/golemfactory/yagna/pull/2056/commits/0571926adaf850f11d637bd409b2d26b0a599a8f#diff-dd6d63f1d0c80bcb08c52a922fe7ee3b6b68ff68fc1a6ae3364b9f3f8925322bR49) so it will be called whenever [Services::gsb](https://github.com/golemfactory/yagna/pull/2056/commits/0571926adaf850f11d637bd409b2d26b0a599a8f#diff-ff490246b403bbf6555dcfcf6cb38d31758d138d6fadbe6cd529de60d4374a99R539) is, and to not mix order there is an ugly check+panic before calling [Services::rest](https://github.com/golemfactory/yagna/pull/2056/commits/0571926adaf850f11d637bd409b2d26b0a599a8f#diff-ff490246b403bbf6555dcfcf6cb38d31758d138d6fadbe6cd529de60d4374a99R562). (calling Services::gsb mut be called before Services::rest to initialize ClientProxy)

core/net/src/hybrid/rest_api.rs
pwalski: In REST endpoint handlers? It was the initial idea. Instead task evolved into initializing ClientProxy in Services::gsb and doing some checks to call it before Services::rest

core/net/src/hybrid/client.rs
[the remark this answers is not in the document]
pwalski: I moved back to `tokio::sync::RwLock` because I will call ClientProxy in REST handler which is already async, so `async move` can stay.

core/net/src/hybrid/rest_api.rs
pwalski: Moved back to creating `ClientProxy` in REST endpoint handlers, because the race condition seems to be caused by thread spawned in [Net::gsb](https://github.com/golemfactory/yagna/blob/master/core/net/src/hybrid/service.rs#L80). I am waiting for test results - this time should finally pass on GitHub too.

core/serv/src/main.rs
```diff
@@ -559,13 +559,9 @@ impl ServiceCommand {
                         .wrap(auth::Auth::default())
                         .route("/me", web::get().to(me))
                         .service(forward_gsb);
-                    if is_gsb_configured {
```
pwalski: I reverted everything to more or less original implementation (creating ClientProxy in [endpoint handler](https://github.com/golemfactory/yagna/pull/2056/commits/73ccff10c4fcd54d12fa384f09a7b02a59c2d14e#diff-3e6a65e63556a2ea137ec68bfeda8f092ae1351655220cf5e8a690ca1fcfa1a7R13)) because race condition seems to be caused by thread spawned in [Net::gsb](https://github.com/golemfactory/yagna/blob/master/core/net/src/hybrid/service.rs#L80), and I did not wanted to return some JoinHandle or sth to `ServiceCommand::run_command` because it is too long already.

core/net/src/central/rest_api.rs
```diff
@@ -1,22 +1,24 @@
...
 #[actix_web::get("/status")]
-async fn get_info() -> Result<impl Responder> {
+async fn get_info(node_id: web::Data<NodeId>) -> Result<impl Responder> {
     let status = Status {
-        node_id: NodeId::default(),
+        node_id: (**node_id).clone(),
         listen_ip: None,
         public_ip: None,
-        sessions: 0,
+        sessions: 1,
```
nieznanysprawiciel, line 21: You can get default id using gsb call. Check in `core/market/src/identity.rs` how to do this.

core/net/src/service.rs
```diff
@@ -34,6 +34,7 @@ pub struct Net;
 
 lazy_static::lazy_static! {
     pub(crate) static ref NET_TYPE: Arc<RwLock<NetType>> = Default::default();
+    pub(crate) static ref NODE_ID: Arc<RwLock<NodeId>> = Default::default();
```
nieznanysprawiciel, line 37: Please don't use `lazy_static` if not necassary

pwalski: Did not like it either. Just wanted to avoid repeating gsb calls.
Latest commit deletes them and uses gsb call in REST endpoint instead.

nieznanysprawiciel (approved):

PR #2075, opened 2022-07-07
Offer with payload-manifest capability

### LINKED ISSUE #1875: Outbound network - providers include `inet` capability in their offer
What:
- providers supporting the outbound network feature add an `inet` capability to their offers
- **Edit**: providers need to declare that they support payload manifests by setting 
the `golem.srv.caps.payload-manifest` property to `true`

Why:
- requestors can filter providers supporting outbound network

pwalski:
Resolves #1875

### DIFF
--- a/agent/provider/src/provider_agent.rs
+++ b/agent/provider/src/provider_agent.rs
@@ -14,13 +14,14 @@ use ya_agreement_utils::*;
 use ya_client::cli::ProviderApi;
 use ya_core_model::payment::local::NetworkName;
 use ya_file_logging::{start_logger, LoggerHandle};
-use ya_manifest_utils::Keystore;
+use ya_manifest_utils::{manifest, Feature, Keystore};
 
 use crate::config::globals::GlobalsState;
 use crate::dir::clean_provider_dir;
 use crate::events::Event;
 use crate::execution::{
-    GetExeUnit, GetOfferTemplates, Shutdown as ShutdownExecution, TaskRunner, UpdateActivity,
+    ExeUnitDesc, GetExeUnit, GetOfferTemplates, Shutdown as ShutdownExecution, TaskRunner,
+    UpdateActivity,
 };
 use crate::hardware;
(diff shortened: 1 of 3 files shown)


utils/agreement-utils/src/typed_props.rs
```diff
@@ -211,21 +225,34 @@ mod test {
...
                 inf: InfNodeInfo::default().with_mem(5.0).with_storage(50.0),
                 exeunit_info: serde_json::json!({"wasm.wasi.version@v".to_string(): "0.9.0".to_string()}),
                 multi_activity: false,
+                payload_manifest: true,
             },
             com_info: Default::default(),
             offer: OfferTemplate::default(),
         };
 
-        eprintln!(
```
pwalski, line 226: I had no idea what this test did. It was supposed to print to std:err only on error? Is something checking when test is executed?
I converted it to some rudimentary check of serialized offer because we have hardcoded str literals so it may at least prevent some typos.

agent/provider/src/provider_agent.rs
```diff
@@ -278,6 +277,25 @@ impl ProviderAgent {
...
+        inf_node_info: InfNodeInfo,
+        exeunit_desc: ExeUnitDesc,
+        offer: &OfferTemplate,
+    ) -> anyhow::Result<ServiceInfo> {
+        let exeunit_desc = exeunit_desc.build();
+        let support_payload_manifest = match offer.property(manifest::CAPABILITIES_PROPERTY) {
+            Some(value) => {
+                serde_json::from_value(value.clone()).map(|capabilities: Vec<Feature>| {
+                    capabilities.contains(&Feature::ManifestSupport)
+                })?
```
pwalski, line 290: It will NOT fail on missing `manifest::CAPABILITIES_PROPERTY` BUT it will fail on an unknown `Feature`.
Runtimes and Yagna daemons are distributed together, but I am not sure it should fail or just ignore and log it as warn.

Issue #1925, opened 2022-04-04
Multiple invoice events


johny-b:
Sometimes `yapapi` is flooded with maaany "invoice events" for the same invoice, e.g.
```
$ grep "Received invoice event: {'invoice_id': '24b6d2a2-7316-461b-bb52-bf96031a8f13'}" blender-yapapi-2022-04-04_13.22.00.log  | wc -l
40712
```

I've seen this twice:
1. 2022-04-01, around 9:10. I don't have `yapapi` log file, but `yagna` files are included. 
* `yagna` version: `0.9.1`.
* `yapapi` version: IDK, probably `master`
2. 2022-04-04, around 13:30, `yagna` database and `yapapi` log files included.
* `yagna` version: `yagna 0.10.0-reputation-2`
* `yapapi` version: `master`

UPDATE:
1. This sometimes happens. I've just run blender few times in a row and it happened 3 out of 9 (every third run? not sure about this).
2. These are all `<class 'ya_payment.models.invoice_settled_event.InvoiceSettledEvent'>`
3. I see this also on yapapi b0.9, not only on master

[yapapi_log.zip](https://github.com/golemfactory/yapapi/files/8410353/yapapi_log.zip)
[yagna-2022-04-04.zip](https://github.com/golemfactory/yapapi/files/8410363/yagna-2022-04-04.zip)
[yagna-2022-04-01.zip](https://github.com/golemfactory/yapapi/files/8410364/yagna-2022-04-01.zip)

johny-b:
Just a reminder - today I've had quite a lot of entries for a single debit note in my log file:
```
$ grep '5368905f-1cc6-4389-b554-800dff5cf4eb' log_1000-iter-2.log | wc -l
201885
```
:)

(as expected, it had a 0-milisecond timestamp: "timestamp":"2022-06-09T18:20:18Z")

etam:
Here's an example of this happening. Look for 4d1aef67-e5e9-4f5b-9159-3d4502e7d1e0.
[datadir.zip](https://github.com/golemfactory/yagna/files/9054154/datadir.zip)

PR #2086, opened 2022-07-12
Explicitly store invoice events' timestamp

jiivan:
fixes: #1925

### DIFF
--- a/core/payment/src/models/invoice_event.rs
+++ b/core/payment/src/models/invoice_event.rs
@@ -16,6 +16,7 @@ pub struct WriteObj {
     pub owner_id: NodeId,
     pub event_type: String,
     pub details: Option<String>,
+    pub timestamp: NaiveDateTime,
 }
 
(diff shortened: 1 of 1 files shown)


mfranciszkiewicz (commented):
Looks good, although let’s make sure that the issue is resolved

jiivan:
After local testing this solution seems to work as expected with one caveat:

Yagna stores timestamp value with nanosecond resolution i.e. `2022-07-14T13:25:47.670616597Z`, but yapapi anly takes microsecond part i.e. `datetime.datetime(2022, 7, 14, 13, 25, 47, 670616`. Thus in next invoice query _.670616597_ is greater than _.670616[000]_. In a result a single event will often be returned two times (the second time it will have status ACCEPTED and be ignored).

I've decreased resolution to match yapapi & haven't seen any duplicates since then.

wkargul (approved):

Issue #2088, opened 2022-07-13
Certificates management documentation


mfranciszkiewicz:
- [x] Create handbook chapter about outbound net (actually about Computation Payload Manifest)
  - [x] Computation Payload Manifest schema generated from `ya-manifest-utils` (might require publishing schema on github)
  - [x] Computation Payload Manifest schema documentation generated from schema (might require automated generating it from schema)
  - [x] Computation Payload Manifest configuration guide
  - [x] Example showing outbound network demand without signature/cert (case of whitelisted domains)
- [x] Create a handbook chapter on managing certificates in yagna
  - [x] Listing relevant CLI commands in handbook
  - [x] Certificates related part of Computation Payload Manifest guide
- [x] Example describing creation of self signed CA, making provider to use it, and creation of Demand with signed manifest.
  - [x] Example showing outbound network demand requiring signature/cert

PR #2090, opened 2022-07-18
Keystore with certificates support

### LINKED ISSUE #1873: Outbound network - support certificates and certificate chains
- [ ] #2093

- [x]  Keystore dir monitoring (the implementation in yagna does not trigger on changes inside a directory)

- [ ]  Distribution/setup of Golem CA cert (+ intermediate cert(s)) in a release build or through https://join.golem.network/as-provider script - will be done within:  #1874
- [ ] #2134

What:
- implement X.509 certificate and certificate chain support in the keystore
  - verify cert expiration dates
- use the public signed key within the certificate to verify the signatures of payload manifests

Why:
- the keystore is public-key based and 
  - there are no expiration dates for the keys
  - it's not possible to create a key chain
- GolemFactory can issue certificates for third-party developers in the future

pwalski:
Resolves #1873


utils/manifest-utils/src/policy.rs
```diff
@@ -90,119 +92,83 @@ impl FromStr for Match {
...
 
-    pub fn insert(
+    pub fn verify_signature(
         &self,
-        key: impl Into<Box<[u8]>>,
-        scheme: Option<String>,
-        name: Option<String>,
+        cert: String,
+        sig: String,
+        sig_alg: String,
```
pwalski, line 143: It should be an enum with supported algorithms.

utils/manifest-utils/src/policy.rs
```diff
@@ -90,119 +92,83 @@ impl FromStr for Match {
...
+    pub fn load(cert_dir: impl AsRef<Path>) -> anyhow::Result<Self> {
+        let mut store = X509StoreBuilder::new()?;
+        let cert_dir = std::fs::read_dir(cert_dir)?;
+        for dir_entry in cert_dir {
+            let cert_file = dir_entry?;
+            let cert_file = cert_file.path();
+            let mut cert_file = File::open(cert_file)?;
+            let mut cert_buffer = Vec::new();
+            cert_file.read_to_end(&mut cert_buffer)?;
+            for cert in X509::stack_from_pem(&cert_buffer)? {
```
pwalski, line 120: Could support both DER and PEM (& PEM stacks)

pwalski: Alternatively consider adding lookup
https://docs.rs/openssl/latest/openssl/x509/store/struct.X509StoreBuilder.html#method.add_lookup
Precisely this
https://docs.rs/openssl/latest/openssl/x509/store/struct.X509Lookup.html#method.add_dir
(I failed to figure out how to create X509Lookup so I could call #add_dir)

pwalski: DER & PEM support done.

agent/provider/src/main.rs
```diff
@@ -42,7 +42,6 @@ async fn main() -> anyhow::Result<()> {
         Commands::Preset(presets_cmd) => presets_cmd.run(config),
         Commands::Profile(profile_cmd) => profile_cmd.run(config),
         Commands::ExeUnit(exe_unit_cmd) => exe_unit_cmd.run(config),
-        Commands::Keystore(keystore_cmd) => keystore_cmd.run(config),
```
nieznanysprawiciel, line 45: As we discussed: We need cli for adding removing certificates

agent/provider/src/market/negotiator/builtin/manifest.rs
```diff
@@ -25,28 +25,27 @@ impl NegotiatorComponent for ManifestSignature {
             return Ok(NegotiationResult::Ready { offer });
         }
 
-        let manifest = match demand.get_property::<String>(DEMAND_MANIFEST_PROPERTY) {
+        let data = match demand.get_property::<String>(DEMAND_MANIFEST_PROPERTY) {
             Err(Error::NoKey(_)) => return Ok(NegotiationResult::Ready { offer }),
             Err(e) => return rejection(format!("invalid manifest type: {:?}", e)),
-            Ok(s) => match decode_manifest(s) {
+            Ok(s) => match decode_data(s) {
```
nieznanysprawiciel, line 31: What is reason to have two functions instead of one `decode_manifest`?

agent/provider/src/market/negotiator/builtin/manifest.rs
```diff
@@ -25,28 +25,27 @@ impl NegotiatorComponent for ManifestSignature {
...
                 Ok(manifest) => manifest,
-                Err(e) => return rejection(format!("invalid manifest: {:?}", e)),
+                Err(e) => return rejection(format!("invalid manifest encoding: {:?}", e)),
             },
         };
 
+        let manifest = decode_manifest(&data)?;
```
nieznanysprawiciel, line 37: Probably you want to reject proposal here, not return with error

agent/provider/src/market/negotiator/builtin/manifest.rs
pwalski: It is actually a source of a bug I realize I did. While ago I noticed that we extract manifest from demand two times. First to check if features are empty, and sedond time to verify signature. I "fixed" it by extracting it once and reusing it later to verify signature, therefore I started verifying decoded manifest (instead of base64 encoded as I should).

This line reveals another issue - I noticed that some `base64` data decoding is done by attempting to decode `base64` only, but manifest has fallback decoding using `snailquote::unescape`:
```
pub fn decode_manifest<S: AsRef<str>>(input: S) -> Result<AppManifest, Error> {
    match decode_base64(&input) {
        Ok(manifest) => Ok(manifest),
        Err(_) => decode_escaped_json(input),
    }
}

/** .. */

fn decode_escaped_json<S: AsRef<str>>(input: S) -> Result<AppManifest, Error> {
    let decoded = snailquote::unescape(input.as_ref())?;
    Ok(serde_json::de::from_str(&decoded)?)
}
```
I had no idea why manifest base64 encoded data has this special fallback decoding so I started using it on other encoded fields (it will not break anything and makes things more consistent).

agent/provider/src/market/negotiator/builtin/manifest.rs
pwalski: Fixed

agent/provider/src/startup_config.rs
```diff
@@ -64,15 +63,21 @@ pub struct ProviderConfig {
         env = "PROVIDER_LOG_DIR",
     )]
     pub log_dir: Option<DataDir>,
+    /// Certificates directory
+    #[structopt(
+        long,
+        set = clap::ArgSettings::Global,
+        env = "CERT_DIR",
```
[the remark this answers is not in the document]
pwalski: Will change it to `PROVIDER_CERT_DIR`.
There is a default `SSL_CERT_DIR` but I disabled monitoring it by openssl.
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_default_verify_paths.html
Actually the way I monitor cert directory is less than ideal - there seems to be a way to make openssl to do it automatically using [hash dir feature](https://docs.rs/openssl/latest/openssl/x509/store/struct.X509Lookup.html#method.hash_dir). 
When I initially tried to implement cert dir monitoring I failed to figure out how set up x509Lookup (this library is a thin wrapper using openssl over FFI, so almost a black box, and I found no examples how to set X509Lookup over a directory using this library). I will try again while working on cmdline args to manage certificates.

agent/provider/src/main.rs
pwalski: Will be done here: #2134

agent/provider/src/startup_config.rs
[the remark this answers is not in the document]
pwalski: Env variable got changed to `PROVIDER_CERT_DIR`. Lookup will be implemented here ##2134

utils/manifest-utils/src/policy.rs
pwalski: Rust `openssl` library does not support list of hash algorithms either. WIll leave it as a string for now.

agent/provider/src/startup_config.rs
```diff
@@ -64,15 +63,21 @@ pub struct ProviderConfig {
         env = "PROVIDER_LOG_DIR",
     )]
     pub log_dir: Option<DataDir>,
+    /// Certificates directory
+    #[structopt(
+        long,
+        set = clap::ArgSettings::Global,
```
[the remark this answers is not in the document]
pwalski: Keystore commands task: #2088

agent/provider/src/market/negotiator/builtin/manifest.rs
```diff
@@ -90,27 +85,36 @@ impl From<PolicyConfig> for ManifestSignature {
...
+    /// Verifies fields base64 encoding, then validates certificate, then validates signature, then verifies manifest content and returns it
+    fn verify_signature<S: AsRef<str>>(
+        &self,
+        demand: &ProposalView,
+        manifest: S,
+    ) -> anyhow::Result<()> {
+        let sig: String = demand.get_property(DEMAND_MANIFEST_SIG_PROPERTY)?;
+        log::trace!("sig_hex: {}", sig);
+        let sig_alg: String = demand.get_property(DEMAND_MANIFEST_SIG_ALGORITHM_PROPERTY)?;
+        log::trace!("sig_alg: {}", sig_alg);
```
nieznanysprawiciel, line 103: Do we still support these fields?

pwalski: Yes, because there can be various digest algorithms. The reason it is still a `String` is that rust `openssl` library does not provide any list of available options as a code (enum, types). As far as I remember original `openssl` lib can list digest algorithms through function.

nieznanysprawiciel: Ok, sorry. I thought that we have only `cert` field containing all necessary info, but I was wrong. Merging

nieznanysprawiciel (approved):

PR #2135, opened 2022-08-01
NET - query node by id (CLI)

mfranciszkiewicz:


core/net/src/hybrid/client.rs
```diff
@@ -134,6 +136,29 @@ proxy!(
...
+
+        Ok(FindNodeResponse {
+            identities,
+            endpoints,
+            seen: res.seen_ts,
+            slot: res.slot,
+            encryption: res.supported_encryptions,
+        })
+    }
+);
```
nieznanysprawiciel, line 161: I would prefer to have better structure for Node information in `ya-relay-client` than to do heavy mapping in macro

nieznanysprawiciel (approved):

PR #2136, opened 2022-08-02
Fix non-binding let on a synchronization lock

MrDarthShoe:
Clippy returned error:
``` 
error: non-binding let on a synchronization lock
   --> core/persistence/src/executor.rs:228:9
    |
228 |         let _ = pool.tx_lock.read().unwrap();
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::let_underscore_lock)]` on by default
    = help: consider using an underscore-prefixed named binding or dropping explicitly with `std::mem::drop`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_lock
```
    
Which means that quard is dropped in the same time it is locked. This PR provides fix for that


jiivan (approved):

wkargul:
To be merged before v.0.11 release.

PR #2151, opened 2022-08-08
Keystore cmdline

### LINKED ISSUE #2134: Restore `keystore` cmdline args
What:
Reintroduce commands to list add and remove certificates.

Why:
Requestors may issue own certificates which means Providers will need an easy way to manage them.

To determine:
- What should be the output of certificates list
- How to store certificates. Right now we have a monitored directory. Any certificate file copied there will be loaded to the store. PEM certificates can be merged together into one file. In order to support commands for managing certificates it would be better to have some constistent naming and to keep one certificate in one file. Fortunately `openssl` already supports so called `hash dirs` https://www.openssl.org/docs/manmaster/man3/X509_LOOKUP_hash_dir.html Yagna will maintain content of this directory, and `openssl` will lookup for new certificates on demand. Openssl already has a feature called `rehash` which automatically creates symlinks for certificates. Unfortunately it is not provided by rust `openssl`, and also it would be better to just rename files. To decide is whether we want to use `rehash` or implement it (it is fairly simple).

pwalski:
Resolves #2134

### DIFF
--- a/agent/provider/src/cli/keystore.rs
+++ b/agent/provider/src/cli/keystore.rs
@@ -1,0 +1,91 @@
+use std::collections::HashSet;
+use std::path::PathBuf;
+
+use structopt::StructOpt;
+
+use ya_manifest_utils::{util, Keystore, KeystoreLoadResult};
+
+use crate::startup_config::ProviderConfig;
+
+#[derive(StructOpt, Clone, Debug)]
+#[structopt(rename_all = "kebab-case")]
+pub enum KeystoreConfig {
+    /// List trusted X.509 certificates
+    List,
+    /// Add new trusted X.509 certificates
+    Add(Add),
+    /// Remove trusted X.509 certificates
+    Remove(Remove),
+}
+
+#[derive(StructOpt, Clone, Debug)]
+pub struct Add {
+    /// Paths to X.509 certificates (PEM or DER) or certificates chains
+    #[structopt(parse(from_os_str))]
+    certs: Vec<PathBuf>,
+}
+
+#[derive(StructOpt, Clone, Debug)]
+#[structopt(rename_all = "kebab-case")]
+pub struct Remove {
+    /// Certificate ids
+    ids: Vec<String>,
+}
+
+impl KeystoreConfig {
+    pub fn run(self, config: ProviderConfig) -> anyhow::Result<()> {
+        match self {
+            KeystoreConfig::List => list(config),
+            KeystoreConfig::Add(cmd) => add(config, cmd),
+            KeystoreConfig::Remove(cmd) => remove(config, cmd),
+        }
+    }
+}
+
+fn list(config: ProviderConfig) -> anyhow::Result<()> {
+    let cert_dir = cert_dir_path(&config)?;
+    let keystore = Keystore::load(&cert_dir)?;
+    util::print_cert_list_header();
+    keystore.visit_certs(util::print_cert_list_row)?;
+    Ok(())
+}
+
+fn add(config: ProviderConfig, add: Add) -> anyhow::Result<()> {
+    let cert_dir = cert_dir_path(&config)?;
+    let keystore_manager = util::KeystoreManager::try_new(&cert_dir)?;
+    match keystore_manager.load_certs(&add.certs)? {
+        KeystoreLoadResult::Loaded { loaded, skipped } => {
+            println!("Added certificates:");
+            util::print_cert_list(&loaded)?;
+            if !skipped.is_empty() {
+                println!("Certificates already loaded to keystore:");
+                util::print_cert_list(&skipped)?;
+            }
+        }
+        KeystoreLoadResult::NothingNewToLoad { skipped } => {
+            println!("No new certificate to add. Skipped:");
+            util::print_cert_list(&skipped)?;
+        }
+    }
+    Ok(())
+}
+
+fn remove(config: ProviderConfig, remove: Remove) -> anyhow::Result<()> {
+    let cert_dir = cert_dir_path(&config)?;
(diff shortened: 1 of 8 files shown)


utils/manifest-utils/src/util.rs
```diff
@@ -15,6 +27,276 @@ pub enum DecodingError {
...
+            let mut new_cert_path = self.cert_dir.clone();
+            new_cert_path.push(format!("{}.{}{}", file_stem, id, dot_extension));
+            let cert = cert.to_pem()?;
+            fs::write(new_cert_path, cert)?;
+        }
+        Ok(())
+    }
+}
+
+pub enum KeystoreLoadResult {
```
pwalski, line 285: It looks like if it should implement `Result`. Might rename it later.

agent/provider/src/cli/keystore.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use ya_manifest_utils::{util, Keystore, KeystoreLoadResult};
+
+use crate::startup_config::ProviderConfig;
+
+#[derive(StructOpt, Clone, Debug)]
+#[structopt(rename_all = "kebab-case")]
+pub enum KeystoreConfig {
+    /// List trusted X.509 certificates
+    List,
```
pwalski, line 14: Probably should add `Show` to display equivalent of `openssl x509 -text`, because `List` does not show all the data.

utils/manifest-utils/src/util.rs
```diff
@@ -15,6 +27,276 @@ pub enum DecodingError {
...
+
+    /// Copies certificates from given file to `cert_dir` and returns newly added certificates.
+    /// Certificates already existing in `cert_dir` are skipped.
+    pub fn load_certs(self, cert_paths: &Vec<PathBuf>) -> anyhow::Result<KeystoreLoadResult> {
+        let mut loaded = HashMap::new();
+        let mut skipped = HashMap::new();
+
+        for cert_path in cert_paths {
+            let mut new_certs = Vec::new();
+            let file_certs = parse_cert_file(&cert_path)?;
```
pwalski, line 157: Add clear error message.

utils/manifest-utils/src/util.rs
```diff
@@ -15,6 +27,276 @@ pub enum DecodingError {
...
+            .map_err(|err| anyhow::anyhow!(err))?;
+        let dot_extension = get_file_extension(&cert_path)
+            .map(|ex| format!(".{}", ex))
+            .unwrap_or(String::from(""));
+        for cert in certs.into_iter() {
+            let id = cert_to_id(&cert)?;
+            let mut new_cert_path = self.cert_dir.clone();
+            new_cert_path.push(format!("{}.{}{}", file_stem, id, dot_extension));
+            let cert = cert.to_pem()?;
+            fs::write(new_cert_path, cert)?;
```
pwalski, line 279: Add clear/readable error message.

pwalski: To `cert.to_pem()?;` too.

pwalski:
@nieznanysprawiciel  I added steps describing the way handling certificate files is done - https://github.com/golemfactory/yagna/issues/2134#issuecomment-1209285291

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

utils/manifest-utils/src/util.rs
```diff
@@ -15,6 +27,276 @@ pub enum DecodingError {
...
+pub fn print_cert_list_header() {
+    println!(
+        "{0: <32} {1: <24} {2}",
+        "ID", "Not After", "Subject"
+    );
+    println!(
+        "{0: <32} {1: <24} {2}",
+        "-----------", "-------", "---------"
+    );
+}
```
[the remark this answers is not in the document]
pwalski: I have not used it because table takes lots of horizontal space.

utils/manifest-utils/src/util.rs
```diff
@@ -15,6 +27,271 @@ pub enum DecodingError {
...
+            .map_err(|err| anyhow::anyhow!(err))?;
+        let mut new_cert_path = self.cert_dir.clone();
+        new_cert_path.push(file_name);
+        if new_cert_path.exists() {
+            let file_stem = get_file_stem(&new_cert_path).expect("Has to have stem");
+            let dot_extension = get_file_extension(&new_cert_path)
+                .map(|ex| format!(".{}", ex))
+                .unwrap_or(String::from(""));
+            for i in 0..u32::MAX {
+                let numbered_filename = format!("{0}.{1}{2}", file_stem, i, dot_extension);
```
MrDarthShoe, line 242: That should probably work and look better:
```suggestion
                let numbered_filename = format!("{file_stem}.{i}{dot_extension}");
```

MrDarthShoe: Oslo consider using it in other places where fmt is used

utils/manifest-utils/src/util.rs
```diff
@@ -15,6 +27,271 @@ pub enum DecodingError {
...
+
+        let removed: Vec<X509> = removed.into_values().collect();
+        Ok(KeystoreRemoveResult::Removed { removed })
+    }
+
+    /// Loads keychain file to `cert_dir`
+    fn load_as_keychain_file(&self, cert_path: &PathBuf) -> anyhow::Result<()> {
+        let file_name = get_file_name(cert_path)
+            .ok_or_else(|| format!("Cannot get filename of {:?}", cert_path))
+            .map_err(|err| anyhow::anyhow!(err))?;
```
MrDarthShoe, line 233: Consider constructing error directly inside `ok_or_else` without `map_err`.

PR #2152, opened 2022-08-08
Return app-key if it already exists

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2117

### DIFF
--- a/core/identity/src/cli/appkey.rs
+++ b/core/identity/src/cli/appkey.rs
@@ -68,7 +68,6 @@ impl AppKeyCommand {
                     .await
                     .map_err(anyhow::Error::msg)?
                     .unwrap();
-
                 Ok(CommandOutput::Object(serde_json::to_value(key)?))
             }
             AppKeyCommand::Drop { name, id } => {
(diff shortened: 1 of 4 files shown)


core/identity/src/cli/appkey.rs
```diff
@@ -66,8 +66,7 @@ impl AppKeyCommand {
                 let key = bus::service(model::BUS_ID)
                     .send(create)
                     .await
-                    .map_err(anyhow::Error::msg)?
-                    .unwrap();
+                    .map_err(anyhow::Error::msg)??;
```
nieznanysprawiciel, line 69: If you don't add anything to error message, you can just use `?` operator, because anyhow will be automatically converted

core/model/src/appkey.rs
```diff
@@ -24,6 +24,13 @@ impl Error {
             message: e.to_string(),
         }
     }
+
+    pub fn client(e: impl std::fmt::Display) -> Self {
```
nieznanysprawiciel, line 28: I would name it bad request. All errors 4xx are clients errors not only 400

nieznanysprawiciel (approved):

Issue #2145, opened 2022-08-04
Failed to accept an invoice or a debit note (UNIQUE constraint failed)


mfranciszkiewicz:
```
[2022-08-04T15:50:20.953+0200 ERROR yapapi.summary] [Job 1] Failed to accept an invoice or a debit note from 'ProviderInfo(id='0x0c609ee85eb3214ecb4c8eb3254186b194ce9412', name='BHS5-12.h', subnet_tag='hybrid')', reason: (400)
Reason: Bad Request
HTTP response headers: <CIMultiDictProxy('Content-Length': '107', 'Content-Type': 'application/json', 'Date': 'Thu, 04 Aug 2022 13:50:20 GMT')>
HTTP response body: {"message":"UNIQUE constraint failed: pay_debit_note_event.debit_note_id, pay_debit_note_event.event_type"}
```

On Provider side we are flooded with logs about DebitNote acceptance. 
Database from ``:
```
c58c82f4-7315-4f3a-897e-05eb358bb2f5|0xd1dd36c60c5b3bf723a93e16ce472ae0e66c664d|ACCEPTED|2022-07-29 12:31:13.562|
a676b7e9-4c65-48cc-b342-c26debfe056b|0xd1dd36c60c5b3bf723a93e16ce472ae0e66c664d|RECEIVED|2022-07-29 12:33:13.182|
a676b7e9-4c65-48cc-b342-c26debfe056b|0xd1dd36c60c5b3bf723a93e16ce472ae0e66c664d|ACCEPTED|2022-07-29 12:33:14.000|
fbc76fb3-e95a-44a9-b445-cc431e82296a|0xd1dd36c60c5b3bf723a93e16ce472ae0e66c664d|RECEIVED|2022-07-29 12:35:13.183|
```

It seems related to timestamp ending with `000`. Provider sends query to yagna on address `/payment-api/v1/debitNoteEvents?timeout=50&afterTimestamp=2022-07-29T12:33:14Z&appSessionId=DE1-4.h-755723`

nieznanysprawiciel:
It's interesting:
```
sqlite> select * from pay_debit_note_event where timestamp > "2022-07-29 12:33:14";
a676b7e9-4c65-48cc-b342-c26debfe056b|0xd1dd36c60c5b3bf723a93e16ce472ae0e66c664d|ACCEPTED|2022-07-29 12:33:14.000|
[...]

PR #2153, opened 2022-08-08
Create TimestampAdapter to unify timestamps format in db

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/2145

### DIFF
--- a/core/payment/src/dao/invoice_event.rs
+++ b/core/payment/src/dao/invoice_event.rs
@@ -13,7 +13,7 @@ use ya_client_model::NodeId;
 use ya_persistence::executor::{
     do_with_transaction, readonly_transaction, AsDao, ConnType, PoolType,
 };
-use ya_persistence::types::Role;
+use ya_persistence::types::{AdaptTimestamp, Role};
 
 pub fn create<T: Serialize>(
     invoice_id: String,
@@ -68,7 +68,7 @@ impl<'c> InvoiceEventDao<'c> {
                 .order_by(read_dsl::timestamp.asc())
                 .into_boxed();
             if let Some(timestamp) = after_timestamp {
-                query = query.filter(read_dsl::timestamp.gt(timestamp));
+                query = query.filter(read_dsl::timestamp.gt(timestamp.adapt()));
             }
             if let Some(app_session_id) = app_session_id {
                 query = query.filter(read_dsl::app_session_id.eq(app_session_id));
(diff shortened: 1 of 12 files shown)


core/persistence/src/timestamp.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::types::AdaptTimestamp;
+    use chrono::{NaiveDate, NaiveDateTime, NaiveTime};
+
+    #[test]
+    fn test_timestamp_adapter_formatting() {
```
MrDarthShoe, line 63: Consider using https://crates.io/crates/test-case crate, as this UT has the same logic multiple times

core/payment/src/models/invoice_event.rs
```diff
@@ -50,7 +50,7 @@ impl WriteObj {
             owner_id,
             event_type: event_type.to_string(),
             details,
-            timestamp,
+            timestamp: Utc::now().adapt(),
```
MrDarthShoe, line 53: It seems like whole block from lines 34 to 47 is unnecessary now. I wonder why warning is not emitted? :fearful:

core/payment/src/dao/invoice_event.rs
```diff
@@ -68,7 +68,7 @@ impl<'c> InvoiceEventDao<'c> {
                 .order_by(read_dsl::timestamp.asc())
                 .into_boxed();
             if let Some(timestamp) = after_timestamp {
-                query = query.filter(read_dsl::timestamp.gt(timestamp));
+                query = query.filter(read_dsl::timestamp.gt(timestamp.adapt()));
```
MrDarthShoe, line 71: [DISCUSSION_STARTER] I wonder is it possible to avoid these explicit fn calls and have it done somehow automatically?

nieznanysprawiciel: The problem is, that `NaiveDateTime` already implements `ToSql` trait, so I have to substitute new type here.

Other option would be to disable `chrono` feature in `diesel` which provides `ToSql` implementation. But I suspect we can't implement this trait ourselves, because both types are from external crates and rust doesn't allow this

core/payment/src/models/invoice_event.rs
nieznanysprawiciel: Thanks for noticing. I was convinced, I removed this.

core/persistence/src/timestamp.rs
nieznanysprawiciel: Good idea in general, thanks for hint.
This is hard to apply for all cases here, so I will leave implementation as is

MrDarthShoe (approved):

MrDarthShoe (approved):

Issue #2166, opened 2022-08-12
Investigate how PaymentDriver::init and payment init should work


kamirr:

kamirr:
Concerns #2116 and #2118

kamirr:
During the preparation of the draft PR (#2167), a couple issues have surfaced:
1. As of now, initializing the account as receiver-only protected the user from unintentional spending of funds. With `payment init` made implicit, this is now functionally a non-feature, as any attempt to allocate funds will automatically enable the capability to send funds.
2. `PaymentDriver::init` will be called on each allocation and invocation of `payment fund`, and has always been capable of being used this way (i.e. multiple `yagna payment init [--sender] [--receiver]`). Meaning there are undocumented semantics, that aren't exactly what one would call "initialization". These have been explained in more detail in the last commit of the PR ([Add assumption comments](https://github.com/golemfactory/yagna/pull/2167/commits/08d6fec73197106d2349050e0d8a5ce6f10a28af)).
3. `send` and `receive` (`: bool`) properties of an account can't be simply removed to drop the support for protection mentioned in (1), as it is actually important to differentiate how an account has been initialized. That is, the initialization might do different things depending on those properties and these properties are not just a soft protection for the end-user.

nieznanysprawiciel:
It seems current version is ok

Issue #2118, opened 2022-07-27
`yagna payment fund` calls `yagna payment init`


nieznanysprawiciel:
Why:
- We want to simplify user onboarding process and minimize number of commands he needs to execute
- `yagna payment init` call is not clear, why it is necessary

What:
- `yagna payment fund` should execute `payment init`, if it wasn't called by the user manually

MrDarthShoe:
It seems that `payment init` is not necessary for `payment fund`.

PR #2167, opened 2022-08-12
Make payment init optional

### LINKED ISSUE #2116: `yagna payment init` could be done during creating allocation
Why:
- We want to simplify user onboarding process and minimize number of commands he needs to execute
- `yagna payment init` call is not clear, why it is necessary

What:
- During creating allocation using REST api, payment init could be done automatically

kamirr:
Resolves #2116 & #2118

Relevant #2166


core/payment/src/api/allocations.rs
```diff
@@ -50,6 +52,42 @@ async fn create_allocation(
...
+                    return response::bad_request(
+                        &"paymentPlatform must be of the form driver-network-token",
+                    )
+                }
+            };
+
+        // The token needs to be uppercase excluding the initial 't', eg. tGLM
+        let token_no_prefix = match token.strip_prefix("t") {
+            Some(tok) => tok,
+            None => return response::bad_request(&"token must begin with 't'"),
```
nieznanysprawiciel, line 73: Only testnet tokens begin with `t`. Mainnet doesn't have this prefix.

core/payment/src/api/allocations.rs
```diff
@@ -50,6 +52,42 @@ async fn create_allocation(
         .unwrap_or(DEFAULT_PAYMENT_PLATFORM.to_string());
     let address = allocation.address.clone().unwrap_or(node_id.to_string());
 
+    // If the request contains information about the payment platform, initialize the account
+    // by setting the `send` field to `true`, as it is implied by the intent behing allocation of funds.
+    if let Some(platform) = &allocation.payment_platform {
+        // payment_platform is of the form driver-network-token
+        // eg. erc20-rinkeby-tglm
+        let [driver, network, token]: [&str; 3] =
+            match platform.split("-").collect::<Vec<_>>().try_into() {
```
nieznanysprawiciel, line 61: I'm not sure, what will happen in case we will have for example 4 or 2 elements after split. Can you ensure it won't panic?

core/payment/src/api/allocations.rs
nieznanysprawiciel: This is crucial, because it disables `mainnet` ;)

nieznanysprawiciel (changes requested):
I like your comments in code :)

Please change token parsing, because it disables `mainnet`

core/payment/src/api/allocations.rs
kamirr: It won't panic, an invalid number of elements will trigger the `Err(_e) => ...` match arm and return `response::bad_request(...)`.

core/payment/src/api/allocations.rs
kamirr: I've dropped token parsing and made it default to `None`.

nieznanysprawiciel (approved):

Issue #1658, opened 2021-10-26
REST and GSB APIs for listing agreements


nieznanysprawiciel:
Why:
- Request from Box team
    - This information will be displayed to users
- Other potential usages (in the future)
    - It will make Reloading Agent state after restart possible
    - Node retrospection (tools that would check content and state of yagna daemon)

What:
- [ ] yagna REST
- [ ] yagna GSB
- [ ] Agreements are listed only for identity that created them
- [ ] Ability to filter by `app_session_id`, date and state
- [ ] We need ability to list all Agreements (not only active)
- [ ] Update `ya-client`
    - [ ] openapi spec
    - [ ] rust client implementation

Wiezzel:
@nieznanysprawiciel What is the status of this?

nieznanysprawiciel:
Postponed due to higher priority things (for example basic authentication runtime). If anything in terms of priority changed, let us know

PR #2173, opened 2022-08-19
Add REST, GSB and CLI endpoints for listing and getting agreements

kamirr:
Adresses #1658 and #2175.

Depends on https://github.com/golemfactory/ya-client/pull/124 -- when it gets merged, the patched `ya-client` and `ya-client-model` crates may be updated to point to the revision on `master`.

### DIFF
--- a/core/market/src/cli.rs
+++ b/core/market/src/cli.rs
@@ -1,0 +1,100 @@
+use chrono::{DateTime, Utc};
+use structopt::StructOpt;
+use ya_client::model::market::{agreement::State, Role};
+use ya_core_model::market::{GetAgreement, ListAgreements};
+use ya_service_api::{CliCtx, CommandOutput, ResponseTable};
+use ya_service_bus::{typed as bus, RpcEndpoint};
+
+/// Market management
+#[derive(StructOpt, Debug)]
+pub enum Command {
+    Agreements(AgreementsCommand),
+}
+
+impl Command {
+    pub async fn run_command(self, ctx: &CliCtx) -> anyhow::Result<CommandOutput> {
+        match self {
+            Command::Agreements(agreements_cmd) => agreements_cmd.run_command(ctx).await,
+        }
+    }
+}
+
+#[derive(StructOpt, Debug)]
+pub enum AgreementsCommand {
+    List {
+        #[structopt(long, help = "Only show agreements with this state")]
+        state: Option<State>,
+        #[structopt(long, help = "Only show agreements after this date, rfc3339")]
+        before: Option<DateTime<Utc>>,
+        #[structopt(long, help = "Only show agreements before this date, rfc3339")]
+        after: Option<DateTime<Utc>>,
+        #[structopt(long, help = "Only show agreements with this app session id")]
+        app_session_id: Option<String>,
+    },
+    Get {
+        #[structopt(long, help = "Agreement ID, may be obtained via list-agreements")]
+        id: String,
+        #[structopt(long, help = "Your role in the agreement (Provider | Requestor)")]
+        role: Role,
+    },
+}
+
+impl AgreementsCommand {
+    pub async fn run_command(self, _ctx: &CliCtx) -> anyhow::Result<CommandOutput> {
+        match self {
+            AgreementsCommand::List {
+                state,
+                before,
+                after,
+                app_session_id,
+            } => {
+                let request = ListAgreements {
+                    state: state.into(),
+                    before_date: before,
+                    after_date: after,
+                    app_session_id,
+                };
+
+                let agreements = bus::service(ya_core_model::market::BUS_ID)
+                    .send(request)
+                    .await??;
+
+                let mut agreements_json = Vec::new();
+                for agreement in agreements {
+                    agreements_json.push(serde_json::to_value([
+                        agreement.id,
+                        agreement.role.to_string(),
+                        agreement.creation_ts.to_rfc3339(),
+                        agreement
+                            .approve_ts
+                            .map(|ts| ts.to_rfc3339())
+                            .unwrap_or("N/A".to_owned()),
+                    ])?);
+                }
+
+                Ok(ResponseTable {
+                    columns: vec![
+                        "id".to_owned(),
+                        "role".to_owned(),
+                        "created".to_owned(),
+                        "approved".to_owned(),
+                    ],
+                    values: agreements_json,
+                }
+                .with_header("\nMatching agreements:\n".to_owned()))
+            }
+            AgreementsCommand::Get { id, role } => {
+                let request = GetAgreement {
+                    agreement_id: id,
+                    role,
+                };
+
+                let agreement = bus::service(ya_core_model::market::BUS_ID)
+                    .send(request)
+                    .await??;
+
+                CommandOutput::object(agreement)
+            }
+        }
+    }
+}
--- a/core/market/src/db/dao/agreement.rs
+++ b/core/market/src/db/dao/agreement.rs
@@ -1,4 +1,4 @@
-use chrono::{NaiveDateTime, Utc};
+use chrono::{DateTime, NaiveDateTime, Utc};
 use diesel::prelude::*;
 
 use ya_client::model::market::Reason;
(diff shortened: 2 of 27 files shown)


core/market/src/db/dao/agreement.rs
```diff
@@ -58,9 +59,66 @@ pub enum AgreementDaoError {
...
 
 impl<'c> AgreementDao<'c> {
+    pub async fn list(
+        &self,
+        node_id: Option<NodeId>,
+        state: Option<AgreementState>,
+        before: Option<NaiveDateTime>,
+        after: Option<NaiveDateTime>,
+        app_session_id: Option<String>,
+    ) -> Result<Vec<AgreementListEntry>, AgreementDaoError> {
```
nieznanysprawiciel, line 74: This is database level. I would list Agreements here, without converting to `AgreementListEntry`.
Current solution has a few drawbacks:
- On database level you would like to have as general functions as possible. Here you limit yourself to information that `AgreementListEntry` contains
- You anyway have to query whole Agreements from database, so there is no gain from database processing and filtering. The situation would be different, if you would query only specific fields
- You have to convert to `ClientAgreement` on database level. But in whole rest of the Market you use database Agreement. So this function won't be useful in any other context.
- You had to add `AgreementDaoError::ClientAgreement` error, which doesn't belong here, because it is not database error.
- Having `ClientAgreement` here is against market design, because conversion to client structures happens only on most external REST level or GSB level

core/market/src/db/dao/agreement.rs
```diff
@@ -58,9 +59,66 @@ pub enum AgreementDaoError {
...
+                let role = match agreement.id.owner() {
+                    Owner::Provider => Role::Provider,
+                    Owner::Requestor => Role::Requestor,
+                };
+
+                let client_agreement = agreement
+                    .into_client()
+                    .map_err(AgreementDaoError::ClientAgreement)?;
+
+                result.push(AgreementListEntry::new(client_agreement, role));
```
nieznanysprawiciel, line 114: As in comment above, this could be done on other layer

core/model/src/market.rs
```diff
@@ -41,6 +41,33 @@ impl RpcMessage for GetAgreement {
     type Error = RpcMessageError;
 }
 
+/// Lists all agreements
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ListAgreements {
+    pub state: Option<String>,
+    pub before_date: Option<String>,
+    pub after_date: Option<String>,
```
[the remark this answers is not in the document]
kamirr: Right, it should be typed. Fixed.

core/market/src/db/dao/agreement.rs
kamirr: I removed the use of `ClientAgreement` entirely and now directly construct `AgreementListEntry` from the database agreement in REST and GSB endpoints.

Does this part look OK now?

core/market/src/market.rs
```diff
@@ -234,6 +235,41 @@ impl MarketService {
...
+        before: Option<DateTime<Utc>>,
+        after: Option<DateTime<Utc>>,
+        app_sesssion_id: Option<String>,
+    ) -> Result<Vec<AgreementListEntry>, AgreementError> {
+        let agreements = self
+            .db
+            .as_dao::<AgreementDao>()
+            .list(Some(id.identity), state, before, after, app_sesssion_id)
+            .await
+            .map_err(|e| AgreementError::Internal(format!("{:?}", e)))?;
```
nieznanysprawiciel, line 251: Could work without debug formatting?

core/market/src/market/agreement.rs
```diff
@@ -13,10 +9,50 @@ use crate::db::DbMixedExecutor;
...
+            id: agreement.id.into_client(),
+            creation_ts: naive_to_utc(agreement.creation_ts),
+            approve_ts: agreement.approved_ts.map(naive_to_utc),
+            role: role,
+        });
+    }
+
+    Ok(result)
+}
+
```
nieznanysprawiciel, line 55: You could use the second `list_agreements` function internally and pass parameters from `msg`

core/market/src/db/dao/agreement.rs
nieznanysprawiciel: It's ok

core/market/src/cli.rs
```diff
@@ -0,0 +1,100 @@
...
+use ya_client::model::market::{agreement::State, Role};
+use ya_core_model::market::{GetAgreement, ListAgreements};
+use ya_service_api::{CliCtx, CommandOutput, ResponseTable};
+use ya_service_bus::{typed as bus, RpcEndpoint};
+
+/// Market management
+#[derive(StructOpt, Debug)]
+pub enum Command {
+    Agreements(AgreementsCommand),
+}
```
nieznanysprawiciel, line 12: Like we discussed during demo it would be better to have command `yagna market agreement list` or something like this

nieznanysprawiciel (dismissed):

core/market/src/cli.rs
kamirr: That's how it already works in this PR (`yagna market agreements list`).

core/market/src/market.rs
```diff
@@ -234,6 +235,41 @@ impl MarketService {
...
+            let role = match agreement.id.owner() {
+                Owner::Provider => Role::Provider,
+                Owner::Requestor => Role::Requestor,
+            };
+
+            result.push(AgreementListEntry {
+                id: agreement.id.into_client(),
+                creation_ts: naive_to_utc(agreement.creation_ts),
+                approve_ts: agreement.approved_ts.map(naive_to_utc),
+                role: role,
```
wkargul, line 266:
```suggestion
                role,
```

core/market/src/market/agreement.rs
```diff
@@ -13,10 +9,50 @@ use crate::db::DbMixedExecutor;
...
+        let role = match agreement.id.owner() {
+            Owner::Provider => Role::Provider,
+            Owner::Requestor => Role::Requestor,
+        };
+
+        result.push(AgreementListEntry {
+            id: agreement.id.into_client(),
+            creation_ts: naive_to_utc(agreement.creation_ts),
+            approve_ts: agreement.approved_ts.map(naive_to_utc),
+            role: role,
```
wkargul, line 49:
```suggestion
            role,
```

core/model/src/market.rs
```diff
@@ -41,6 +42,33 @@ impl RpcMessage for GetAgreement {
     type Error = RpcMessageError;
 }
 
+/// Lists all agreements
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
```
wkargul, line 46:
```suggestion
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Default)]
```

core/model/src/market.rs
```diff
@@ -41,6 +42,33 @@ impl RpcMessage for GetAgreement {
...
+impl Default for ListAgreements {
+    fn default() -> Self {
+        ListAgreements {
+            state: None,
+            before_date: None,
+            after_date: None,
+            app_session_id: None,
+        }
+    }
+}
```
wkargul, line 64:
```suggestion

```

wkargul: use Default derive macro and delete implementation

wkargul (changes requested):
@kamirr just a few cosmetic changes

core/market/src/market.rs
kamirr: Done.

core/model/src/market.rs
kamirr: Done.

core/model/src/market.rs
kamirr: Done.

core/market/src/market/agreement.rs
kamirr: Done.

core/market/src/market.rs
kamirr: Now that I think about it, `e.to_string()` is better.

core/market/src/market/agreement.rs
kamirr: Not exactly, the other `list_agreements` requires a reference to MarketService, which can't be constructed just from `DbMixedExecutor`.

I don't like the duplicated logic either, but I don't see a clear way to extract it. You've done the same with your `get_agreement` functions, too.

wkargul (approved):
LGTM

core/market/src/market/agreement.rs
nieznanysprawiciel: `ServiceBinder::bind_with_processor` can pass additional generic parameter to bound functions. And `agreement::bind_gsb` could have `self` easily.

Anyway, approving

nieznanysprawiciel (dismissed):

nieznanysprawiciel (approved):

PR #2178, opened 2022-08-23
Allowed domains whitelist

### LINKED ISSUE #2082: Provider supports whitelist of allowed domains
Why:
- Whitelisting domains will allow community users to create dApps accessing blockchain without need to distribute their certificates or signing their images by us.

What:
- Provider have configurable whitelist with domain names
- We want to distribute whitelist with yagna releases
- During negotiations Provider will accept Proposals from Requestors if, their manifest includes access only to domains on the whitelist

Progress:

- [x] Serializable domain whitelist + matching utils (regex, strict) + using whitelist in Manifest Negotiator
- [x] Config file monitoring (similar as for Globals or Presets)
- [x] Cmdline for whitelist (list, add, remove)
- [x] Yapapi modifications (optional sig+sig_alg+cert)
- [x] negotiator functional tests 
- [x] cmdline functional tests

pwalski:
Resolves #2082


kamirr (dismissed):

agent/provider/src/cli/whitelist.rs
```diff
@@ -0,0 +1,200 @@
...
+    let domain_patterns = DomainPatterns::load_or_create(&config.domain_whitelist_file)?;
+    let mut domain_patterns = DomainPatternIds::from(domain_patterns);
+    let (added, skipped) = domain_patterns.add(add);
+    let domain_patterns: DomainPatterns = domain_patterns.into();
+    domain_patterns.save(&config.domain_whitelist_file)?;
+    if !added.is_empty() {
+        println_conditional(&config, "Added patterns:");
+        WhitelistTable::from(DomainPatterns { patterns: added }).print(&config)?
+    } else if config.json {
+        println_conditional(&config, "No new patterns to add.");
```
kamirr, line 70: In this case `println_conditional` will *never* print,

agent/provider/src/cli/keystore.rs
```diff
@@ -67,9 +68,15 @@ fn add(config: ProviderConfig, add: Add) -> anyhow::Result<()> {
...
+            println_conditional(&config, "No new certificate to add.");
+            println_conditional(&config, "Dropped duplicated certificates:");
             let certs_data = util::to_cert_data(&skipped)?;
-            print_cert_list(&config, certs_data)?;
+            if !config.json {
+                print_cert_list(&config, certs_data)?;
+            } else {
+                // no new certificate added, so empty list for json output
+                print_cert_list(&config, Vec::new())?;
+            }
```
kamirr, line 79: Wouldn't it be clearer to just do
```rust
let certs_data = util::to_cert_data(&skipped)?;
if !config.json {
    println!("No new certificate to add.")
    println!("Dropped duplicated certificates:")
    print_cert_list(&config, certs_data)?;
} else {
    // no new certificate added, so empty list for json output
    print_cert_list(&config, Vec::new())?;
}
```

agent/provider/src/cli/whitelist.rs
```diff
@@ -0,0 +1,200 @@
...
+    #[structopt(long = "type", short = "t", default_value = "strict")]
+    pattern_type: ArgMatch,
+}
+
+#[derive(StructOpt, Clone, Debug)]
+#[structopt(rename_all = "kebab-case")]
+pub struct Remove {
+    /// Domain whitelist pattern ids.
+    ids: Vec<String>,
+}
```
kamirr, line 42: Please add appropriate structopt `help="..."` attributes to explain the non-trivial commands and arguments (specifically `Add::{patterns, pattern_type}`), and their possible values.

agent/provider/src/cli/whitelist.rs
```diff
@@ -0,0 +1,200 @@
...
+        WhitelistTable::from(DomainPatterns {
+            patterns: Vec::new(),
+        })
+        .print(&config)?
+    }
+    if !skipped.is_empty() && !config.json {
+        println_conditional(&config, "Dropped duplicated patterns:");
+        if !config.json {
+            WhitelistTable::from(DomainPatterns { patterns: skipped }).print(&config)?
+        }
```
kamirr, line 81: Redundant logic:
* `println_conditional` will always print.
* The inner `if !config.json` will always execute
Both are because you already assert `config.json` in the outer if expression.

I suppose you could be using `println_conditional` everywhere for consistency, that would be fine. All in all I would prefer normal `println!` when possible to clear up the logic, but those issues won't block my approval.

agent/provider/src/cli/whitelist.rs
```diff
@@ -0,0 +1,200 @@
...
+        for pattern in domain_patterns.patterns {
+            table.add(pattern);
+        }
+        table
+    }
+}
+
+struct DomainPatternIds {
+    pattern_ids: HashMap<String, DomainPattern>,
+}
```
kamirr, line 146: Please add doc comments. It is not clear outright what the `String` type in `HashMap<String, DomainPattern>` does.

agent/provider/src/cli/whitelist.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl From<DomainPatternIds> for DomainPatterns {
+    fn from(val: DomainPatternIds) -> Self {
+        let patterns = val.pattern_ids.into_values().collect();
+        DomainPatterns { patterns }
+    }
+}
+
+impl DomainPatternIds {
+    fn remove(&mut self, ids: Vec<String>) -> (Vec<DomainPattern>, Vec<String>) {
```
kamirr, line 168: I do not know what this function returns without reading its body, please add a doc comment or define a proper structure for the return type instead of a tuple.

agent/provider/src/cli/whitelist.rs
```diff
@@ -0,0 +1,200 @@
...
+            if let Some(pattern) = self.pattern_ids.remove(&id) {
+                removed.push(pattern);
+            } else {
+                skipped.push(id);
+            }
+        }
+        (removed, skipped)
+    }
+
+    fn add(&mut self, add: Add) -> (Vec<DomainPattern>, Vec<DomainPattern>) {
```
kamirr, line 181: Same as `DomainPatternIds::remove`.

agent/provider/src/provider_agent.rs
```diff
@@ -66,6 +67,56 @@ impl GlobalsManager {
     }
 }
 
+struct WhitelistManager {
+    state: DomainWhitelistState,
+    monitor: Option<FileMonitor>,
+}
```
kamirr, line 73: Please add a short comment documenting what this structure does.

agent/provider/src/provider_agent.rs
```diff
@@ -66,6 +67,56 @@ impl GlobalsManager {
...
+
+impl WhitelistManager {
+    fn try_new(whitelist_file: &Path) -> anyhow::Result<Self> {
+        let mut patterns = DomainPatterns::load_or_create(whitelist_file)?;
+        patterns.update_and_save(whitelist_file, patterns.clone())?; //TODO fix ownership issue
```
kamirr, line 78: Why are you updating the `patterns` variable with itself and saving it instead of just doing `patterns.save()?`?

utils/manifest-utils/src/matching/domain.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl TryFrom<&PathBuf> for DomainPatterns {
+    type Error = anyhow::Error;
+
+    fn try_from(path: &PathBuf) -> Result<Self, anyhow::Error> {
+        let file = File::open(path)?;
+        let reader = BufReader::new(file);
+        Ok(serde_json::from_reader(reader)?)
+    }
+}
```
kamirr, line 54: I don't think this matches the semantics of `TryFrom`. (Try)From encode (fallible) type conversions, where you intend to keep the same data but in a different form, like `2i32` -> `2.0f32`. It also seems to partly duplicate the logic from `DomainPatterns::load`, is there a reason for that?

utils/manifest-utils/src/matching/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::ArgMatch;
+
+trait MatchPattern {
+    fn value(&self) -> String;
+    fn match_type(&self) -> ArgMatch;
+}
+
+pub trait Matcher: Debug + Send + Sync {
+    fn matches(&self, txt: &str) -> bool;
+}
```
kamirr, line 16: Please document what these traits do and their methods.

utils/manifest-utils/src/util.rs
```diff
@@ -323,6 +319,14 @@ pub enum KeystoreRemoveResult {
     Removed { removed: Vec<X509> },
 }
 
+/// Calculates Md5 of `txt` and returns first 8 characters.
+pub fn str_to_short_hash<STR: AsRef<[u8]>>(txt: STR) -> String {
```
kamirr, line 323: nitpick: considering that you don't need this type to be named, you could just do `pub fn str_to_short_hash(txt: impl AsRef<[u8]>) -> String`.

kamirr (changes requested):

agent/provider/src/cli/whitelist.rs
pwalski: Thanks! I made my life easier by testing json output only, so I missed it.

agent/provider/src/cli/keystore.rs
pwalski: To some degree. I wanted to use `println_conditional` for sake of consistency - without it these `println`s are suddenly special, but I will change it.
I am not exactly proud with the way it ended up - I have been tweaking printed output after implementing it to make it more clear for the user  (adding these `skipped` lists everywhere), so it is dotted with conditional statements and tuples you rightly complained on later.
Done.

agent/provider/src/cli/whitelist.rs
pwalski: Will change this println. A consistency was the reason for using it.
Second `if` was a shameful mistake (again on `skipped`..).
Done.

agent/provider/src/provider_agent.rs
pwalski: Adding type named "manager" always feels like admitting to a design/naming effort defeat (in my defense there are other "managers" nearby it). 
Will add a description.
Done.

utils/manifest-utils/src/matching/mod.rs
pwalski: The first one had even no implementation. Got removed. Weird that `clippy` had not reported it.
Second. Srsly. `Matcher` `matches` `txt`. Code cannot be more self descriptive than this.
I think I did not named it "Pattern", because I felt like it indicates some regex or glob, and matcher felt more general. Maybe I should.
Done first, thinking on second.

utils/manifest-utils/src/util.rs
pwalski: I will change it, but I used to avoid anonymous generic parameters in public functions because someone may want to use it for different types and they will need to change it back into a named generic.
Done.

agent/provider/src/cli/whitelist.rs
pwalski: Agree. I do not see any naming convention for maps in this project.
I would even make it a `type alias` if I would not need to implement `From` for it. It could also be a newtype idiom, but I hate seeing these `.0` everywhere.
Done.

agent/provider/src/cli/whitelist.rs
pwalski: I really hoped you will not notice it - it was this late addition to the CLI output, and I had no patience for naming new types.
Changed it to some generic type shared between `fn remove` and `fn add`. I am aware field `skipped` gets ignored in this function, but I am implementing this CLI it without any requirements, so it is not unlikely someone will request it later. Had not done it because IMO it was unnecessary in case of `remove`.
Done.

agent/provider/src/cli/whitelist.rs
pwalski: As in previous one, only in this `skipped` does not get skipped.
Done.

utils/manifest-utils/src/matching/mod.rs
kamirr: Right, I was more concerned about `MatchPattern`, if it's removed and there's nothing useful to say about `Matcher`, then it's fine by me.

agent/provider/src/provider_agent.rs
pwalski: I guess I needed it at some point, but not anymore. I runned tests and have checked feature manually after removing it and works.
Done.

utils/manifest-utils/src/matching/domain.rs
pwalski: At some point I started using `From`/`TryFrom` as a default trait for a constructor.
I was not even aware it is in `core::convert`. Thanks for this one. I removed some other case of incorrect `impl TryFrom`, but probably there are others.
Done.

agent/provider/src/cli/whitelist.rs
pwalski: I completely forgot about 'help' messages, but to be fair we do not have many of them in other commands.
Suitable messages have been added to both `whitelist` and 'keystore' CLI commands.
Done.

Issue #2183, opened 2022-08-25
When using locked default key, yagna daemon cannot start


MrDarthShoe:
Yagna version: 0.11.0-rc7

When trying to use **locked** default key, yagna daemon cannot start:

```
[2022-08-25T12:18:27.682+0200 INFO  ya_net::hybrid::service] Hybrid NET - Using default identity as network id: 0x487d6f663a0c7e2ee65edc48d9f2e0cfc9a8cd42
[2022-08-25T12:18:27.683+0200 INFO  ya_net::hybrid::service] Starting network (hybrid) with identity: 0x487d6f663a0c7e2ee65edc48d9f2e0cfc9a8cd42
[2022-08-25T12:18:27.711+0200 DEBUG ya_utils_networking::resolver] Resolved address: yacn2a.dev.golem.network:7477
[2022-08-25T12:18:27.711+0200 INFO  ya_net::hybrid::service] Hybrid NET relay server configured on url: udp://yacn2a.dev.golem.network:7477
[2022-08-25T12:18:27.736+0200 INFO  ya_service_bus::connection] stopped connection to gsb
Error: key locked```

PR #2187, opened 2022-08-26
2183 locked account locks daemon

MrDarthShoe:
resolves: https://github.com/golemfactory/yagna/issues/2183

### DIFF
--- a/core/identity/src/dao/identity.rs
+++ b/core/identity/src/dao/identity.rs
@@ -64,14 +64,6 @@ impl<'c> IdentityDao<'c> {
         .await
     }
 
-    pub async fn get_default_identity(&self) -> Result<Identity> {
-        use crate::db::schema::identity::dsl::*;
-        readonly_transaction(self.pool, |conn| {
-            Ok(identity.filter(is_default.eq(true)).first(conn)?)
-        })
-        .await
-    }
-
     pub async fn init_preconfigured(&self, preconfigured_identity: Identity) -> Result<Identity> {
         use crate::db::schema::identity::dsl as id_dsl;
         self.with_transaction(move |conn| {
--- a/core/identity/src/dao/identity.rs
+++ b/core/identity/src/dao/identity.rs
@@ -64,6 +64,14 @@ impl<'c> IdentityDao<'c> {
         .await
     }
 
+    pub async fn get_default_identity(&self) -> Result<Identity> {
+        use crate::db::schema::identity::dsl::*;
+        readonly_transaction(self.pool, |conn| {
+            Ok(identity.filter(is_default.eq(true)).first(conn)?)
+        })
+        .await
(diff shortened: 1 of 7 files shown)


core/identity/src/service/identity.rs
```diff
@@ -504,3 +526,86 @@ impl IdentityService {
...
     }
 }
+
+pub async fn wait_for_default_account_unlock(db: &DbExecutor) -> anyhow::Result<()> {
+    let identity_key = get_default_identity_key(&db).await?;
+
+    if identity_key.is_locked() {
+        log::info!("{}", yansi::Color::RGB(0xFF, 0xA5, 0x00).paint(
+            "Daemon cannot start because default account is locked. Unlock it by running 'yagna id unlock'"
+        ));
```
nieznanysprawiciel, line 536: I would move this, after subscribed endpoint is already bound

core/identity/src/service/identity.rs
```diff
@@ -504,3 +526,86 @@ impl IdentityService {
...
+                            tx_clone.send(()).await.expect("Receiver is closed");
+                        }
+                    }
+                };
+                Ok(())
+            }
+        });
+        subscribe(endpoint.clone()).await?;
+
+        wait_for_unlock(rx).await?;
```
nieznanysprawiciel, line 559: We should check `get_default_identity_key` for the second time here.
User could unlock his account, after the first check but before you subscribed to events and were able to capture event

core/model/src/identity.rs
```diff
@@ -20,6 +20,8 @@ pub enum Error {
     InternalErr(String),
     #[error("bad keystore format: {0}")]
     BadKeyStoreFormat(String),
+    #[error("bad request: {0}")]
+    BadRequest(String),
```
nieznanysprawiciel, line 24: I would create less general error, like InvalidPassword

core/identity/src/dao/identity.rs
```diff
@@ -64,6 +64,14 @@ impl<'c> IdentityDao<'c> {
...
     }
 
+    pub async fn get_default_identity(&self) -> Result<Identity> {
+        use crate::db::schema::identity::dsl::*;
+        readonly_transaction(self.pool, |conn| {
+            Ok(identity.filter(is_default.eq(true)).first(conn)?)
+        })
+        .await
+    }
+
```
nieznanysprawiciel, line 74: You don't have to implement this by yourself, because there is `self.lock().await.get_default_id()` on IdentityService

MrDarthShoe: Yes and no, because `get_default_id()` is a method made on IdentityService and this wait function is not part of it (and I think that it shouldn't)

nieznanysprawiciel (approved):

Issue #2138, opened 2022-08-03
Implement `cargo clippy`


wkargul:
Why:
- get rid off common known mistakes and improve Rust code,
- implement [clippy](https://github.com/rust-lang/rust-clippy)
- many lints appear when running clippy on current `yagna` (i.e. https://github.com/golemfactory/yagna/pull/2136) what helps catching bugs

What:
- run `cargo clippy` and solve issues in accordance with clippy hints (be carefull if fixing by `cargo clippy --fix`)
- adjust CI - run with every PR
- research on possible issues after implementing clippy

Notice:
- this should released no sooner than before `yagna-v0.11` release as this must be tested profoundly

PR #2238, opened 2022-09-27
Clippy style lints on workspace

MrDarthShoe:
Partially resolves https://github.com/golemfactory/yagna/issues/2138

`cargo clippy --jobs 4 --all-targets --workspace -- -A clippy::all -W clippy::style`


agent/provider/src/execution/registry.rs
```diff
@@ -146,16 +146,18 @@ pub struct ExeUnitsRegistry {
...
-    pub fn new() -> ExeUnitsRegistry {
-        ExeUnitsRegistry {
-            descriptors: HashMap::new(),
+impl Default for ExeUnitsRegistry {
+    fn default() -> Self {
+        Self {
+            descriptors: Default::default(),
         }
     }
+}
```
MrDarthShoe, line 155: I've removed `new` functions in some places but they can be left and added only `Default`. I am not sure if we want to have both or just `default()`

mfranciszkiewicz (approved):

Issue #2224, opened 2022-09-21
Verify certificate expiration works


mateuszsrebrny:
What:
- [ ] automated test: generate shortlived cert and verify expiration

pwalski:
Test scenario:
1. generate keypair
2. generate 1st x.509 certificate with short "expiration date"
3. generate x.509 CSR (certificate sign request)
4. generate 2nd certificate using CSR
5. load 1st certificate into keystore
6. wait until 1st keystore expires
7. verify 2dn certificate

PR #2240, opened 2022-09-29
Certificate expiration tests

MrDarthShoe:
resolves https://github.com/golemfactory/yagna/issues/2224

### DIFF
--- a/utils/manifest-utils/tests/certificate_expiration.rs
+++ b/utils/manifest-utils/tests/certificate_expiration.rs
@@ -20,6 +20,7 @@ fn accept_not_expired_certificate() {
 
     let (self_signed_cert, ca_key_pair) =
         create_self_signed_certificate(valid_from, valid_to).unwrap();
+
     write_cert_to_file(
         &self_signed_cert,
         &test_cert_dir.path().join("self_signed.pem"),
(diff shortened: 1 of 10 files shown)


wkargul (approved):
LGTM due to pair programming

wkargul (approved):

utils/manifest-utils/tests/certificate_expiration.rs
```diff
@@ -0,0 +1,284 @@
...
+    let sig = sign_data(data.as_bytes(), &csr_key_pair, &sig_alg);
+
+    assert!(sut
+        .verify_signature(
+            base64::encode(signed_cert.to_pem().unwrap()),
+            base64::encode(sig),
+            sig_alg,
+            data
+        )
+        .is_err());
```
pwalski, line 80: The error message could contain messages like "Invalid certificate" "Unknown signature algorithm: {}" and "Invalid signature". 
Maybe we should expect "Invalid certificate" here?

pwalski (changes requested):

utils/manifest-utils/tests/certificate_expiration.rs
wkargul: We have considered this profoundly and it seems like such approach with focus on error appearance might be more reliable. What in situation when error message changes? What other advantage can you see when focused on exact error message?

pwalski: You may end up with false positive in case certificate validation will work but only signature validation will fail.
To be fair if this Error would be an Enum type there would be no excuse, but I did not needed it to control the Error handling flow - I only needed a different message, so it is a String.
The only importance of this message is the fact it will appear in an Offer Rejection, which is a kind of important feedback for the Requestor, because it tells him whether theirs certificate is invalid, or they provided wrong algorithm, or the signature does not work, which is why I tested these messages in other tests (so the fact you do not have there is not a huge deal).

Also it is more of a review of my own code, but probably we should have some proper Error hierarchy where an Error reaching user in one form or another should be of a different type, but we do not have such hierarchy, and I had not introduced it either...

pwalski (approved):

PR #2255, opened 2022-10-12
github action to add and verify DoD with PRs

mateuszsrebrny:
:D look below \o/


MrDarthShoe (approved):

mateuszsrebrny:
@MrDarthShoe you did approve but didn't check the checkboxes... please do :D 
I shouldn't be the one checking them out :P

wkargul:
I double-reviewed it and it looks good to me. I also checked the checkboxes and it's awesome

mateuszsrebrny:
hm... they don't look checked to me :/ interesting

wkargul:
> hm... they don't look checked to me :/ interesting

I unchecked it just to see what is happening then :P All works correctly. The only thing is about possible extra commit arrival. What is happening then? I updated my comment above with some more description.

mateuszsrebrny:
> 

good catch :heart:
I don't see an easy way around that though :/

maybe we could agree that the reviewer doesn't checkout the checkboxes if their comments recommend doing changes so big that the testability topic should be revisited?

wkargul:
> > 
> 
> good catch ❤️ I don't see an easy way around that though :/
> 
> maybe we could agree that the reviewer doesn't checkout the checkboxes if their comments recommend doing changes so big that the testability topic should be revisited?
> 
> ... and will check them out during the next round of review

Yep, that's fair! 👍

PR #2257, opened 2022-10-12
UDP network endpoint support for runtimes

mfranciszkiewicz:


exe-unit/src/bin.rs
```diff
@@ -315,6 +315,10 @@ async fn run() -> anyhow::Result<()> {
 
 #[actix_rt::main]
 async fn main() {
+    std::panic::set_hook(Box::new(|e| {
+        log::error!("ExeUnit Supervisor panic: {e}");
+    }));
```
kamirr, line 320: Does this mean the binary won't print backtrace even with `RUST_BACKTRACE` set? This would limit the usefulness of logs supplied by users.

Nevertheless, this change deserves a comment with some explaination.

kamirr (changes requested):

exe-unit/src/bin.rs
mfranciszkiewicz: On the contrary - panics were not present in logs at all, just printed to stderr, which we would miss while retrieving logs from users.

I've modified the code to also execute the default panic hook

kamirr: ack

kamirr (approved):

PR #2280, opened 2022-11-06
Fix CI failing since rustc 1.65.0

kamirr:
Three issues.

1. Market test suite
BT format changed in Rust 1.65, causing `bactrace_util::generate_backtraced_name` to return an autogenerated name created via uuidv4. However, changes applied since then have actually made the happy path a hard requirement by expecting the resultant string to be of the form `(.*)::(.*)::.*`. This was the case when the test names were pulled from the backtrace, but didn't work for autogenerated uuidv4 ones.
This patch removed the uuid name generation from `backtrace_util`, and implemented it in `mock_node` in a different manner.

2. `split-debuginfo=unpacked` fails on Windows.
Cargo should not forward this option if not supported, but this does not work in this case. Likely a regression in cargo-rustc interaction. For now removed this flag with an appropriate remark to get the CI going. My fix in `cargo` is awaiting review (https://github.com/rust-lang/cargo/pull/11347), the option should be reenabled when it's merged and released.

3. clippy
We passed `-D warnings` to clippy, which (of course) made it fail on all warnings. Clippy 0.1.65 found more issues, and so it failed, bringing most of CI with it. This PR removes `-D warnings`.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -15,7 +15,7 @@ jobs:
     name: Unit Tests
     env:
       # `-D warnings` means any warnings emitted will cause build to fail
-      RUSTFLAGS: "-D warnings -C opt-level=z -C target-cpu=x86-64 -C debuginfo=1"
+      RUSTFLAGS: "-C opt-level=z -C target-cpu=x86-64 -C debuginfo=1"
       X86_64_PC_WINDOWS_MSVC_OPENSSL_DIR: c:/vcpkg/installed/x64-windows
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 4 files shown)


.github/workflows/unit-test.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features --workspace -- -D warnings
+          args: --all-targets --all-features --workspace
```
MrDarthShoe, line 51: This is workaround IMO and doesn't solve root cause

kamirr: As discussed, this is followed up in https://github.com/golemfactory/yagna/issues/2281.

nieznanysprawiciel (approved):

Issue #2270, opened 2022-10-25
erc20: transactions stalling due to insufficient gas / invalid nonce


mfranciszkiewicz:
Payment processing by the erc20 driver may stall when gas prices in the network reach an unusually high level. 

1. Sent transactions are not processed by miners due to low profit and kept in the memory pool instead. These pending transactions, which are never present on chain, are never finalized but marked as "sent" (**not** confirmed)  in the database. The daemon ought to mark such transactions for "resending" with a higher gas price and the same nonce.

2. Next transaction nonce is based on pending (memory pool) transactions. If any of the previous txs is in the state as mentioned in 1., the new transaction will not be finalized due to an invalid nonce.

This issue was _partially_ addressed in a draft PR https://github.com/golemfactory/yagna/pull/1789

mateuszsrebrny:
Testability discussion:
- [ ] can we make ganache to have high gas prices?
   - [ ] if we can we could test the payments separately
- [ ] if ganache is not good for this, we could mock the blockchain calls
- [ ] it might be easier to mock only a function that decides the gas price

etam:
@scx1332 is working on a new implementation of payments

evik42:
Testing:
- we would need a proxy to intercept some requests to the blockchain and return custom values (for example 1 for gas price) to get into this situation

Problems that I found:
- if max_gas_price is not defined when sending in the transaction then there is no limit for the initial gas price, but when reading the transaction back from the database, the `NULL` is converted to `0` and that blocks retrying the transaction with bumped gas price (0 is treated as invalid value there)
- same log lines are produced on multiple branches, it is not feasible to figure out from logs what is happening in the logic

.. will extend still investigating how it behaves when max_gas_price is not null

evik42:
Closing this issue. After checking with @scx1332 it is clear that the new implementation will use different method for transaction costs. (Using type 2 transactions.)

PR #2287, opened 2022-11-07
Fix max gas price handling when bumping gas price for resend

evik42:
Fixes some of the problems leading to #2270 


evik42:
I tested these changes locally as I had to intercept some blockchain calls and return specific values to get into the situation described in the ticket.

evik42:
Discussed this change and will not merge before the release. Will revisit later when we can automate tests for this scenario.

evik42:
Closing the PR as the expected rewrite of the payment driver will not contain this logic.

PR #2297, opened 2022-11-09
Fix binding broadcasts handlers

nieznanysprawiciel:


core/market/src/protocol/discovery.rs
```diff
@@ -54,6 +54,9 @@ pub struct DiscoveryImpl {
     offer_unsubscribe_handler: HandlerSlot<UnsubscribedOffersBcast>,
 
     config: DiscoveryConfig,
+    /// We need this to determine, if we use hybrid NET. Should be removed together
+    /// with central NET implementation in future.
+    net_config: net::Config,
```
evik42, line 59: If it is only used for checking if net type is hybrid, maybe adding only the bool value to the struct would be enough?

evik42 (changes requested):

PR #2304, opened 2022-11-11
Add `activity/{id}/agreement` endpoint

kamirr:
Returns an agreement corresponding to the given activity.

### DIFF
--- a/core/activity/src/api.rs
+++ b/core/activity/src/api.rs
@@ -36,6 +36,7 @@ mod common {
         scope
             // .service(get_activities_web)
             .service(get_events)
+            .service(get_activity_agreement_web)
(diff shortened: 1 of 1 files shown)


evik42 (approved):

johny-b:
Hmm, I've a feeling that `activity/{id}/agreement` is quite adhoc - do we have any other endpoint like this?

E.g. when I want to check `agreement_id` for a given invoice, I send `GET /invoices/{invoiceId}` and get an object with `agreement_id` key. Why not `GET /activities/{activityId}`?

(Just to make it clear, this endpoint is OFC also fine for me, but I think we should care about the API consistency)

PR #2313, opened 2022-11-15
Add VPN timestamping

kamirr:
To test run yagna as follows:
```bash
RUST_LOG="packet-trace=trace" cargo run --features=packet-trace-enable service run
```

And run a task utilizing VPN, e.g. ssh.

### DIFF
--- a/core/net/src/hybrid/service.rs
+++ b/core/net/src/hybrid/service.rs
@@ -22,6 +22,7 @@ use ya_core_model::net::local::{
     BindBroadcastError, BroadcastMessage, NewNeighbour, SendBroadcastMessage, SendBroadcastStub,
 };
 use ya_core_model::{identity, net, NodeId};
+use ya_packet_trace::packet_trace_maybe;
 use ya_relay_client::codec::forward::{PrefixedSink, PrefixedStream, SinkKind};
 use ya_relay_client::crypto::CryptoProvider;
 use ya_relay_client::proto::Payload;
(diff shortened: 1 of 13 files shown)


core/net/src/hybrid/service.rs
```diff
@@ -474,6 +477,10 @@ fn forward_bus_to_net(
     let state = state.clone();
     let request_id = gen_id().to_string();
 
+    packet_trace_maybe!("net::forward_bus_to_net", {
+        try_extract_from_ip_frame(msg)
+    });
+
```
MrDarthShoe, line 483: I would use qualified importshere like so:
```rust
ya_packet_trace::packet_trace_maybe!("net::forward_bus_to_net", {
    try_extract_from_ip_frame(msg)
});
```
It would enable to drop:
```rust
#[allow(unused_imports)]
use ya_packet_trace::{packet_trace_maybe, try_extract_from_ip_frame};
```
in every file

core/vpn/src/network.rs
```diff
@@ -577,6 +590,11 @@ impl Handler<Egress> for Vpn {
     fn handle(&mut self, msg: Egress, _: &mut Self::Context) -> Self::Result {
         let frame = msg.event.payload.into_vec();
 
+        #[cfg(feature = "packet-trace-enable")]
```
MrDarthShoe, line 593: If you must use `#[cfg(feature = "packet-trace-enable")]` anyway, so why `packet_trace_maybe` has `enable` feature inside? IMO it should be either of these, not combined, because they are redundant and causing unnecessary chaos.
My proposals:
1. `enable` in `ya_packet_trace`: make crate in such way that it will be unnecessary to use redundant `#[cfg(feature`
2. Delete `enable` in `ya_packet_trace`: and write `#[cfg(feature` when it is used in our code

kamirr: After some consideration, I disagree. I cannot modify the `ya_packet_trace` in a way which will make this uses redundant, because this is related to the lifetime of data that we're tracing in any way other than providing a `ya_packet_trace::copy` for making a conditional copy, which has way more cognitive load than `#[cfg(..)]`.

Option two is no good, because I'd be adding about 20 additional `#[cfg(..)]` just because of the 3 you're concerned about.

I assumed the naming would show that the feature-gated copies (`packet-trace-enable`) and macro invocations (`packet_trace[_maybe]!`) are tightly coupled, but to be sure I've added a clarifying comment to the copies, as they're a kind of an edge case.

core/net/src/hybrid/service.rs
kamirr: Done

MrDarthShoe (approved):
I didn't check if places where we have timestamping are sufficient enough.
Code looks good besides redundant cfg features

PR #2317, opened 2022-11-16
Add timeout for NET shutdown

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna-triage/issues/235

This is rather workaround than solution, but users won't have to kill yagna process


wkargul (approved):
👍

PR #2319, opened 2022-11-17
enforce broadcast propagation

wkargul:

### DIFF
--- a/core/market/src/matcher.rs
+++ b/core/market/src/matcher.rs
@@ -25,8 +25,11 @@ pub(crate) mod store;
 use crate::db::dao::{DemandDao, DemandState};
 use error::{MatcherError, MatcherInitError, QueryOfferError, QueryOffersError};
 use futures::FutureExt;
+use log::{debug, info};
 use resolver::Resolver;
 use store::SubscriptionStore;
+use ya_core_model::net::local::{BroadcastMessage, NewNeighbour, SendBroadcastMessage};
+use ya_net::bind_broadcast_with_caller;
 
 /// Stores proposal generated from resolver.
 #[derive(Debug)]
@@ -109,9 +112,31 @@ impl Matcher {
         tokio::task::spawn_local(cyclic::bcast_offers(self.clone()));
         tokio::task::spawn_local(cyclic::bcast_unsubscribes(self.clone()));
 
+        self.bind_neighbourhood_bcast(local_prefix).await.ok();
+
         self.bind_expiration_tracker()
             .await
             .map_err(|e| MatcherInitError::ExpirationTrackerError(e.to_string()))?;
+
+        Ok(())
+    }
+
+    async fn bind_neighbourhood_bcast(&self, local_prefix: &str) -> anyhow::Result<()> {
+        let bcast_address = format!("{local_prefix}/{}", NewNeighbour::TOPIC);
+        let myself = self.clone();
+        bind_broadcast_with_caller(
+            &bcast_address,
+            move |_caller, _msg: SendBroadcastMessage<NewNeighbour>| {
+                let myself = myself.clone();
+                async move {
+                    info!("Received new neighbour broadcast");
+                    cyclic::bcast_offers_once(myself.clone()).await;
+                    return Ok(());
+                }
+            },
+        )
+        .await?;
+
         Ok(())
     }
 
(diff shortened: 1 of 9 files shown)


core/market/src/matcher.rs
```diff
@@ -109,9 +112,31 @@ impl Matcher {
...
+
+    async fn bind_neighbourhood_bcast(&self, local_prefix: &str) -> anyhow::Result<()> {
+        let bcast_address = format!("{local_prefix}/{}", NewNeighbour::TOPIC);
+        let myself = self.clone();
+        bind_broadcast_with_caller(
+            &bcast_address,
+            move |_caller, _msg: SendBroadcastMessage<NewNeighbour>| {
+                let myself = myself.clone();
+                async move {
+                    info!("Received new neighbour broadcast");
```
nieznanysprawiciel, line 132: I would log caller, to know from which Node this broadcast came. And probably we could lower log level now, since it is not so important.

core/net/src/hybrid/service.rs
```diff
@@ -1116,6 +1121,38 @@ fn gen_id() -> u64 {
...
+    Ok(())
+}
+
+pub async fn send_bcast_new_neighbour() {
+    let node_id = crate::service::identities().await.unwrap().0;
+
+    let net_type = { *NET_TYPE.read().unwrap() };
+    match net_type {
+        NetType::Hybrid => {
+            log::info!("Broadcasting new neighbour");
```
nieznanysprawiciel, line 1147: Probably debug level

core/net/src/hybrid/service.rs
```diff
@@ -1116,6 +1121,38 @@ fn gen_id() -> u64 {
...
+            if let Err(e) = broadcast(node_id, NewNeighbour {}).await {
+                log::error!("Error broadcasting new neighbour: {:?}", e);
```
nieznanysprawiciel, line 1149: Could use `Display` instead of `Debug`?

core/serv/src/main.rs
```diff
@@ -553,6 +553,8 @@ impl ServiceCommand {
                 let api_host_port = rest_api_host_port(api_url.clone());
                 let rest_address = api_host_port.clone();
 
+                ya_net::hybrid::send_bcast_new_neighbour().await;
```
nieznanysprawiciel, line 556: After rethinking: this could cause, that we need to wait a few seconds (it's a timeout for finding Nodes on relay) before we start yagna.
We should spawn it.
This is the only important comment in this review

nieznanysprawiciel (commented):
There is only one important thing here, rest you can ignore

core/serv/src/main.rs
wkargul: resolved, thank you

nieznanysprawiciel (dismissed):

core/net/src/hybrid/service.rs
```diff
@@ -1116,6 +1121,38 @@ fn gen_id() -> u64 {
...
+            if let Ok(client) = Net::client().await {
+                client.invalidate_neighbourhood_cache().await.ok();
+            }
+
+            return Ok(());
+        },
+    )
+    .await?;
+
+    Ok(())
```
evik42, line 1138: This async block already returns a `Result<(), BindBroadcastError>` so we just need to map the error into `anyhow::Error`.
```suggestion
    .await
    .map_err(Into::into)
```
Alternatively just return the `BindBroadcastError` and let the caller transform it if needed:
```
async fn bind_neighbourhood_bcast() -> Result<(), BindBroadcastError> {
...
    .await
}
````

File: core/net/src/hybrid/service.rs [hunk 6]
```diff
@@ -1116,6 +1121,38 @@ fn gen_id() -> u64 {
...
+async fn bind_neighbourhood_bcast() -> anyhow::Result<()> {
+    let bcast_address = format!("{}/{}", net::local::BUS_ID, NewNeighbour::TOPIC);
+    bind_broadcast_with_caller(
+        &bcast_address,
+        move |_caller, _msg: SendBroadcastMessage<NewNeighbour>| async move {
+            if let Ok(client) = Net::client().await {
+                client.invalidate_neighbourhood_cache().await.ok();
+            }
+
+            return Ok(());
```
evik42, line 1133: I think this would be more idiomatic
```suggestion
            Ok(())
```

evik42 (dismissed):

evik42 (approved):

Issue #2294, opened 2022-11-09
Fix toolchain version


MrDarthShoe:
To avoid sudden CI errors after new rust toolchain release, we want to have fixed version on out CI.
Then if needed, we can bump whole toolchain and fix possible issues.

- [ ] Fix versions on CI workflows
- [ ] Include supported version on readme

kamirr:
Besides `ya-client`, I've also pinned `ya-relay` to 1.64.0 in https://github.com/golemfactory/ya-relay/pull/220 (CI broke).

PR #2324, opened 2022-11-18
Fix toolchain version to 1.64

MrDarthShoe:
Partially resolves https://github.com/golemfactory/yagna/issues/2294

### DIFF
--- a/utils/fd-metrics/src/lib.rs
+++ b/utils/fd-metrics/src/lib.rs
@@ -69,6 +69,7 @@ pub fn fd_metrics() -> HashMap<FdType, usize> {
 /// Maps fd to FdType
 ///
 /// Uses fstat on unix, always returns None on Windows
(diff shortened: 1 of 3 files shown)


utils/fd-metrics/src/lib.rs
```diff
@@ -69,6 +69,7 @@ pub fn fd_metrics() -> HashMap<FdType, usize> {
 /// Maps fd to FdType
 ///
 /// Uses fstat on unix, always returns None on Windows
+#[allow(unused_variables)]
```
kamirr, line 72: `#[cfg_attr(target_os = "windows", unused_variables)]` would be more precise, but I'm not sure if it's not too much clutter. I have no strong preference, so feel free to merge if you want to keep it that way.

kamirr (approved):

PR #2335, opened 2022-11-25
Fix poetry version for nightly goth build

evik42:
Poetry version was missed in https://github.com/golemfactory/yagna/commit/37539868b0aa52f9df0a133b4aa5467a309c2a7f
And nightly consistently fails since that change got merged so I think this will fix the build.


kamirr (approved):
lgtm

PR #2382, opened 2022-12-16
Allow setting everyone all mode + UI

### LINKED ISSUE #2371: Provider can set "Outbound for all" Rule on or off
**As a** Provider
**I want** to be able to turn Outbound for all Rule on or off
**so that** either only trusted Outbound connections are made or more tasks will be received by me.

MrDarthShoe:
Resolves #2371


utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct RuleStore {
+    pub path: PathBuf,
+    config: Arc<RwLock<RulesConfig>>,
+}
+
+impl RuleStore {
+    pub fn load_or_create(rules_file: &Path) -> Result<Self> {
+        if rules_file.exists() {
+            log::debug!("Loading rule from: {:?}", rules_file);
```
nieznanysprawiciel, line 21: You can use `.display()` on `Path` and `PathBuf`
I would be rather against using debug prints in code if not necessary, because they can be ugly

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+            let file = OpenOptions::new().read(true).open(rules_file)?;
+
+            Ok(Self {
+                config: Arc::new(serde_json::from_reader(BufReader::new(file))?),
```
nieznanysprawiciel, line 25: Here is potential problem, when file is empty, but exists, this will fail.

nieznanysprawiciel: But this isn't the only place when we have this problem

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+                path: rules_file.to_path_buf(),
+            };
+            store.save()?;
+
+            Ok(store)
+        }
+    }
+
+    fn save(&self) -> Result<()> {
+        log::debug!("Saving RuleStore to: {:?}", self.path);
```
nieznanysprawiciel, line 43: `.display()`

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+        Ok(std::fs::write(
+            &self.path,
+            serde_json::to_string_pretty(&*self.config.read().unwrap())?,
+        )?)
+    }
+
+    pub fn reload(&self) -> Result<()> {
+        log::debug!("Reloading RuleStore from: {:?}", self.path);
```
nieznanysprawiciel, line 51: `.display()`

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn replace(&self, other: Self) {
+        let store = std::mem::take(&mut (*other.config.write().unwrap()));
+
+        *self.config.write().unwrap() = store;
+    }
+
+    pub fn set_everyone_mode(&self, mode: Mode) -> Result<()> {
+        log::debug!("Setting outbound everyone mode: {:?}", mode);
```
nieznanysprawiciel, line 66: Could you implement `Display` trait?

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.config.write().unwrap().outbound.everyone = mode;
+
+        self.save()
```
nieznanysprawiciel, line 69: If you acquire lock, you should wrap it in braces.
Otherwise there is danger, that saving will be done under lock
```suggestion
        {
                self.config.write().unwrap().outbound.everyone = mode;
        }
        self.save()
```

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    pub fn set_default_audited_payload_mode(&self, mode: Mode) -> Result<()> {
+        log::debug!("Setting outbound audited_payload default mode: {:?}", mode);
```
nieznanysprawiciel, line 73: `Display` please

utils/manifest-utils/src/rules.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.config
+            .write()
+            .unwrap()
+            .outbound
+            .audited_payload
+            .default
+            .mode = mode;
+
+        self.save()
```
nieznanysprawiciel, line 82: Braces around lock

agent/provider/src/market/negotiator/builtin/manifest.rs
```diff
@@ -27,7 +28,7 @@ impl NegotiatorComponent for ManifestSignature {
         demand: &ProposalView,
         offer: ProposalView,
     ) -> anyhow::Result<NegotiationResult> {
-        if self.enabled.not() {
+        if self.enabled.not() || self.rulestore.always_accept_outbound() {
```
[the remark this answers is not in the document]
MrDarthShoe: No, this `self.enabled` is additional feature which can be set once at the start of  from cmd-line arg `--policy-disable-component manifest_signature_validation`.

utils/manifest-utils/src/rules.rs
MrDarthShoe: I don't think so: in these cases when there is `self.save()`, previous LockGuard is dropped immediately (as it is not bound to longer lived value like in `utils/manifest_utils/src/policy.rs` line 224

utils/manifest-utils/src/rules.rs
MrDarthShoe: That's right. We should answer this question:
What if user deliberately makes invalid config or deletes all contents of a file?
Should we emit an error or maybe replace this config with default one?
IMO we shouldn't replace it as user can have very sophisticated config, he can edit it manually and when editing he can make a mistake and suddenly all of his work is deleted

utils/manifest-utils/src/rules.rs
nieznanysprawiciel: Yes, probably. But we had to many problems with locks that are held longer than we think and these problems are not easy to debug, so to be sure we always limit the scope of the lock explicite

nieznanysprawiciel (approved):

Issue #2396, opened 2023-01-04
Implement Everyone + Audited payload Rules negotiator logic


MrDarthShoe:
1. Provider is able to disable outbound for "Everyone"
2. Provider is able to switch outbound for "Everyone" to whitelist
3. Provider is able to disable outbound for "Audited Payload"
4. Provider is able to switch outbound for "Audited Payload" to whitelist
5. Provider is able to accept all outbound for "Audited Payload"

"Everyone" Rule overrides more restrictive Rules.

PR #2402, opened 2023-01-05
Negotiations: everyone + audited-payload

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2396


Cargo.toml
```diff
@@ -52,7 +52,7 @@ ya-utils-networking = "0.2"
 ya-fd-metrics = { path = "utils/fd-metrics" }
 ya-version = "0.2"
 ya-vpn = "0.2"
-ya-client = "0.7.1"
```
etam, line 55: downgrade?

MrDarthShoe: `ya-client` is actually in 0.7.0 right now https://github.com/golemfactory/ya-client/tags

MrDarthShoe: And actually it should be 0.7 :)

wkargul (approved):
LGTM

PR #2403, opened 2023-01-05
Yagna test framework

nieznanysprawiciel:


tests/core/identity/test_appkey.rs
```diff
@@ -0,0 +1,51 @@
+use ya_test_framework::framework::macros::{prepare_test_dir, serial_test};
+use ya_test_framework::framework::{framework_test, YagnaFramework};
+use ya_test_framework::utils::YagnaCli;
+
+#[cfg_attr(not(feature = "framework-test"), ignore)]
+#[framework_test]
```
MrDarthShoe, line 6: Suggestion: I would rename it to be more specific. `framework_test` seems too generic.
Proposals:
`yagna_test`
`ya_test`
Or to include "layer" of tests:
`e2e_test`
`system_test`

test-utils/test-framework/src/yagna.rs
```diff
@@ -0,0 +1,226 @@
...
+                    if chunk == b"READY=1" {
+                        return Ok(());
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[cfg(not(unix))]
```
MrDarthShoe, line 211: Do we want to support non-unix systems?

test-utils/test-framework/src/framework.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use crate::YagnaMock;
+
+pub use macros::prepare_test_dir;
+pub use ya_framework_macro::framework_test;
+
+#[derive(Clone)]
+pub struct YagnaFramework {
+    inner: Arc<Mutex<YagnaNetworkImpl>>,
+    test_dir: PathBuf,
```
MrDarthShoe, line 15: You could use `tempdir::TempDir` for that. It will allow to run tests in parallel- every testdir will have it's own directory which won't collide

tests/core/identity/test_appkey.rs
nieznanysprawiciel: I agree. We could split tests in multiple test sets, because execution will be rather long.
Maybe let's make survey on priv channels

test-utils/test-framework/src/framework.rs
nieznanysprawiciel: 1. This directory contains yagna datadirs. In case of integration test, there is no chance that tests are self-explanatory. You are running whole yagna. It is not easy to debug it, without having datadir. Logs are included in too.
2. Running in parallel will be theoretically possible, because test name is rather unique. Maybe some name mangling will be necessary. But `framework_test` macro is using `serial_test` to avoid parallel execution. Integrations tests are rather too heavy to run in parallel

MrDarthShoe: Ad 1 & Ad 2:
I don't agree that if test is integration- than it should automatically be complicated and it will test multiple things at once.
When tests are stable, if you must debug some complicated test it means for me that there is lack of other-simpler case test. Because if simpler test passes- fail is because of functionality which is added by this more complex one. If simpler fails- it means that it doesn't matter that complex one is failing as problem lays in simple case.

Of course when tests are very slow, it may be difficult to run many small-case tests. But then we come to second point that we can run them in parallel. I think there is some misconception here:
We design framework where tests can't be ran in parallel -> so running many tests becomes slow -> so we naturally will dive into making big & complicated tests and not many simpler ones -> we will end up in tests which will be hard to understand and maintain. When something will break it will require many time to debug

nieznanysprawiciel: Tests don't have to be complicated. Running yagna makes it complicated in itself.
Running tests in parallel isn't solution for long running tests, because it only can speed up as many times as many cores you have. And running to many yagnas in the background can harm their performance and we can end up with flaky tests like in goth (which we will end up anyway, but we can try to minimize it)

test-utils/test-framework/src/framework.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+pub fn prepare_test_dir_(dir: &str) -> PathBuf {
+    let test_dir = PathBuf::from(&dir);
+
+    #[allow(unused_must_use)]
+    {
+        std::fs::remove_dir_all(&test_dir); // ignores error if does not exist
+        std::fs::create_dir_all(&test_dir);
+    }
```
MrDarthShoe, line 47: [nitpick / personal taste]
```suggestion
    let _ = std::fs::remove_dir_all(&test_dir);
    let _ = std::fs::create_dir_all(&test_dir);
```

test-utils/test-framework/src/framework.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub use prepare_test_dir;
+    pub use serial_test;
+}
+
+pub fn prepare_test_dir_(dir: &str) -> PathBuf {
+    let test_dir = PathBuf::from(&dir);
+
+    #[allow(unused_must_use)]
+    {
+        std::fs::remove_dir_all(&test_dir); // ignores error if does not exist
```
MrDarthShoe, line 45: [Question]
1. Will it remove dirs also after some tests failed? If yes, than it will become impossible to debug such test as there won't be any logs from yagna.
2. I wonder do we want to reuse existing test directories or make another ones with folder of different timestamp (as in Goth or other e2e testing frameworks)?

I believe reusing / removing directories after each run is enough for UT or IT but for e2e ones it will hurt us sooner rather than later.

test-utils/test-framework/src/framework.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub fn new(test_dir: impl Into<PathBuf>, test_name: impl Into<String>) -> YagnaFramework {
+        YagnaFramework {
+            inner: Arc::new(Mutex::new(YagnaNetworkImpl { nodes: vec![] })),
+            test_dir: test_dir.into(),
+            test_name: test_name.into(),
+        }
+    }
+
+    pub fn new_node(&self, name: impl ToString) -> YagnaMock {
+        let yagna_dir = self.test_dir.join(&self.test_name).join(name.to_string());
```
MrDarthShoe, line 82: Should't `test_dir` already include `test_name` in it?

test-utils/test-framework/src/utils/yagna_cli.rs
```diff
@@ -0,0 +1,23 @@
+use anyhow::anyhow;
+
+use crate::YagnaMock;
+
+pub trait YagnaCli {
+    fn appkey_list_json(&self) -> anyhow::Result<Vec<serde_json::Value>>;
+}
```
MrDarthShoe, line 7: What is a purpose of this trait?

test-utils/test-framework/src/yagna.rs
```diff
@@ -0,0 +1,232 @@
...
+            if process.terminate(timeout).await.is_err() {
+                process.kill();
+            }
+        }
+        Ok(())
+    }
+}
+
+#[cfg(unix)]
+mod tracker {
```
MrDarthShoe, line 134: Consider using existing method for this: https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.kill_on_drop

MrDarthShoe (commented):
It would be good to have  roadmap  of features & insctuctions how to use it & how it roughly works in `test-utils/test-framework/README.md`

test-utils/test-framework/src/utils/yagna_cli.rs
nieznanysprawiciel: Helper functions to provide common functionalities which will be used in many tests

test-utils/test-framework/src/framework.rs
nieznanysprawiciel: Directories will remain after test, but will be removed on next run.
Someone running tests can consciously choose to save directory by just changing it's name.

Do you think it's enough?

test-utils/test-framework/src/framework.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::sync::{Arc, Mutex};
+
+use crate::YagnaMock;
+
+pub use macros::prepare_test_dir;
+pub use ya_framework_macro::framework_test;
+
+#[derive(Clone)]
+pub struct YagnaFramework {
+    inner: Arc<Mutex<YagnaNetworkImpl>>,
```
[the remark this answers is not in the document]
nieznanysprawiciel: `inner` is convention used in many places. I don't think we need longer name here, since it is obvious from context

test-utils/test-framework/framework-macro/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    if function.sig.inputs.len() != 1 {
+        panic!(
+            "Framework test should have only one argument, but {} given.",
+            function.sig.inputs.len()
+        );
+    }
+
+    if let Some(_param) = function.sig.inputs.first() {
+        // TODO: Argument should be `YagnaFramework`
```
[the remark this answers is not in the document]
nieznanysprawiciel: Macro expects function with `YagnaFramework` argument
I will change to make more clear

test-utils/test-framework/src/framework.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+pub mod macros {
+    #[macro_export]
+    macro_rules! prepare_test_dir {
+        () => {
+            // CARGO_TARGET_TMPDIR is available in compile time only in binary modules, so we can't
+            // use it in this library. Thanks to using macro it will be resolved in final code not here
+            // and it will work.
+            ya_test_framework::framework::prepare_test_dir_(env!("CARGO_TARGET_TMPDIR"))
```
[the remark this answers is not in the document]
nieznanysprawiciel: Tests are compiled into binaries, so it should work.
I don't want `temp_dir`, exactly for this reason that it creates directory in `/tmp`. If you are using any developer environment, `CARGO_TARGET_TMPDIR` will be inside your project and `/tmp` requires additional steps to find it.
Goth is good example how cumbersome it can be

.github/workflows/system-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Install Rust ${{ env.rust_stable }}
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ env.rust_stable }}
+          components: rustfmt, clippy
```
MrDarthShoe, line 32: unneeded rustfmt & clippy. I don;t know however if it will speed up things or not

tests/readme.md
```diff
@@ -0,0 +1,45 @@
+# Testing framework based tests
+
+Test framework should be use, when test requires running yagna daemon in the background.
```
MrDarthShoe, line 3:
```suggestion
Test framework should be used, when test requires running yagna daemon in the background.
```

MrDarthShoe (approved):

.github/workflows/system-test.yml
nieznanysprawiciel: I forgot about this, maybe I will fix this at some point later

PR #2407, opened 2023-01-09
Easy backend access - skip authorization. Return autoconfigured key when no key match. Only dev option

scx1332:


nieznanysprawiciel (changes requested):
We discussed this inside team and decided to reject this PR.

- It isn't good practice to change yagna usage between production and developer environment
- This may be potential vulnerability, that allows to exploit yagna and omit authorization
- Maybe this problem could be solved by some kind of script, without modifying yagna code

PR #2421, opened 2023-01-17
Proposal: remove double checking CI after merge to master / release

MrDarthShoe:
image

### DIFF
--- a/.github/workflows/binaries-aarch64.yml
+++ b/.github/workflows/binaries-aarch64.yml
@@ -1,10 +1,6 @@
 name: Build binaries (aarch64)
 
 on:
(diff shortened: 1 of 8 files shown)


evik42:
After CI builds run correctly on the PR, master branch can have new commits so it needs to be checked after merge.

MrDarthShoe:
> After CI builds run correctly on the PR, master branch can have new commits before the actual merge happens. It needs to be checked after merge.

When new commits arrive to master, PR should be out-of-date (which will disable merge button) and CI should be ran on PR again with new commits on master

evik42 (approved):
I agree

PR #2422, opened 2023-01-23
Outbound 2.0: Partner rule configuration

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2401

### DIFF
--- a/agent/provider/src/cli/rule.rs
+++ b/agent/provider/src/cli/rule.rs
@@ -31,7 +31,13 @@ pub enum SetOutboundRule {
     },
     AuditedPayload {
         #[structopt(long)]
-        certificate: Option<String>,
+        cert_id: Option<String>,
+        #[structopt(subcommand)]
+        mode: Mode,
+    },
+    Partner {
+        #[structopt(long)]
+        cert_id: String,
         #[structopt(subcommand)]
         mode: Mode,
     },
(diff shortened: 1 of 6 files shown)


agent/provider/src/cli/rule.rs
```diff
@@ -147,7 +159,11 @@ impl From<RulesManager> for RulesTable {
 
         table.with_header(outbound.enabled);
         table.add_everyone(outbound.everyone);
-        table.add(outbound.audited_payload.default);
+        table.add_audited_payload(outbound.audited_payload.default);
+
+        for (cert_id, rule) in outbound.partner.iter() {
+            table.add_partner(cert_id, rule);
+        }
```
wkargul, line 166: I'd use the same convention as between lines 160-162 -> 
`table.add_partner(outbound.partner);`

and the iterate by (cert_id, rule) under `add_partner` which should be sth like

```
    fn add_partner(&mut self, partner: HashMap<String, CertRule>) {
        let _ = partner.iter().map(|(cert_id, rule)| {
            let row = serde_json::json! {[ "Partner", rule.mode, cert_id, rule.description ]};

            self.table.values.push(row);
        });
    }

wkargul (changes requested):

wkargul (approved):
LGTM

Issue #2423, opened 2022-08-25
Make contents of presets.json stable


etam:
Why:
As DevOps I'd like to be able to generate this file from a template. When deploying updates I want to see if the file actually changed. Right now I get false positives, because values are in different order and it makes the file differ.

What:
When ya-provider writes down presets.json, the the order of some objects is unpredictable:
- the list under `.["presets"]` (I don't know why it changes order)
- keys under `.["presets"][]["usage-coeffs"]` (it's unpredictable, because it's serialized from HashMap)

Those can be resolved either by changing data structures (for example HashMap to BTreeMap) or by implementing a custom serializer, which sorts those fields.

PR #2424, opened 2023-01-24
Stabilize contents of presets.json

jiivan:
fixes: https://github.com/golemfactory/yagna/issues/2423

### DIFF
--- a/agent/provider/src/market/presets.rs
+++ b/agent/provider/src/market/presets.rs
@@ -21,8 +21,7 @@ pub struct Preset {
     pub exeunit_name: String,
     pub pricing_model: String,
     pub initial_price: f64,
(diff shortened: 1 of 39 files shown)


agent/provider/src/market/presets.rs
```diff
@@ -21,7 +21,9 @@ pub struct Preset {
     pub exeunit_name: String,
     pub pricing_model: String,
     pub initial_price: f64,
-    pub usage_coeffs: HashMap<String, f64>,
+    // It's important that all values are sorted, so that ansible can easily detect significant
+    // changes. In case of doubt, please ask DevOps.
```
etam, line 25: also here
```suggestion
    // It's important that all values are sorted, so that other tools can easily detect changes.
```

nieznanysprawiciel (approved):

Issue #2419, opened 2023-01-16
Strict whitelist matching works strange


MrDarthShoe:
When implementing Outbound v2.0, I found interesting case:

When setting `strict` whitelist i.e. `domain.com`, `hackered-domain.com` also matches this whitelist.

PR #2425, opened 2023-01-25
Fix manifest negotiator test whitelists. Change default matching to more restrictive option.

evik42:
Resolves: #2419 

The whitelist was malformed in the tests.
The default matching should be the most restrictive option.

### DIFF
--- a/agent/provider/tests/manifest_negotiator.rs
+++ b/agent/provider/tests/manifest_negotiator.rs
@@ -37,29 +37,13 @@ fn manifest_negotiator_test_accepted_because_outbound_is_not_requested() {
     let whitelist = r#"{ "patterns": [] }"#;
     let rulestore = r#"{"outbound": {"enabled": false, "everyone": "none", "audited-payload": {"default": {"mode": "none", "description": ""}}}}"#;
 
-    // signature does not matter here
-    let signature = Signature {
-        private_key_file: None,
-        algorithm: None,
-        certificate: None,
-    };
     let comp_manifest_b64 = create_comp_manifest_b64(urls);
-    let signature_b64 = signature.private_key_file.map(|signing_key| {
-        MANIFEST_TEST_RESOURCES.sign_data(comp_manifest_b64.as_bytes(), signing_key)
-    });
-    let cert_b64 = signature.certificate.map(cert_file_to_cert_b64);
 
-    manifest_negotiator_test_encoded_manifest_sign_and_cert_and_cert_dir_files(
+    manifest_negotiator_test_encoded_manifest_without_signature(
         rulestore,
         whitelist,
         comp_manifest_b64,
-        signature_b64,
-        signature.algorithm,
-        cert_b64,
         None,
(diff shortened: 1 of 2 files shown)


agent/provider/tests/manifest_negotiator.rs
```diff
@@ -121,11 +105,35 @@ fn manifest_negotiator_test_accepted_because_of_no_payload() {
...
+    r#"{"outbound": {"enabled": true, "everyone": "whitelist", "audited-payload": {"default": {"mode": "none", "description": ""}}}}"#, // rulestore config
+    r#"["https://hacked-well-formed-regex.com.mydomain.com"]"#,
+    None; // error msg
+    "Rejected because everyone whitelist NOT matched - regex matching"
+)]
+#[test_case(
+    r#"{"outbound": {"enabled": true, "everyone": "whitelist", "audited-payload": {"default": {"mode": "none", "description": ""}}}}"#, // rulestore config
+    r#"["https://valid.well-formed-regex.com"]"#,
+    None; // error msg
+    "Accepted because everyone whitelist matched - regex matching subdomain"
```
MrDarthShoe, line 136: Whitelist-specific cases should be extended in `manifest_negotiator_test_whitelist` test (as it is created to test all whitelist cases), I would keep `manifest_negotiator_test_manifest_with_urls` focused on negotiations logic with simple whitelisting, and not complicated cases.

evik42: I noticed that the same tests are duplicated multiple times (for example everyone -> whitelist, audited payload -> none is verified for matching and not matching multiple times).
I added these here because there was already a whitelist did not match test in this bunch, did not want to review the whole test file as part of this PR.

agent/provider/tests/manifest_negotiator.rs
```diff
@@ -121,11 +105,35 @@ fn manifest_negotiator_test_accepted_because_of_no_payload() {
...
+    r#"{"outbound": {"enabled": true, "everyone": "whitelist", "audited-payload": {"default": {"mode": "none", "description": ""}}}}"#, // rulestore config
+    r#"["https://hacked-regex.com.mydomain.com"]"#,
+    None; // error msg
+    "Accepted because everyone whitelist matched - regex matching"
+)]
+#[test_case(
+    r#"{"outbound": {"enabled": true, "everyone": "whitelist", "audited-payload": {"default": {"mode": "none", "description": ""}}}}"#, // rulestore config
+    r#"["https://hacked-well-formed-regex.com.mydomain.com"]"#,
+    None; // error msg
+    "Rejected because everyone whitelist NOT matched - regex matching"
```
MrDarthShoe, line 130: Rejected but error is expected to None

MrDarthShoe (changes requested):

MrDarthShoe:
Addresses  https://github.com/golemfactory/yagna/issues/2419

MrDarthShoe (approved):

PR #2439, opened 2023-01-30
Fix pretty table crash by bumping version

scx1332:


scx1332:
If someone can, please check if golemsp is still working as intended (I don't know what kind of tables it should have)

scx1332:
Library fix_pretty_table contained unsafe code that was crashing. It is fixed in 0.10 version so I've updated to that one.

nieznanysprawiciel (approved):

Issue #2442, opened 2023-02-01
Partner negotiations


MrDarthShoe:
Negotiations with mocked certificate & permissions validation.

PR #2451, opened 2023-02-03
Partner negotiations

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2442


agent/provider/src/rules.rs
```diff
@@ -199,105 +202,168 @@ impl RulesManager {
...
+                .read()
+                .unwrap()
+                .outbound
+                .partner
+                .get(cert_id)
+            {
+                return self
+                    .check_mode(&rule.mode, manifest)
+                    .map_err(|e| anyhow!("Partner {e}"));
             }
```
nieznanysprawiciel, line 291: This is incorrect. Not all certificates provided by Requestor need to have permissions. Only the leaf certificate,
which directly signed `node.json`.

You can have for example Golem Root, which you trust only with whitelist, but other Golem certificates signed by root, which you trust more for some reason

agent/provider/src/rules.rs
```diff
@@ -199,105 +202,168 @@ impl RulesManager {
...
+        }
+        Err(anyhow!(
+            "Partner rule whole chain of cert_ids is not trusted: {:?}",
```
nieznanysprawiciel, line 294: In general you don't trust leaf certificate, not whole chain

agent/provider/tests/manifest_negotiator.rs
```diff
@@ -218,6 +218,149 @@ fn manifest_negotiator_test_with_valid_payload_signature(
...
+    Some("outbound-urls|https://domain.com"),
+    None;
+    "Accepted as requested domain is in the permitted ones in cert"
+)]
+#[test_case(
+    r#""outbound": { "mode": "all", "description": ""}"#,
+    r#"["https://domain.com"]"#, // compManifest.net.inet.out.urls
+    Some("outbound"),
+    None;
+    "Accepted as permission is outobound unrestricted"
```
nieznanysprawiciel, line 240:
```suggestion
    "Accepted as permission is outbound unrestricted"
```

utils/manifest-utils/src/golem_certificate.rs
```diff
@@ -0,0 +1,96 @@
+use url::Url;
+use ya_client_model::NodeId;
+
+pub struct GolemCertificate {
+    pub node_id: NodeId,
```
nieznanysprawiciel, line 5: We have separate structure for `node.json` it is not the same as certificate

nieznanysprawiciel (commented):
Please look at the comments. I think most of the things can be done in other PRs

agent/provider/src/rules.rs
MrDarthShoe: It is correct as discussed

utils/manifest-utils/src/golem_certificate.rs
MrDarthShoe: I am not sure what was intention of naming it that way. Maybe @evik42 could you clarify?

From negotiations perspective: `GolemCertificate` + `some other data` = `node.json`
We get `node.json` from demand, then we pass it to `verify_golem_certificate` function, and we get `GolemCertificate`.

utils/manifest-utils/src/golem_certificate.rs
```diff
@@ -0,0 +1,96 @@
...
+    Expired(CertificateId),
+    InvalidSignature(CertificateId),
+    PermissionsDoNotMatch(CertificateId), // the signer does not have all the required permissions
+    UrlParseError(Vec<String>),
+}
+
+pub enum GolemPermission {
+    All,
+    ManifestOutboundUnrestricted,
+    ManifestOutbound(Vec<Url>),
```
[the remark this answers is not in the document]
MrDarthShoe: 1. Manifest removed
2. I would wait for negotiations to be done to make such decisioons

agent/provider/src/rules.rs
```diff
@@ -199,105 +202,168 @@ impl RulesManager {
...
+    ) -> Result<()> {
+        let node_identity =
+            node_identity.ok_or_else(|| anyhow!("Partner rule requires partner certificate"))?;
+
+        let verified_cert = self
+            .keystore
+            .verify_golem_certificate(&node_identity)
+            .map_err(|e| anyhow!("Partner {e}"))?;
+
+        if requestor_id != verified_cert.node_id {
```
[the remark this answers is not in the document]
MrDarthShoe: It is confusion because GolemCertificate here actually means node.json (discussion in different comment)

PR #2455, opened 2023-02-07
cert-permissions: backward compatibility

kamirr:


agent/provider/src/cli/default-cert-permissions.json
```diff
@@ -0,0 +1,6 @@
+{
+  "4c8d2924": [
+    "outbound-manifest",
+    "unverified-permissions-chain"
+  ]
+}
```
kamirr, line 6: Is this right?

pwalski: I do not know how much time we want to dedicate to it, but there was already a request to change certificate `Id`s. In such case this thing would break silently on next release, and it will manifest itself only by certificates having no permissions.
It could be tested using certificate with id `4c8d2924` (Golem CA? https://ca.golem.network/cert/GOLEM-CA.pem) and this `json` file in same dir https://github.com/golemfactory/yagna/blob/master/utils/manifest-utils/test-utils/src/lib.rs#L35

pwalski (approved):

PR #2469, opened 2023-02-20
X509 certificate related code in decdicated module

### LINKED ISSUE #2465: Keystore can handle Golem certificates
User should be able to add / delete / list golem certificates alongside x509 ones.

Remember to remove mocked id's in Keystore: https://github.com/golemfactory/yagna/blob/master/utils/manifest-utils/src/policy.rs#L239

Remember to switch golem certificate validation lib to new one 
At the end it should replace this mock crate: https://github.com/golemfactory/yagna/blob/master/utils/manifest-utils/src/golem_certificate.rs

pwalski:
Resolves #2465


utils/manifest-utils/src/keystore/mod.rs
```diff
@@ -0,0 +1,301 @@
+pub mod golem_keystore;
```
MrDarthShoe, line 1: `keystore.rs` with `keystore` folder will be more readable

utils/manifest-utils/src/keystore/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+
+impl Cert {
+    pub fn id(&self) -> String {
+        match self {
+            Cert::Golem { id, cert: _ } => id.into(),
+            Cert::X509(cert) => cert.id.to_string(),
+        }
+    }
+
+    pub fn not_after(&self) -> String {
```
MrDarthShoe, line 33: Why it is needed?

utils/manifest-utils/src/keystore/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+impl CommonAddParams for AddParams {
+    fn certs(&self) -> &Vec<PathBuf> {
+        &self.certs
+    }
+}
+
+impl X509AddParams for AddParams {
+    fn permissions(&self) -> &Vec<crate::policy::CertPermissions> {
+        &self.permissions
+    }
```
MrDarthShoe, line 81: Unneeded

MrDarthShoe (approved):
enough for POC

utils/manifest-utils/src/keystore/mod.rs
pwalski: Originally it was X509 expiration date. Our certificates will not have expiration date?
It has been assumed it is an useful info to display (even for us, because DevOps issues certs for 3 months only ;))

utils/manifest-utils/src/keystore/mod.rs
pwalski: Removed it.

utils/manifest-utils/src/keystore/mod.rs
pwalski: Moved it to `keystore.rs`, but initially place it on purpose in `mod.rs`. IDEs tree view do not display `keystore` dir next to `keystore.rs`, which for me is less readable.

PR #2477, opened 2023-02-28
Use certs lib

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2496


utils/manifest-utils/src/keystore/golem_keystore.rs
```diff
@@ -58,14 +59,13 @@ impl KeystoreBuilder<GolemKeystore> for GolemKeystoreBuilder {
...
+fn read_cert(cert_path: &PathBuf) -> anyhow::Result<(String, Certificate)> {
     let mut cert_file = File::open(cert_path)?;
     let mut cert_content = String::new();
     cert_file.read_to_string(&mut cert_content)?;
     let cert_content = cert_content.trim();
-    let cert = golem_certificate::verify_golem_certificate(cert_content)?;
-    let id = str_to_short_hash(cert_content);
-    Ok((id, cert))
+    let cert = golem_certificate::validator::validate_golem_certificate(cert_content)?;
+    Ok((cert.chain[0].clone(), cert.cert))
```
pwalski, line 68: Not important, but in X509 chain cert_0 signs cert_1.
https://knowledge.digicert.com/solution/SO16297.html
_The chain or path begins with the SSL/TLS certificate, and each certificate in the chain is signed by the entity identified by the next certificate in the chain._

pwalski: I assume `chain[0]` is leaf certificate.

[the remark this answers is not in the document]
MrDarthShoe: Won't do for the sake of "convention" only. It is necessary in x509 as it's cert chain behaves like singly linked list so there is no way to check previous one in the chain. In our case we can as signer cert is included in the signed cert itself.

pwalski (approved):

Issue #2436, opened 2023-01-26
Update goth runners


etam:
- [x] #2431 
- [x] add "goth" label to new runners
- [ ] update integration tests for ubuntu-22.04 compatibility https://github.com/golemfactory/yagna/issues/2456
- [ ] destroy old runners

PR #2482, opened 2023-03-02
Shared keystore for X509 and Golem certificates

### LINKED ISSUE #2465: Keystore can handle Golem certificates

pwalski:
Resolves #2465


pwalski:
Keystore supports adding, removing (implemented it yesterday), and listing of both X509 and Golem certificates.
Hash/id of Golem cert for now was done using cert content (so it need to be relpaced with dedicaed function).
Adding/removing/listing of certificates could be tested by adding some Golem certs (when these will become available) into https://github.com/golemfactory/yagna/blob/8b8524d30aa09c8f2ea0e0937bdc2e06d91757f1/utils/manifest-utils/test-utils/resources/test/certificates.tar and writing some scenarios here https://github.com/golemfactory/yagna/blob/8b8524d30aa09c8f2ea0e0937bdc2e06d91757f1/agent/provider/tests/manifest_negotiator.rs

utils/manifest-utils/src/lib.rs
```diff
@@ -1,11 +1,12 @@
...
 pub mod matching;
 pub mod policy;
 pub mod util;
 
 pub use manifest::*;
-pub use policy::{Keystore, Policy, PolicyConfig};
-pub use util::{
-    decode_data, DecodingError, KeystoreLoadResult, KeystoreManager, KeystoreRemoveResult,
-};
+// pub use keystore::
```
MrDarthShoe, line 9: Leftover

agent/provider/tests/manifest_negotiator.rs
```diff
@@ -455,20 +456,6 @@ fn manifest_negotiator_test_with_invalid_payload_signature(
...
-    Some("Audited-Payload rule: Not sufficient permissions. Required: `outbound-manifest|unverified-permissions-chain`, but has only: `outbound-manifest`"); // error msg
-    "Manifest rejected, because certificate has no `unverified-permissions-chain` permission."
-)]
-#[test_case(
-    Signature { private_key_file: Some("foo_req.key.pem"), algorithm: Some("sha256"), certificate: Some("foo_req.cert.pem")},
-    Some("NYI"), // cert_permissions_b64
-    &vec![CertPermissions::All],
-    Some("Audited-Payload rule: Not sufficient permissions. Required: `outbound-manifest|unverified-permissions-chain`, but has only: `all`"); // error msg
-    "Manifest rejected, even when permissions are set to `All` because `unverified-permissions-chain` permission is also required when Demand comes with its permissions"
-)]
```
MrDarthShoe, line 471: Why they are being deleted?

agent/provider/src/rules.rs
```diff
@@ -371,13 +368,9 @@ impl RulesManager {
...
     }
 
-    fn verify_permissions(&self, cert: &str, demand_permissions_present: bool) -> Result<()> {
-        let mut required = vec![CertPermissions::OutboundManifest];
-
-        if demand_permissions_present {
-            required.push(CertPermissions::UnverifiedPermissionsChain);
-        }
-
+    //TODO delete when deleting X509 permissions feature
```
MrDarthShoe, line 371: This comment is misleading, because ultimately we will have two use-cases for audited-payload rule:
1. Incoming is x509 cert: in this case we don't check permissions as there are no permissions in cert
2. Incoming is golem cert: in this case we need to check cert permissions for it

agent/provider/src/rules.rs
```diff
@@ -371,13 +368,9 @@ impl RulesManager {
...
+    fn verify_permissions(&self, cert: &str) -> Result<()> {
```
MrDarthShoe, line 372: Why `demand_permisisons_present` was removed?

agent/provider/tests/keystore_cli.rs
```diff
@@ -180,7 +180,7 @@ fn test_keystore_add_certificate_permissions(
     command.arg("--json").assert().success();
 
     let result = list_certificates_command(&cert_dir).unwrap();
-
+    println!("Result: {result:?}");
```
MrDarthShoe, line 183: Please add task for outbound 2.0 about adding keystore CLI tests for golem certs

utils/manifest-utils/src/keystore.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    pub fn list_ids(&self) -> HashSet<String> {
+        self.list()
+            .into_iter()
+            .map(|cert| cert.id())
+            .collect::<HashSet<String>>()
+    }
+
+    pub fn verify_signature(
```
MrDarthShoe, line 195:
```suggestion
    pub fn verify_x509_signature(
```

I propose this change to expicit say what we are verifying

MrDarthShoe (changes requested):

agent/provider/tests/manifest_negotiator.rs
pwalski: `unverified-permissions-chain` is this "future proofing" permission which it is not used anywhere.

agent/provider/src/rules.rs
pwalski: Because there are no permissions arriving in demand - it was this permissions related demand property which was supposed to be used in future.

agent/provider/tests/manifest_negotiator.rs
pwalski: Additionally I deleted from these tests every single case of passing `None` as a certificate permissions arriving in demand.
Cert permissions as a demand property were a late addition to this functionality, so I sprinkled them everywhere in tests, and now I just removed them, because it would be hard to figure out what these permissions are even for as they are not used anywhere during negotiation.

agent/provider/tests/keystore_cli.rs
pwalski: Created task https://github.com/golemfactory/yagna/issues/2483

MrDarthShoe (dismissed):

pwalski:
Updated integration test runners to `ubuntu-22.04` and `Python-3.10.0`.
Integration tests were hanging in queue for hours. Searching for solution resulted in finding an advice to use latest `ubuntu`, and then I found task about updating goth runners (#2436).

MrDarthShoe (approved):

PR #2504, opened 2023-03-13
 Support for passing authToken to browser WebSocket connection

prekucki:
# Problem

Standard [WebSocket](https://websockets.spec.whatwg.org/) does not allow you to add your own headers to the http request. nor to pass on your own authentication.

The standard approaches are:

  - authorization is part of the internal protocol and is performed during the interaction
  - a token is added to the url address and authorization is performed based on it.

# Solution 

Every http request with `Upgrade: websocket` header can pass appKey i authToken query parameter instead of in the autohrization header.

# Test / Example

Short `deno` script:

```typescript
const ws = new WebSocket(`ws://127.0.0.1:7465/net-api/v1/net/0/tcp/127.0.0.1/22?authToken=test-key`);
ws.onopen = () => console.log('open');
ws.onerror = (err) => console.error('error', err.message);
ws.onmessage = (msg) => console.log('msg', msg);
```

### Before change

```sh
$ YAGNA_AUTOCONF_APPKEY=test-key yagna service run
....
$ deno run -A test-it.ts 
error NetworkError: failed to connect to WebSocket: HTTP error: 401 Unauthorized
```

### After change

```sh
$ YAGNA_AUTOCONF_APPKEY=test-key yagna service run
....
$ deno run -A test-it.ts 
error NetworkError: failed to connect to WebSocket: HTTP error: 404 Not Found
```

### DIFF
--- a/core/serv-api/web/src/middleware/auth/mod.rs
+++ b/core/serv-api/web/src/middleware/auth/mod.rs
@@ -8,9 +8,10 @@ pub use crate::middleware::auth::resolver::AppKeyCache;
 use actix_service::{Service, Transform};
 use actix_web::dev::{ServiceRequest, ServiceResponse};
 use actix_web::error::{Error, ErrorUnauthorized, ParseError};
-use actix_web::HttpMessage;
+use actix_web::{web, HttpMessage};
 use actix_web_httpauth::headers::authorization::{Bearer, Scheme};
 use futures::future::{ok, Future, Ready};
+use serde::Deserialize;
 use std::cell::RefCell;
 use std::pin::Pin;
 use std::rc::Rc;
@@ -65,9 +66,24 @@ where
     }
 
     fn call(&self, req: ServiceRequest) -> Self::Future {
+        #[derive(Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct QueryAuth {
+            auth_token: String,
+        }
+
         let header = parse_auth::<Bearer, _>(&req)
             .ok()
-            .map(|b| b.token().to_string());
+            .map(|b| b.token().to_string())
+            .or_else(|| {
+                if Some("websocket".as_bytes()) == req.headers().get("upgrade").map(AsRef::as_ref) {
+                    web::Query::<QueryAuth>::from_query(req.query_string())
+                        .ok()
+                        .map(|q| q.into_inner().auth_token)
+                } else {
+                    None
+                }
+            });
 
         let cache = self.cache.clone();
         let service = self.service.clone();


pwalski:
I believe tokens to access WS are usually generated on demand and have short expiration time
(like in this guide https://devcenter.heroku.com/articles/websocket-security or this API https://api.slack.com/methods/apps.connections.open).
Yagna API does not have sessions. Bound GSB services probably should be linked to some sort of internal GSB API session and get unbound the moment it expires. Introduction of tokens/tickets to access WS could be an elegant way to handle such sessions using API (and it would allow for appkey to stay in `Authorization` header only).

Usage scenarios:
1. Initial GSB API connection:
1.1 `POST /gsb-api/v1/services` with `Authorization` header returns `{ services_id, ticket, listen: { on, components} }` and `Location: /{services_id}?ticket={ticket}`
1.2 `GET /gsb-api/v1/services/{services_id}?ticket={ticket}` with `Connection: Upgrade` + `Upgrade: websocket` headers initializes WS connection.

2. Reconnecting to GSB API:
2.1 `GET /gsb-api/v1/services/{services_id}` with `Authorization` header returns `{ services_id, ticket, listen: { on, components }}` (or `404` if session has expired and GSB services got unbound).
2.2 `GET /gsb-api/v1/services/{services_id}?ticket={ticket}` with `Connection: Upgrade` + `Upgrade: websocket` headers initializes WS connection.

Step `2.1` seems redundant but it could allow app to assert if GSB services it expects are still bound. In case of using `token={appkey}` it could connect to WS and wait for a message that will never arrive. The problem this solution introduces is that `GET /gsb-api/v1/{services_id}` would need to support requests both with and without `Authorization` header, but it probably can be done.

There is also a problem with ownership of bound GSB services.
Currently authorization is used only to give access to a specific API endpoint, but there is nothing preventing identity/user A to delete/access services bound by identity B. Probably following use cases should be implemented:
1. `POST /gsb-api/v1/services` returns `400` with some specific message if given GSB address prefix is already used by B.
1. `DELETE /gsb-api/v1/services/{services_id}` as A returns `401` if services have been bound by B.
2.  `GET /gsb-api/v1/services/{services_id}` with `Authorization` as A returns `401` if services have been bound by B.

pwalski (approved):
It can be merged and I will enable authorization for `/gsb-api/v1/services/`, add `Identity` param to GSB API, and update `ya-client` to include `authToken` for `gsb-api/v1/services` in this task:
https://github.com/golemfactory/yagna/issues/2510

core/serv-api/web/src/middleware/auth/mod.rs
```diff
@@ -65,9 +66,24 @@ where
...
+            auth_token: String,
+        }
+
         let header = parse_auth::<Bearer, _>(&req)
             .ok()
-            .map(|b| b.token().to_string());
+            .map(|b| b.token().to_string())
+            .or_else(|| {
+                if Some("websocket".as_bytes()) == req.headers().get("upgrade").map(AsRef::as_ref) {
+                    web::Query::<QueryAuth>::from_query(req.query_string())
```
pwalski, line 80: It is not a huge concern, but query param `authToken` might be printed out in logs.
It is unusual to see `Authorization` header in logs.

PR #2512, opened 2023-03-16
Cert id length

### LINKED ISSUE #2435: Change `cert_id` to whole fingerprint

### LINKED ISSUE #2493: Use short version of fingerprints in CLI
Right now 
- X509 uses short hash both in code & CLI
- golem certificates use long fingerprints in code & CLI

Ultimately we want to have:
- long fingeprints in code
- shortened versions in CLI

This task is about using only short version of fingerprints in CLI

We should display shortest- non conflicting hash.
Minimum length of hash should be 8 characters.

```
In case of conflict, logic should be intelligent enough.
```

kamirr:
Closes #2435 
Closes #2493


agent/provider/src/cli/keystore.rs
```diff
@@ -131,44 +152,116 @@ fn remove(config: ProviderConfig, remove: Remove) -> anyhow::Result<()> {
...
     Ok(())
 }
 
-struct CertTable {
-    table: ResponseTable,
+fn find_ids_by_prefix(certs: &[Cert], prefix: &str) -> Vec<String> {
+    certs
+        .iter()
+        .map(|cert| cert.id())
+        .filter(|id| id.starts_with(prefix))
```
pwalski, line 168: We could build once a https://docs.rs/trie-rs/latest/trie_rs from `certs` and use it for prefix search, but it would be an overkill.

agent/provider/src/cli/keystore.rs
```diff
@@ -131,44 +152,116 @@ fn remove(config: ProviderConfig, remove: Remove) -> anyhow::Result<()> {
...
+                    break;
+                }
+            }
+        }
+
+        let mut values = Vec::new();
+        for (id_prefix, cert) in ids.into_iter().zip(self.entries.into_iter()) {
+            values.push(match cert {
+                Cert::X509(cert) => {
+                    serde_json::json! { [ id_prefix, cert.not_after, cert.subject, cert.permissions ] }
```
pwalski, line 236: Output of this table will probably change, but it is out of scope of this task.
I could not find relevant task. Maybe it was not clarified yet.

pwalski (dismissed):
Tests for changes introduced here will be done in this task
https://github.com/orgs/golemfactory/projects/30/views/2?pane=issue&itemId=21858005
before it will be done clarification of `keystore` cmd output is required.

agent/provider/src/cli/keystore.rs
kamirr: Agreed, I think it's better to leave it simple

pwalski (dismissed):

pwalski (approved):

agent/provider/src/rules.rs
```diff
@@ -90,30 +90,43 @@ impl RulesManager {
...
+        let cert_id = {
+            let certs: Vec<_> = self
+                .keystore
+                .list()
+                .into_iter()
+                .filter(|cert| cert.id().starts_with(&cert_id))
+                .collect();
+
+            if certs.is_empty() {
+                return Err(anyhow!(
```
MrDarthShoe, line 102: Consider `bail!`

agent/provider/src/cli/keystore.rs
```diff
@@ -116,7 +111,33 @@ fn add(config: ProviderConfig, add: Add) -> anyhow::Result<()> {
...
+        if full_ids.is_empty() {
+            ids.insert(remove_prefix.clone()); //won't match anyway
+        } else if full_ids.len() == 1 {
+            ids.insert(full_ids[0].clone());
+        } else {
+            println_conditional(
+                &config,
+                &format!(
+                    "Prefix '{remove_prefix}' isn't unique, consider using full certificate id"
+                ),
```
MrDarthShoe, line 129: Should't we have a following logic:
If prefix is not unique then extend prefixes?

agent/provider/tests/rule_cli.rs
```diff
@@ -172,8 +172,13 @@ fn rule_set_should_edit_certificate_rules(rule: &str, mode: &str) {
...
 
     let result = list_rules_command(data_dir.path());
+    let mode_actual = result["outbound"][rule]
+        .as_object()
+        .and_then(|obj| obj.iter().find(|(id, _cert)| id.starts_with(&cert_id)))
+        .map(|(_id, value)| &value["mode"]);
 
-    assert_eq!(&result["outbound"][rule][&cert_id]["mode"], mode);
+    assert!(mode_actual.is_some());
+    assert_eq!(mode_actual.unwrap(), mode);
```
MrDarthShoe, line 181: Two asserts here doesn't make sense. Second one will Oslo fail when `mode_actual` will be none so I propose either:
1. Leave only second one as is
2. Leave second one as so: `assert_eq!(mode_actual, mode);` and `unwrap` is before assert when creating `let mode_actual`

agent/provider/tests/rule_cli.rs
```diff
@@ -201,7 +206,13 @@ fn rule_set_with_import_cert_should_add_to_keystore_and_rulestore(rule: &str, mo
     let added_certs = list_certs(data_dir.path());
 
     for cert in added_certs {
-        assert_eq!(result["outbound"][rule][cert]["mode"], mode);
+        let mode_actual = result["outbound"][rule]
+            .as_object()
+            .and_then(|obj| obj.iter().find(|(id, _cert)| id.starts_with(&cert)))
+            .map(|(_id, value)| &value["mode"]);
+
+        assert!(mode_actual.is_some());
```
MrDarthShoe, line 214: dittos

agent/provider/tests/rule_cli.rs
```diff
@@ -172,8 +172,13 @@ fn rule_set_should_edit_certificate_rules(rule: &str, mode: &str) {
         .success();
 
     let result = list_rules_command(data_dir.path());
+    let mode_actual = result["outbound"][rule]
+        .as_object()
+        .and_then(|obj| obj.iter().find(|(id, _cert)| id.starts_with(&cert_id)))
+        .map(|(_id, value)| &value["mode"]);
```
MrDarthShoe, line 178: Maybe it could be extracted to func?

utils/manifest-utils/tests/certificate_store.rs
MrDarthShoe, line 1: Shouldn't we have shortened versions here?

pwalski (approved):

utils/manifest-utils/tests/certificate_store.rs
kamirr: No, ids are collected directly from the `HashSet` in the keystore, and these are the complete versions.

pwalski (approved):

MrDarthShoe (approved):

Issue #2438, opened 2023-01-30
Bump Rust to 1.68


kamirr:
In all Rust repositories we pin the version in.

golmek:
somewhere between 0.5 - 4h

PR #2530, opened 2023-04-05
Bump rust 1.68.2

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2438


agent/provider/src/hardware.rs
```diff
@@ -499,6 +499,7 @@ fn partition_space<P: AsRef<Path>>(path: P) -> Result<u64, Error> {
         use nix::sys::statvfs::statvfs;
         let stat =
             statvfs(path.as_os_str()).map_err(|e| sys_info::Error::General(e.to_string()))?;
+        #[allow(clippy::unnecessary_cast)]
```
kamirr, line 502: why allow instead of removing the cast?

kamirr (approved):

agent/provider/src/hardware.rs
MrDarthShoe: This is because on mac & linux these values differ :<

kamirr: cool

Issue #2456, opened 2023-02-07
try running goth on new runners


jiivan:

etam:
What to do:
- In files `.github/workflows/integration-test-*.yml` replace
  - runner label
    ```patch
    -runs-on: [goth, ubuntu-18.04]
    +runs-on: [goth, ubuntu-22.04]
    ```
  - python version
    ```patch
       - name: Configure Python
         uses: actions/setup-python@v2
         with:
    -      python-version: "3.8.0"
    +      python-version: "3.10.0"
    ```
  - everything else is unknown

golmek:
Error: 
![image](https://user-images.githubusercontent.com/126684244/230023409-44752970-c102-4782-b9b7-d3b5195535e7.png)

@Smokowski will check with SDK first

nieznanysprawiciel:
Since we weren't able to find the cause of this problem and we can't spend more time on this in this quarter, we will make a decision before planning Q3, how to proceed.

golmek:
Moving from In Progress -> Ready as it will be re-visited before Q3

jalas167:
comments and history of resolutino attempts are in https://github.com/golemfactory/yagna/pull/2534

evik42:
I am looking into this.

evik42:
My findings:
- Even after changing docker network from default to a named one to be able to disable ipv6, mitmproxy is not serving connections
- The routing addon in mitmproxy is not called, the connection is lost before that
- Tried to upgrade mitmproxy as v5.3 is 3 years old, unfortunately they do not support running mitmproxy embedded into a python application since 7.0 (latest version is 9.0.1)
  - This is causing problem with signal registration, I tried to monkey patch it but did not succeed
  - Tried to run mitmproxy in a separate process, it starts up but does not forward traffic
- `actions/setup-python@v4` can install Python 3.8 to Ubuntu 22.04, tried that also did not change a thing
  
 I am not a python expert so maybe I missed something but I was not able to make it work.

PR #2534, opened 2023-04-12
try running goth on new runners

jiivan:
fixes: #2456 
depends on: https://github.com/golemfactory/goth/pull/620


goth_tests/pyproject.toml
```diff
@@ -23,7 +23,7 @@ pytest-asyncio = "^0.20.2"
 pytest-rerunfailures = "^10.3"
 #goth = "^0.11"
 # to use development goth version uncomment below
-goth =  { git = "https://github.com/golemfactory/goth.git", rev = "a6e63b875fd4b254476eaafcc20657e5a99f7f73" }
+goth =  { git = "https://github.com/golemfactory/goth.git", rev = "master" }
```
pwalski, line 26: If you want to have yagna goth tests passing you need to use this branch
https://github.com/golemfactory/goth/pull/617
(it is not merged - needs verification (you can mark it as verified and merge it, but this branch should not be deleted immediately when doing a squash merge because `yagna` master uses it))

pwalski: Merged to `master`, but now it fails because `docker-compose-hybrid` has been removed.

nieznanysprawiciel:
Mitmproxy is not printing any logs, it probably didn't started working.
`nginx-proxy` timeouts when trying to connect to mitmproxy.

I suspect that it is not compatible with new version of python,

nieznanysprawiciel:
I'm unable to update mitmproxy:
https://github.com/golemfactory/goth/pull/620

I will go back to the start and verify if it is really the cause for runners problem

nieznanysprawiciel:
What is happening:
Tests are failing on waiting for Offer subscription from Providers.
Provider agent initialization is blocked on call for getting payment accounts, so we never reach Offer subscription.

Goth's architecture looks like this:
- There is docker-compose containing infrastructure (like ya-relay-server, nginx-proxy, mock blockchain etc)
- Provider agent and daemon are in the same container, but run independently from docker-compose
- Provider makes call to daemon but not directly
- There is `nginx-proxy` in docker-compose which catches the call and forwards to `mitmproxy`
- `mitmproxy` is host service outside of docker network, which monitors events between Nodes
- `mitmproxy` forwards requests to Provider daemon

`paymentAccounts` request is timeouting in `proxy-nginx` and is certainly not reaching `mitmproxy`, because `mitmproxy` logs are empty.
Provider initialization fails, because this request is necessary to start Provider.
Than waiting for subscription timeouts.

nieznanysprawiciel:
My attempts:
1. I was unable to reproduce problem locally. Everything is working there.
2. I tried to update `mitmproxy`, because I suspected that it stopped working, because of python version update. This attempt failed (https://github.com/golemfactory/goth/pull/620). Later @shadeofblue stated, that he didn't suspect, that `mitmproxy` is a culprit
3. Since `mitmproxy` is on localhost, I suspected, that maybe there is problem with connecting to localhost after docker version update. I tried using `host.docker.internal` (https://github.com/golemfactory/goth/pull/622). This attempt failed as well.
4. Docker [documentation](https://docs.docker.com/desktop/networking/) states about `host.docker.internal`: `This is for development purpose and does not work in a production environment outside of Docker Desktop`. Maybe 
docker handles host access differently then in previous versions and for example started to forbid using this feature. I tried to find what happens if I use it anyway, but couldn't find anything.

Issue #2445, opened 2023-02-01
Remove certificate permissions logic from provider


MrDarthShoe:
Rules are replacing existing "certificate permissions" logic from provider, so they are redundant.

We decided that we don't do backward compatibility (only warning on install if provider modified standard permissions).

Moved to https://github.com/golemfactory/yagna/issues/2566 
~~i.e. `keystore list` command returns `permissions` column:
We should replace "permissions" column with "rules" (i.e. `Outbound: audited-payload | partner`)~~

Testing:
- remove some existing tests which use permissions
- edit cli tests for `keystore list`

PR #2564, opened 2023-05-05
Remove permissions logic

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2445

### DIFF
--- a/utils/manifest-utils/src/keystore/x509_keystore.rs
+++ b/utils/manifest-utils/src/keystore/x509_keystore.rs
@@ -2,7 +2,6 @@ use super::{
     AddParams, AddResponse, Cert, CommonAddParams, Keystore, KeystoreBuilder, RemoveParams,
     RemoveResponse,
 };
-use crate::{policy::CertPermissions, util::format_permissions};
 use anyhow::{anyhow, bail};
 use chrono::{DateTime, Duration, NaiveDateTime, Utc};
 use openssl::{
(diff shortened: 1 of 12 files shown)


utils/manifest-utils/src/keystore/x509_keystore.rs
```diff
@@ -645,69 +506,6 @@ fn cert_subject_entries(cert: &X509Ref, nid: Nid) -> Option<String> {
...
-    pub fn get(&self, cert: &X509Ref) -> Vec<CertPermissions> {
-        let id = match cert_to_id(cert) {
-            Ok(id) => id,
-            Err(_) => return vec![],
-        };
-        self.permissions.get(&id).cloned().unwrap_or_default()
-    }
-
-    pub fn save(&self, path: &Path) -> anyhow::Result<()> {
-        let mut file = File::create(path.join(PERMISSIONS_FILE))?;
```
pwalski, line 706: `cert-permissions.json` file is no longer created so we no longer need `ya-provider pre-install` logic applying permissions default values https://github.com/golemfactory/yagna/blob/remove-permissions-logic/agent/provider/src/cli/pre_install.rs#L6

pwalski: Also maybe we should delete this file using `ya-installer` command?
https://github.com/golemfactory/ya-installer/blob/master/commands.json

pwalski (approved):

Issue #2520, opened 2023-03-30
Specific rejection message when ya-provider's keystore is empty


pwalski:
Problem:
Demands containing signed Manifest are rejected with message `'failed to verify manifest signature: Invalid certificate'` when `ya-provider`'s keystore is empty. 

Todo:
There should be some specific message indicating case when certificate cannot be verified because `ya-provider`'s keystore is empty.

Why:
It will make diagnosing such cases much easier. Such situation might happen because certificates are downloaded by `ya-installer`, so if `ya-provider` is installed in some different way it may end up with empty `cert-dir`.

golmek:
Not critical path

PR #2583, opened 2023-05-16
Certificate verification rejection messages update

pwalski:
Resolves [#2520](https://github.com/golemfactory/yagna/issues/2520)


utils/manifest-utils/src/keystore/x509_keystore.rs
```diff
@@ -420,21 +420,24 @@ impl X509Keystore {
...
+            bail!("Unable to verify X509 certificate. No X509 certificates in keystore.")
+        }
         let mut cert_stack = openssl::stack::Stack::new()?;
         for cert in cert_chain {
             cert_stack.push(cert).unwrap();
         }
         let mut ctx = X509StoreContext::new()?;
         if !(ctx.init(&store.store, &cert, &cert_stack, |ctx| ctx.verify_cert())?) {
-            bail!("Invalid certificate");
+            bail!("Unable to verify X509 certificate.");
```
pwalski, line 440: Instead of returning "invalid certificate" I changed it to "Unable to verify X509 certificate", because:
1. I wanted to be clear it is a part of X509 verification logic.
2. Inability to verify certificate happens when Provider has no issuer in keystore. Message "Invalid certificate" was misleading in this context (it could be interpreted as malformed cert content).

utils/manifest-utils/src/keystore/x509_keystore.rs
```diff
@@ -420,21 +420,24 @@ impl X509Keystore {
...
         let store = self
             .store
             .read()
             .map_err(|err| anyhow::anyhow!("Err: {}", err.to_string()))?;
-        let cert = match cert_chain.last().map(Clone::clone) {
-            Some(cert) => cert,
-            None => bail!("Unable to verify certificate. No certificate."),
-        };
+        if store.store.objects().is_empty() {
+            bail!("Unable to verify X509 certificate. No X509 certificates in keystore.")
```
pwalski, line 432: It was important to me because it reveals the state of Provider's keystore.

kamirr (approved):
LGTM

PR #2597, opened 2023-05-25
mergeback release/v0.12

kamirr:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "yagna"
-version = "0.11.0"
+version = "0.12.1"
```
evik42, line 3: Should not master be 0.13?
I think currently we have the released version in cargo files and maybe the version for the next release would make more sense.

evik42 (approved):

Cargo.toml
kamirr: I believe the convention is to bump the version when you're doing the release, [tokio](https://github.com/tokio-rs/tokio/commit/e87ff8a83a2edea1b23cf62269b8b31ef7f1df2d#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R57) and [ripgrep](https://github.com/BurntSushi/ripgrep/commit/af6b6c543b224d348a8876f0c06245d9ea7929c5) (first library and binary that I thought of) do so.

PR #2614, opened 2023-06-07
P2p stability - support for updated ya-relay

nieznanysprawiciel:
partially resolves: https://github.com/golemfactory/ya-relay/issues/238


core/gsb-api/src/api.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(non_snake_case)]
```
MrDarthShoe: Why this is needed?

MrDarthShoe (dismissed):
LGTM, as hybrid_v2 is a copy of hybrid aligned to new version of ya-relay-client

MrDarthShoe (changes requested):
It shouldn't be merged now as we are not sure about stability and potential problems with new relay-client

Issue #2605, opened 2023-06-01
Auto add specific rules for certs during installation [1/2]


golmek:
**User Story:**
As a provider
I want to have default golem certificates & rules set accordingly
So that I don't need to do it manually by myself

**Acceptance Criteria:**
Add specific rules for certs during installation:
1. Audited payload cert + rule
2. Partner cert + rule

This task is  should proceed on all elements which can be done without adding yet Golem Root Certificate. Placeholders should be used as possible. 

**Testability:**
1. Manual tests
2. Testing session

**Not In Scope:**

- adding Golem Root Cert: https://github.com/golemfactory/yagna/issues/2561

**Additional Information**

- Next task: https://github.com/golemfactory/yagna/issues/2433

PR #2615, opened 2023-06-09
Add certs during installation

MrDarthShoe:
Resolves https://github.com/golemfactory/yagna/issues/2605

### DIFF
--- a/golem_cli/src/command/provider.rs
+++ b/golem_cli/src/command/provider.rs
@@ -1,6 +1,7 @@
 use anyhow::Context;
 use serde::Deserialize;
 use std::ffi::OsStr;
+use std::path::Path;
 use std::{collections::BTreeMap, process::Stdio};
 use tokio::process::{Child, Command};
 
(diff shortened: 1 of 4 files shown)


golem_cli/src/command/provider.rs
```diff
@@ -302,6 +302,36 @@ impl YaProviderCommand {
...
+        if output.status.success() {
+            Ok(())
+        } else {
+            let output = String::from_utf8_lossy(&output.stderr);
+            Err(anyhow::anyhow!("{}", output))
+        }
+    }
+
+    #[allow(dead_code)]
     pub async fn add_certs<I, S>(self, certs: I) -> anyhow::Result<()>
```
staszek-krotki, line 335: If this function is not used is it really worth keeping it here?

golem_cli/src/manifest.rs
```diff
@@ -35,40 +35,48 @@ pub async fn manifest_bundle(command: ManifestBundleCommand) -> Result<i32> {
 }
 
 pub async fn add_manifest_bundle(path: String) -> Result<i32> {
-    add_certs(&path).await?;
     add_whitelisted_domains(&path).await?;
+    set_rules(&path).await?;
```
staszek-krotki, line 39: Minor aesthetics: It is probably only a matter of preference but I'd rather get rid of the `set_rules` function and would put its content directly here - this make it 0.1 sec faster to read.  
```suggestion
    set_audited_payload_rules(&path).await?;
    set_partner_rules(&path).await?;
```

golem_cli/src/manifest.rs
```diff
@@ -35,40 +35,48 @@ pub async fn manifest_bundle(command: ManifestBundleCommand) -> Result<i32> {
...
-    let certs = directory
-        // canonicalize first to resolve symlinks
-        .map(|path| path.and_then(|p| p.path().canonicalize()))
-        // this will ignore paths that has errors (for example broken symlinks)
-        .filter(|path| path.as_ref().map(|p| p.is_file()).unwrap_or(false))
-        .map(|path| path.unwrap())
-        .collect::<Vec<_>>();
+async fn set_rules(path: &String) -> Result<()> {
+    set_audited_payload_rules(path).await?;
+    set_partner_rules(path).await?;
```
staszek-krotki, line 46: This could then be removed
```suggestion
```

golem_cli/src/command/provider.rs
MrDarthShoe: leftover

staszek-krotki (approved):
Looks good!

Issue #2609, opened 2023-06-05
Progress suport in yagna


prekucki:
- [ ] extend api to support progress tracing in utils/transfer
- [ ] modify example in utils/transfer to add progress bar.
- [ ] implement `progress` option in transfer. (TODO spec)
- [ ] implement `progress` option in deploy. (TODO spec)

prekucki:
yagna branch:  evik42/deploy-detailed-progress

PR #2622, opened 2023-06-14
Transfer progress support on streams

evik42:
Transfer progress support wrapper for streams and sinks.

Extended the logic of TransferStreams to set the total data size in context when it is known. `DirTransferProvider` does streaming archiving/compression so it is not able to set total data size.

Related to #2609

### DIFF
--- a/utils/transfer/src/progress.rs
+++ b/utils/transfer/src/progress.rs
@@ -1,0 +1,41 @@
+use crate::error::Error;
+use crate::{TransferStream, abortable_stream};
+use crate::{TransferContext, TransferData};
+use futures::{SinkExt, StreamExt, TryFutureExt};
+use tokio::task::spawn_local;
+
+type Stream = TransferStream<TransferData, Error>;
+
+/// Wraps a stream to report progress.
+/// The `report` function is called with the current offset and the total size.
+/// The total size is 0 if the size is unknown. (For example, when the source is a directory.)
+pub fn wrap_with_progress_reporting<F>(mut source: Stream, ctx: &TransferContext, report: F) -> Stream
+where
+    F: Fn(u64, u64) -> () + Send + 'static,
+{
+    let (stream, tx, abort_reg) = Stream::create(1);
+    let mut txc = tx.clone();
+    let state = ctx.state.clone();
+
+    spawn_local(async move {
+        let fut = async move {
+            while let Some(result) = source.next().await {
+                if let Ok(data) = result.as_ref() {
+                    let data_len = data.as_ref().len() as u64;
+                    let offset = state.offset() + data_len as u64;
+                    report(offset, state.size().unwrap_or(0));
+                }
+                txc.send(result).await?;
+            };
+            Ok::<(), Error>(())
+        }
+        .map_err(|error| {
+            log::error!("Error forwading data: {}", error);
+            error
+        });
+
+        abortable_stream(fut, abort_reg, tx).await
+    });
+
+    stream
+}
--- a/utils/transfer/src/progress.rs
+++ b/utils/transfer/src/progress.rs
@@ -1,5 +1,5 @@
 use crate::error::Error;
-use crate::{TransferStream, abortable_stream};
+use crate::{TransferStream, abortable_stream, TransferSink, abortable_sink};
 use crate::{TransferContext, TransferData};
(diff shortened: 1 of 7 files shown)


utils/transfer/src/progress.rs
```diff
@@ -0,0 +1,93 @@
...
+use crate::{TransferContext, TransferData};
+use futures::{SinkExt, StreamExt, TryFutureExt};
+use tokio::task::spawn_local;
+
+type Stream = TransferStream<TransferData, Error>;
+
+/// Wraps a stream to report progress.
+/// The `report` function is called with the current offset and the total size.
+/// The total size is 0 if the size is unknown. (For example, when the source is a directory.)
+/// The reporting function should not block as it will block the transfer.
```
nieznanysprawiciel, line 12: I would recommend using `Option::None` in case size is unknown. This way we don't have magic number with special semantic.
Other possibility is to create enum with variant Unknown

utils/transfer/src/progress.rs
```diff
@@ -0,0 +1,93 @@
...
+                    break;
+                }
+            }
+
+            Ok::<(), Error>(())
+        }
+        .map_err(|error| {
+            log::error!("Error forwarding data: {}", error);
+            error
+        });
```
nieznanysprawiciel, line 87: Could https://docs.rs/futures/latest/futures/sink/trait.SinkExt.html#method.with be used here?

utils/transfer/src/progress.rs
```diff
@@ -0,0 +1,93 @@
...
+use futures::{SinkExt, StreamExt, TryFutureExt};
+use tokio::task::spawn_local;
+
+type Stream = TransferStream<TransferData, Error>;
+
+/// Wraps a stream to report progress.
+/// The `report` function is called with the current offset and the total size.
+/// The total size is 0 if the size is unknown. (For example, when the source is a directory.)
+/// The reporting function should not block as it will block the transfer.
+pub fn wrap_stream_with_progress_reporting<F>(
```
nieznanysprawiciel, line 13: I would make this member function on `TransferStream`

utils/transfer/src/progress.rs
```diff
@@ -0,0 +1,93 @@
...
+    stream
+}
+
+type Sink = TransferSink<TransferData, Error>;
+
+/// Wraps a sink to report progress.
+/// The `report` function is called with the current offset and the total size.
+/// The total size is 0 if the size is unknown. (For example, when the source is a directory.)
+/// The reporting function should not block as it will block the transfer.
+pub fn wrap_sink_with_progress_reporting<F>(
```
nieznanysprawiciel, line 55: I would make this member function on `TransferSink`

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

utils/transfer/src/progress.rs
```diff
@@ -0,0 +1,93 @@
...
+            }
+            Ok::<(), Error>(())
+        }
+        .map_err(|error| {
+            log::error!("Error forwarding data: {}", error);
+            error
+        });
+
+        abortable_stream(fut, abort_reg, tx).await
+    });
```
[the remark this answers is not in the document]
evik42: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.inspect is the appropriate one. I tried but that returns a Stream and trying to package that into TransferStream hit the problem that TransferState is not Send.

utils/transfer/src/progress.rs
evik42: The problem TransferStream is generic (although I think at all places it is `TransferStream<TransferData,  ya_transfer::error>` but for some reason this needs to be an open type) so to accomodate that I would need to either do:

```rust
impl TransferStream<TransferData, Error> {
```
or
```rust
impl <E> TransfreStream<TransferData, E> {
```
to accommodate a generic error type _or_ to better support general usage, I would need to create a trait to get the data length like:
```rust
trait Measurable {
  get_len(&self) -> u64
}
impl <T: Measurable, R> TransferStream<T, E> {
...
}
impl Measurable for Transfer Data {
...
```
I am not sure which version would be better or if it is worth it.

nieznanysprawiciel: It seems not worth if it requires so much work.

PR #2627, opened 2023-06-19
Add goth test_e2e_outbound_perf

kamirr:
This test deploys an outbound-test image (see `assets/test_e2e_outbound_perf`) which runs outbound diagnostics at parameters specified in the python test. Then the correct results are asserted in the test.

Payment / invoicing are omitted from testing because other integration tests already cover those and this test is already rather long.


goth_tests/e2e/vm/assets/test_e2e_outbound_perf/image/outbound-bench/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    let mut requests_to_run = Vec::new();
+    for request in requests {
+        for _ in 0..(total_reqs / requests.len()) {
+            requests_to_run.push(request);
+        }
+    }
+    while requests_to_run.len() < total_reqs {
+        requests_to_run.push(requests[0]);
+    }
```
MrDarthShoe, line 120: Maybe one loop with some indexing logic would be simpler?
```rust
for i in 0..total_reqs {
    requests_to_run.push(requests[i % requests.len()]);
}
```

goth_tests/e2e/vm/assets/test_e2e_outbound_perf/image/outbound-bench/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    requests_to_run.shuffle(&mut rand::thread_rng());
+
+    let started_at = Instant::now();
+    let result = tokio::runtime::Runtime::new()
+        .unwrap()
+        .block_on(async move {
+            let mut set = JoinSet::new();
+            for url in requests_to_run {
+                set.spawn(reqwest::get(url));
```
MrDarthShoe, line 130: you can use "blocking" feature in reqwest crate so you probably don't have to use tokio runtime here

goth_tests/e2e/vm/assets/test_e2e_outbound_perf/image/outbound-bench/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+    time::{Duration, Instant},
+};
+
+use clap::Parser;
+use rand::{seq::SliceRandom, RngCore};
+use serde_json::Value;
+use tokio::task::JoinSet;
+
+fn test_roundtrip(mib_per_s: f32, stream: &impl Fn() -> TcpStream) -> anyhow::Result<bool> {
+    let mib: usize = (mib_per_s * 10.0) as usize;
```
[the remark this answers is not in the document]
kamirr: Done

goth_tests/e2e/vm/assets/test_e2e_outbound_perf/image/outbound-bench/src/main.rs
kamirr: I opted to use `tokio` to run all of those requests asynchronously instead of spawning a thread for each of them. Notice that they run at the same time using the `JoinSet`.

MrDarthShoe (approved):

MrDarthShoe (approved):

MrDarthShoe (approved):

PR #2649, opened 2023-06-30
Added possibility of exporting raw keys.

scx1332:
Added possibility of exporting accounts to plain ETH private keys. Exporting to keystore files is extremely annoying, because most wallets do not support of importing keystores with empty passwords.

yagna app-key export --plain
yagna app-key export <key_id> --plain

If account ID is locked then you have unlock it first before you can export by --plain (it is handled properly).

I had to make a bit of refactor, because SecretKey from web3 does not allow exporting to string. This design choice of web3 make this task from few easy lines of code to bigger refactor and additional imports :(. 
Accessing private fields of library is not possible in Rust (even with unsafe code).
And we don't want to create fork of web3.


golmek:
Blocked: To figure out approach towards Web3 fork we have in Golem Factory.

nieznanysprawiciel (approved):

PR #2653, opened 2023-07-10
Runtime test started by exe-unit

pwalski:


exe-unit/src/bin.rs
```diff
@@ -247,6 +249,16 @@ async fn run() -> anyhow::Result<()> {
             println!("{}", serde_json::to_string(&offer_template)?);
             return Ok(());
         }
+        Command::Test => {
+            let args = cli.runtime_arg.clone();
+            let output = ExeUnit::<RuntimeProcess>::test(cli.binary, args)?;
+            println!("{}", String::from_utf8_lossy(&output.stdout));
+            eprintln!("{}", String::from_utf8_lossy(&output.stderr));
```
pwalski, line 256: Printing of `stdout` and `stderr` is only to make code here able to find `--help` 
https://github.com/golemfactory/yagna/pull/2653/files#diff-511c1f91849eca66f6bbfebdc96af3706d55cdc82f2b30ff13df3a8d4149aa7aR373
The way `test` command in handled is a technical debt, and it should be paid by implementing `ya-runtime-sdk/api` in WASM runtime. Alternatively it should be dropped.

nieznanysprawiciel (approved):

PR #2657, opened 2023-07-13
Outbound 2.0 add warning in CLI & reload keystore when reloading rulestore

MrDarthShoe:


agent/provider/src/rules.rs
```diff
@@ -187,16 +187,29 @@ impl RulesManager {
...
+            let cert_dir = self.cert_dir.clone();
             let manager = self.clone();
-            let handler = move |p: PathBuf| match manager.rulestore.reload() {
-                Ok(()) => {
-                    log::info!("rulestore updated from {}", p.display());
+            let handler = move |p: PathBuf| {
+                // Reload also keystore to avoid file-monitor race when doing `import-cert`
+                match manager.keystore.reload(&cert_dir) {
+                    Ok(()) => {
+                        log::info!("Trusted keystore updated because rulestore changed");
```
pwalski, line 196: Will it not be printed every time yagna starts up?

pwalski (approved):

agent/provider/src/rules.rs
MrDarthShoe: No, it will be printed only when rulestore changes (either from CLI or from manual changes to file directly)

PR #2680, opened 2023-08-03
Bump ya-relay-stack to use more reasonable buffer sizes

evik42:
Yagna uses a lot of memory because ya-relay-stack by default allocates 4 MB receive buffer for each tcp connection.
This is the _maximum_ size of the buffer in the kernel and ya-relay-stack does not resize the buffer based on usage so it is wasting a lot of memory.
The bumped version uses the `default` buffer sizes instead of `max`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8663,8 +8663,7 @@ dependencies = [
 [[package]]
 name = "ya-relay-client"
 version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "48d0c86cd2a4430f08bc60aeeccabcf9448b5e08189b3d2bdce36bb3d67910e0"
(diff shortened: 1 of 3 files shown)


nieznanysprawiciel:
This change might affect GFTP and VPN transfer speed performance. Setting buffer size was one of the optimizations made

evik42:
> This change might affect GFTP and VPN transfer speed performance. Setting buffer size was one of the optimizations made

That can be addressed by using a different buffer size for Transfer type sockets.

This change does not break any tests that verify the minimum required performance.

evik42:
> This change might affect GFTP and VPN transfer speed performance. Setting buffer size was one of the optimizations made

There is also socket metrics saved and based on that the buffer parameters could be adjusted after creation.

PR #2727, opened 2023-09-01
Enable amending allocations

kamirr:


core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
+    let new_allocation: NewAllocation = body.into_inner();
+    let dao: AllocationDao = db.as_dao();
+
+    let current_allocation = match dao.get(allocation_id.clone(), node_id).await {
+        Ok(AllocationStatus::Active(allocation)) => allocation,
+        Ok(AllocationStatus::Gone) => {
+            return response::gone(&format!(
+                "Allocation {} has been already released",
+                allocation_id
+            ))
```
nieznanysprawiciel, line 204: Maybe?
```suggestion
            return response::gone(&format!(
                "Allocation {allocation_id} has been already released"
            ))
```

core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
+    new_allocation: NewAllocation,
+) -> Allocation {
+    Allocation {
+        allocation_id: old_allocation.allocation_id.clone(),
+        address: new_allocation
+            .address
+            .unwrap_or_else(|| old_allocation.address.clone()),
+        payment_platform: new_allocation
+            .payment_platform
+            .unwrap_or_else(|| old_allocation.payment_platform.clone()),
```
nieznanysprawiciel, line 177: I don't think updating platform makes any sense.
We should return error in such case.

Updating platform can cause unwanted problems if you already spent/reserved money on different platform

core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
 
+fn amend_allocation_fields(
+    old_allocation: Allocation,
+    new_allocation: NewAllocation,
+) -> Allocation {
+    Allocation {
+        allocation_id: old_allocation.allocation_id.clone(),
+        address: new_allocation
+            .address
+            .unwrap_or_else(|| old_allocation.address.clone()),
```
nieznanysprawiciel, line 174: I don't think user should be able to update address

core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
+        payment_platform: new_allocation
+            .payment_platform
+            .unwrap_or_else(|| old_allocation.payment_platform.clone()),
+        total_amount: new_allocation.total_amount.clone(),
+        spent_amount: old_allocation.spent_amount.clone(),
+        remaining_amount: new_allocation.total_amount - &old_allocation.spent_amount,
+        timestamp: old_allocation.timestamp,
+        timeout: new_allocation.timeout.or(old_allocation.timeout),
```
nieznanysprawiciel, line 182: Timeout should be validated if it isn't in the past

core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
+        make_deposit: new_allocation.make_deposit,
```
nieznanysprawiciel, line 183: If we made deposit earlier, should we change this?

core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
+        Err(e) => return response::server_error(&e),
+    };
+
+    let amended_allocation = amend_allocation_fields(current_allocation.clone(), new_allocation);
+
+    if amended_allocation.remaining_amount < BigDecimal::from(0) {
+        return response::bad_request(
+            &"Amended allocation would be smaller than the amount already spent",
+        );
+    }
```
nieznanysprawiciel, line 216: I would hide all checks in `amend_allocation_fields`.
I don't think we need incorrect allocation at any point

core/payment/src/api/allocations.rs
```diff
@@ -162,12 +163,83 @@ async fn get_allocation(
...
+
+    let validate_msg = ValidateAllocation {
+        platform: amended_allocation.payment_platform.clone(),
+        address: amended_allocation.address.clone(),
+        amount: amended_allocation.total_amount.clone() - &current_allocation.total_amount,
+    };
+    match async move { Ok(bus::service(LOCAL_SERVICE).send(validate_msg).await??) }.await {
```
nieznanysprawiciel, line 233: Did you check how the validation will behave in corner cases?
For example if we have other allocations, does this function checks the sum of all allocations?
Because if yes, than it already has taken into account the amount from allocation before amendment.

nieznanysprawiciel (changes requested):

core/payment/src/api/allocations.rs
kamirr: Done

core/payment/src/api/allocations.rs
kamirr: Done

core/payment/src/api/allocations.rs
kamirr: make_deposit is no longer amendable

core/payment/src/api/allocations.rs
kamirr: The corner case with amending payment-platform no longer exist.

As for validation of allocations, yes, the implementation checks whether the sum of *all* allocations can be paid. Because this takes into account the yet-unmodified allocation, we should only pass the excess for validation (if increasing the allocation) or 0 (if we're shrinking it).

The validation could be skipped entirely for decreasing allocations, but we might wish to extend the validation logic later on so keeping it here makes things easier.

nieznanysprawiciel (approved):

core/payment/src/api/allocations.rs
kamirr: Done

PR #2739, opened 2023-09-08
faster integration tests

scx1332:


.github/workflows/binaries-x86-64.yml
```diff
@@ -67,11 +67,7 @@ jobs:
 
       - name: Build binaries
         run: |
-          cargo build --features static-openssl --target x86_64-unknown-linux-musl
-          (cd core/gftp && cargo build --bin gftp -p gftp --features bin --target x86_64-unknown-linux-musl)
-          (cd golem_cli && cargo build --bin golemsp -p golemsp --target x86_64-unknown-linux-musl)
-          (cd agent/provider && cargo build --bin ya-provider -p ya-provider --target x86_64-unknown-linux-musl)
-          (cd exe-unit && cargo build --bin exe-unit -p ya-exe-unit --features openssl/vendored --target x86_64-unknown-linux-musl)
+          cargo build --all --features static-openssl --target x86_64-unknown-linux-musl
```
evik42, line 70: Previously `yagna` and `exe-unit` was building with different `openssl` flags.
Why change that?

scx1332: Why is exe-unit build with different openssl flag?

evik42: Ask @prekucki who added these lines to the release build?
Also if you change it here but not in the release build then the tested version differs from the released one...

PR #2769, opened 2023-09-20
ya_vm_runtime when called with offer-template sometimes fails

scx1332:
ya_vm_runtime when called with offer-template sometimes leaves few random bytes at the end of the stream (unfortunately I don't know why). The solution is to try parse serde with cutoff string.


kamirr (approved):
Frankly, I don't like this hack. Can you create an issue for fixing the core problem so that we can remove this at some point?

PR #2777, opened 2023-09-26
file-logging: allow configuration of logging parameters via ENV

kamirr:
The following ENVs have been added:
* `LOG_FILES_UNCOMPRESSED` with default of 1
* `LOG_FILES_COMPRESSED` with default of 10
* `LOG_ROTATE_AGE` with default of a day
* `LOG_ROTATE_SIZE`:
  - if unset no rotation due to size happens
  - default in `.env-template` is 1GiB

### DIFF
--- a/.env-template
+++ b/.env-template
@@ -1,6 +1,15 @@
-# Log level.
-# Default is info, but if you want to debug this is a magic setup, to skip payment driver overwhelming logs
-#RUST_LOG=debug,tokio_core=info,tokio_reactor=info,hyper=info,reqwest=info
+# Logging.
+# Log level. Default is info, but if you want to debug this is a magic setup, to skip payment driver overwhelming logs.
+# RUST_LOG=debug,tokio_core=info,tokio_reactor=info,hyper=info,reqwest=info
+# Configuration of dealing with log file history.
+# How many uncompressed logs to keep (do note that there's an additional yagna_rCURRENT.log).
+LOG_FILES_UNCOMPRESSED=1
+# How many compressed logs to keep.
+LOG_FILES_COMPRESSED=10
+# How often to rotate logs, either DAY or HOUR (case insensitive).
+LOG_ROTATE_AGE=day
+# How big of a log to rotate (even if younger than LOG_AGE). Size in bytes (the default below is 1GiB). Unset do disable.
+LOG_ROTATE_SIZE=1073741824
 
 # Application working directory path.
 YAGNA_DATADIR="."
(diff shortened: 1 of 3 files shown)


kamirr:
Goth run: https://github.com/golemfactory/yagna/actions/runs/6323589497

.env-template
```diff
@@ -1,6 +1,15 @@
...
+# Log level. Default is info, but if you want to debug this is a magic setup, to skip payment driver overwhelming logs.
+# RUST_LOG=debug,tokio_core=info,tokio_reactor=info,hyper=info,reqwest=info
+# Configuration of dealing with log file history.
+# How many uncompressed logs to keep (do note that there's an additional yagna_rCURRENT.log).
+LOG_FILES_UNCOMPRESSED=1
+# How many compressed logs to keep.
+LOG_FILES_COMPRESSED=10
+# How often to rotate logs, either DAY or HOUR (case insensitive).
+LOG_ROTATE_AGE=day
+# How big of a log to rotate (even if younger than LOG_AGE). Size in bytes (the default below is 1GiB). Unset do disable.
```
nieznanysprawiciel, line 11:
```suggestion
# How big of a log to rotate (even if younger than LOG_ROTATE_AGE). Size in bytes (the default below is 1GiB). Unset do disable.
```

utils/file-logging/src/lib.rs
```diff
@@ -61,14 +62,85 @@ fn log_format_color(
...
+    fn new(dir: &'a Path) -> Self {
+        FileLogConfig {
+            dir,
+            keep_compressed: 10,
+            keep_uncompressed: 1,
+            rotate_age: Age::Day,
+            rotate_size: None,
+        }
+    }
+}
```
nieznanysprawiciel, line 128: By using `structopt` you could avoid most of the parsing code.

utils/file-logging/src/lib.rs
```diff
@@ -95,12 +167,26 @@ pub fn start_logger(
...
     let mut logger = Logger::with(log_spec).format(log_format);
+
+    let mut messages = Vec::new();
     if let Some(log_dir) = log_dir {
-        logger = set_logging_to_files(logger, log_dir);
+        let config = FileLogConfig::with_env(log_dir, &mut messages);
+        messages.push((
+            log::Level::Info,
+            format!("Logging configuration: {config:#?}"),
+        ));
```
nieznanysprawiciel, line 177: It looks a little bit over-engineered

kamirr: I don't see any other way to properly report the error here. We could just abort the program, but I think logging errors is more in-line with the design of yagna. Of course you can't just log directly, because the logger can't be initialized here.

utils/file-logging/src/lib.rs
kamirr: I considered that, but `structopt` doesn't allow me to communicate detect per env without writing as much boilerplate as there already is AFAICT.

nieznanysprawiciel (approved):

PR #2788, opened 2023-09-28
Scx1332/no auth

scx1332:


kamirr (approved):
Code looks good but I don't know if it's in line with yagna design goals, so I can't sign off on that part.

PR #2812, opened 2023-10-03
Goth. Docker system prune if older than 12h.

pwalski:


scx1332:
Issue with pull rate limit is fixed now. We can prune everything

Issue #455, opened 2020-07-24
Yagna id drop fails


maaktweluit:
When trying to delete an id i get the following. error:
```
Error: bad request: No service registered under given address
```

Steps taken:
```
$ yagna id list
┌───────────┬──────────┬─────────┬──────────────────────────────────────────────┐
│  default  │  locked  │  alias  │  address                                     │
├───────────┼──────────┼─────────┼──────────────────────────────────────────────┤
│  X        │          │         │  0x17ec8597ff92c3f44523bdc65bf0f1be632917ff  │
│           │          │         │  0x00ab0d402d3d5ebdef4b405a8f6b62d66946db24  │
└───────────┴──────────┴─────────┴──────────────────────────────────────────────┘
```
```
$ yagna id drop --help
yagna-id-drop 0.2.0
Drop given identity

USAGE:
    yagna id drop [FLAGS] [OPTIONS] <node-or-alias>

FLAGS:
        --json       Return results in JSON format
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -d, --datadir <data-dir>    Service data dir [env: YAGNA_DATADIR]  [default: "/Users/mwu-gol/Library/Application
                                Support/GolemFactory.yagna"]

ARGS:
    <node-or-alias>    Identity alias to drop
```
```
$ yagna id drop 0x00ab0d402d3d5ebdef4b405a8f6b62d66946db24
Error: bad request: No service registered under given address
```
```
$ yagna id list
┌───────────┬──────────┬─────────┬──────────────────────────────────────────────┐
│  default  │  locked  │  alias  │  address                                     │
├───────────┼──────────┼─────────┼──────────────────────────────────────────────┤
│  X        │          │         │  0x17ec8597ff92c3f44523bdc65bf0f1be632917ff  │
│           │          │         │  0x00ab0d402d3d5ebdef4b405a8f6b62d66946db24  │
└───────────┴──────────┴─────────┴──────────────────────────────────────────────┘
```

mateuszsrebrny:
Previous attempt to implement drop: https://github.com/golemfactory/yagna/pull/1537

grisha87:
I would like to upvote this issue, as I would like to clean-up redundant IDs on my machine using the CLI and not hack the database.

prekucki:
Test scenarios:

## TC1: Deleting the Only Identity

1. Start the server in a clean environment.
2. List `yagna id list` and observe that only one identity is listed.
[...]

PR #2816, opened 2023-10-04
drop id 

staszek-krotki:


prekucki:
resolves: #455

nieznanysprawiciel (changes requested):
It is not enough to mark identity in database. Other services in yagna might use it.
For example payment module signs transactions and probably should be notified. What happens If you remove identity during task computations? I'm not sure about consequences, but probably we will not be able to sign new transactions.

Hybrid Net module has services vector that stores identity prefixed paths.
And it uses default identity to identify itself in the network. So if we removed default identity. we might have unexpected consequences.

There is authentication middleware, that probably caches identities.

If we remove identity what should happen to app-keys created for this identity? If they should be removed than we have cors middleware, that caches app-keys information.

Dropping identity is not trivial problem and that why it wasn't implemented yet.
Last attempt is here: https://github.com/golemfactory/yagna/pull/1537

kamirr:
> I'm not sure about consequences, but probably we will not be able to sign new transactions.

That is correct. The new payment driver will handle the error somewhat gracefully, but we obviously don't have a copy of the private key and would fail while signing subsequent transactions.

I think a good approach would be to broadcast the intent to remove an `X` identity and only proceed if no subsystems object. The payment driver, for example, would prevent removal of a key that has pending payments. Adding a REST endpoint for that as well would allow the agent to object as well.

prekucki (dismissed):

PR #2825, opened 2023-10-05
Goth supporting multipayments

scx1332:


core/payment/src/processor.rs
```diff
@@ -504,8 +504,13 @@ impl PaymentProcessor {
             Ok(confirmation) => PaymentConfirmation { confirmation },
             Err(e) => return Err(VerifyPaymentError::ConfirmationEncoding),
         };
+        //this need to be rewritten
```
kamirr, line 507: Same concerns as in my `THIS CODE NEEDS  REFACTOR` comment. If there's something specific to do please create an issue and link it in this comment.

goth_tests/pyproject.toml
```diff
@@ -30,7 +30,7 @@ pytest-asyncio = "0.21"
 pytest-split = "^0.8.1"
 goth = "0.15.8"
 # to use development goth version uncomment below
-# goth =  { git = "https://github.com/golemfactory/goth.git", rev = "0845cad9a4d590b06849c6888db7f060788eb7d7" }
+# goth =  { git = "https://github.com/golemfactory/goth.git", rev = "47df5fa77a4e04239cf0f48d9d9a0b7e3c7468e4" }
```
kamirr, line 33: This does nothing, the line is commented out and the version of goth in `poetry.lock` is still the same.

kamirr (changes requested):

PR #2864, opened 2023-10-26
get rid of forever loop

scx1332:


scx1332:
Add parameter to config

prekucki (approved):

PR #2911, opened 2023-11-24
Remove ya-utils-scheduler

nieznanysprawiciel:


scx1332 (approved):
If not used then ok for me

PR #2958, opened 2024-01-03
Provider ctrl-c handling and exe-unit shutdown using ctrl-c

pwalski:
Resolves https://github.com/golemfactory/gamerhash-facade/issues/30

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9184,6 +9184,7 @@ dependencies = [
  "nix 0.22.3",
  "shared_child",
  "tokio",
+ "winapi 0.3.9",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


pwalski:
I dropped idea of implementing by myself signal handling, because it was difficult to match futures handling for both linux and Windows implementation.
Initially I tried to use `ctrlc` crate with `termination` feature https://github.com/Detegr/rust-ctrlc?tab=readme-ov-file#handling-sigterm-and-sighup
Unfortunately it broke handling agreements.
Then I switched to tokio::signal::ctrl_c() https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html
The problem is it handles only `SIGINT` (or ctrl-c on Windows).

nieznanysprawiciel (approved):

PR #2976, opened 2024-01-09
payment-driver: don't create DB for erc20next

kamirr:
Also:
* `db: DbExecutor` arguments in PaymentDriver methods were all dead code, so I removed them.


scx1332 (approved):
No db - good

PR #2978, opened 2024-01-10
Runtime stopping on Windows

pwalski:
Resolves: https://github.com/golemfactory/gamerhash-facade/issues/30


pwalski:
@prekucki https://github.com/golemfactory/yagna/pull/2978/commits/0f9d1d8a20fdbcb414226d4ceff22daeac48b44d
This commit is not the main point of the PR. I introduced this change to quickly make my local setup work (I needed relative path/pattern with multiple `../`, like `../../some/path/ya*.json` and parsing it was failing). As a result pattern `EXE_UNIT_PATH` is now parsed as a `glob`. It is compatible with existing implementation. I assumed it is better to just use a `glob` lib, because anyone who sees this variable with `*` assumes it is a glob.

Issue #2960, opened 2024-01-03
Critical PoisonError due to GSB panic


kamirr:
## Acceptance Criteria
- pls powiedz Tomek

golmek:
Due to limited reproduction (it happens very rarely) it might be released in 14.1

PR #2980, opened 2024-01-11
erc20next: remove remaining uses of LocalPoolHandle

kamirr:
Prevents potential problems with GSB panics.


kamirr:
Similar to https://github.com/golemfactory/yagna/issues/2960, but this case was never reproduced. Nevertheless, there's no need to take the risk.

scx1332 (approved):

PR #2996, opened 2024-01-16
payment: add erc20->erc20next aliasing

kamirr:


grisha87:
@prekucki , @kamirr 

Will the aliasing be consistent in both directions? 

Meaning: as a software component (SDK in our case) I'm using `erc20` as a string value. To me, "aliasing" will be "complete" when:

- I can always send `erc20` like I did since 0.13.2
- I will alrways receive `erc20` from the API like I did since 0.13.2 (no `erc20next` on output)
- I will not have to use the name `erc20next` anywhere in the SDK
  - Specifically: when I'm decorating my demand, I still want to say that I'm accepting erc20 payments

Essetially, from such a ticket I'm kind of expecting backward compatibility implementation within yagna. Is it going to work in such _transparent_ way?

PR #3030, opened 2024-01-30
http gsb proxy clean up

staszek-krotki:


core/activity/src/http_proxy.rs
```diff
@@ -59,12 +59,13 @@ async fn proxy_http_request(
...
-    }
-
-    let call_streaming = move |msg| {
+    let http_to_gsb = HttpToGsbProxy {
+        method: method.to_string(),
+        path,
+        body,
+        headers: request.headers().clone(),
+    };
+    let stream = http_to_gsb.pass(move |msg| {
```
nieznanysprawiciel, line 68: This should be hidden in library as well. You try to assume that your library doesn't know what gsb is and you must pass method of transfer from outside.
Of course you can keep this separation inside library, but library consumer rather assumes that it should use gsb.

exe-unit/components/http-proxy/src/gsb_to_http.rs
```diff
@@ -0,0 +1,106 @@
...
+            let mut builder = client.request(method, &url);
+
+            builder = Headers::add(builder, message.headers);
+
+            log::info!("Calling {}", &url);
+            let response = builder.send().await;
+            let response = match response {
+                Ok(response) => response,
+                Err(err) => {
+                    panic!("Error {}", err);
```
nieznanysprawiciel, line 40: Don't panic in production code

exe-unit/components/http-proxy/src/gsb_to_http.rs
```diff
@@ -0,0 +1,106 @@
...
+
+            let str_bytes = String::from_utf8(bytes.to_vec()).unwrap();
+
+            let response = GsbHttpCallResponseEvent {
+                index: 0,
+                timestamp: Utc::now().naive_local().to_string(),
+                msg: str_bytes,
+            };
+
+            tx.send(response).await.unwrap();
```
nieznanysprawiciel, line 56: Don't `unwrap` in production code

exe-unit/components/http-proxy/src/gsb_to_http.rs
```diff
@@ -0,0 +1,106 @@
...
+                Err(err) => {
+                    panic!("Error {}", err);
+                }
+            };
+
+            log::info!("Got response");
+
+            let bytes = response.bytes().await.unwrap();
+
+            let str_bytes = String::from_utf8(bytes.to_vec()).unwrap();
```
nieznanysprawiciel, line 48: I'm not sure if it is a good idea. Is image convertible to utf8 string?
And in general the content can contain anything. We should have bytes array instead.

exe-unit/components/http-proxy/src/message.rs
```diff
@@ -0,0 +1,28 @@
...
+use serde_json::value::Value;
+use std::collections::HashMap;
+use ya_service_bus::{RpcMessage, RpcStreamMessage};
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GsbHttpCallMessage {
+    pub method: String,
+    pub path: String,
+    pub body: Option<Map<String, Value>>,
```
nieznanysprawiciel, line 14: Why do you assume any structure for body?

exe-unit/components/http-proxy/src/response.rs
```diff
@@ -0,0 +1,8 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct GsbHttpCallResponseEvent {
+    pub index: usize,
+    pub timestamp: String,
+    pub msg: String,
```
nieznanysprawiciel, line 7: This should be bytes array. We aren't always able to convert to string.

exe-unit/components/http-proxy/src/gsb_to_http.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+impl GsbToHttpProxy {
+    pub fn pass(
+        &mut self,
+        message: GsbHttpCallMessage,
+    ) -> impl Stream<Item = GsbHttpCallResponseEvent> {
+        let url = format!("{}{}", self.base_url, message.path);
+
+        let (tx, mut rx) = mpsc::channel(24);
```
nieznanysprawiciel, line 21: Not sure why you allocate 24 elements?
This channel will be used only once as I understand code. Am I right?

exe-unit/components/http-proxy/src/headers.rs
```diff
@@ -0,0 +1,40 @@
+use crate::http_to_gsb::HttpToGsbProxy;
+use actix_http::header::HeaderMap;
+use reqwest::RequestBuilder;
+use std::collections::HashMap;
+
+pub struct Headers {}
```
nieznanysprawiciel, line 6: You use this struct as namespacing. It doesn't have any other meaning.
You could make all these functions outside of `Headers` scope and write everywhere in code `headers::filter` etc. You would have the same effect.
Other option is in every module you can define submodule like this:
```
mod headers {
    // code here
}
```

exe-unit/components/http-proxy/src/headers.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl Headers {
+    const IGNORED_HEADERS: [&'static str; 3] = ["host", "content-length", "connection"];
```
nieznanysprawiciel, line 9: You hardcoded headers.
There is always low effort way to make it more flexible. For example you could have field in `Headers` struct for keeping track what to ignore and a function for adding new headers. This could be used like this:
```
let filter = Headers::default().ignore("sth").ignore("sth2");
```
`default` would add default ignores to list.

exe-unit/components/http-proxy/src/gsb_to_http.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        tokio::task::spawn_local(async move {
+            let client = reqwest::Client::new();
+
+            let method = match message.method.to_uppercase().as_str() {
+                "POST" => actix_http::Method::POST,
+                "GET" => actix_http::Method::GET,
+                _ => actix_http::Method::GET,
```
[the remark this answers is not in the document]
nieznanysprawiciel: This is error prone, because it changes method to `GET` so user will get responses from other endpoint than expected. It is better to get nothing

exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+impl Display for GsbToHttpProxyError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            GsbToHttpProxyError::InvalidMethod => write!(f, "Invalid Method"),
+            GsbToHttpProxyError::ErrorInResponse(e) => write!(f, "Error in response {}", e),
+        }
+    }
+}
```
nieznanysprawiciel, line 29: `thiserror` has macro, you don't need to implement Display manually

exe-unit/components/gsb-http-proxy/src/message.rs
```diff
@@ -0,0 +1,26 @@
...
+use serde_derive::{Deserialize, Serialize};
+use std::collections::HashMap;
+use ya_service_bus::{RpcMessage, RpcStreamMessage};
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GsbHttpCallMessage {
+    pub method: String,
+    pub path: String,
+    pub body: Option<Vec<u8>>,
```
nieznanysprawiciel, line 12: In future we should serialize it as https://docs.rs/serde_bytes/latest/serde_bytes/
Serde handles `Vec<u8>` in very ugly way. In json it is an array with each element being separate number.

Issue #3023, opened 2024-01-24
Driver & Network Selection - new mechanism


golmek:
## Context:

Currently configuration of payment platform (driver + network + token) is done by application. It has some benefits, however also introduces important problems:

- When default driver is changed in Yagna, SDK still uses old driver
- When default network is changed in Yagna, SDK still uses old default network
- At least few times a year ETH testnet has smaller or bigger downtime or is overloaded

Consequently, we want to change our approach and get rid off precise specification of payment platform on allocation creation (flow between SDK → API → Yagna requestor). 

### Payment Platform:

Payment Platform consists of 3 values: driver, network, token

We want:

- driver - to have default value of `erc20`  (underneath new payment driver will be hidden - https://github.com/golemfactory/yagna/issues/3019)
- network - to have default value of `holesky` for `tglm` and `polygon` for `glm`
- token → For instance for default values of `erc20` or/and `holesky` it should return `tglm`. For values of `erc20` or/and `polygon` it should return `glm`

### Allocation Creation

SDK will create allocation in following fashion:
---
Example A: default driver, netwrok and token:
Request:

```jsx
{
    "totalAmount": "10",
    "makeDeposit": false
}

```

Alternative Request:

```json
{
    "paymentPlatform": {},
    "totalAmount": "10"
    "makeDeposit": false
}
```

Response:

```jsx
{
    "allocationId": "520008f2-ffb5-427e-b9ec-17173520b78b",
    "address": "0x889ff52ece3d5368051f4f8216650a7843f8926b",
    "paymentPlatform": "erc20-holesky-tglm",
    "totalAmount": "10",
    "spentAmount": "0",
    "remainingAmount": "0",
    "timestamp": "2024-01-19T12:29:35.278Z",
    "timeout": "2034-01-19T15:29:34.535594Z",
    "makeDeposit": false
}

```
---
Example B: default driver on polygon

```jsx
{
    "paymentPlatform": {
        "network": "polygon",  // Optional
    },
    "totalAmount": "10"
    "makeDeposit": false
}
```

Result:

```jsx
{
    "allocationId": "520008f2-ffb5-427e-b9ec-17173520b78b",
    "address": "0x889ff52ece3d5368051f4f8216650a7843f8926b",
    "paymentPlatform": "erc20-polygon-glm",
    "totalAmount": "10",
    "spentAmount": "0",
    "remainingAmount": "0",
    "timestamp": "2024-01-19T12:29:35.278Z",
[...]

PR #3031, opened 2024-01-30
Extend create allocation API

scx1332:
resolves: https://github.com/golemfactory/yagna/issues/3023


core/payment/src/api/allocations.rs
```diff
@@ -46,44 +112,147 @@ async fn create_allocation(
...
+    if let Err(err) = validate_driver(network, driver) {
+        let err_msg = format!("Validate driver failed (2): {err}");
+        log::error!("{}", err_msg);
+        return response::bad_request(&err_msg);
+    }
+    if let Err(err) = validate_token(network, driver, token) {
+        let err_msg = format!("Validate token failed (2): {err}");
+        log::error!("{}", err_msg);
+        return response::bad_request(&err_msg);
+    }
```
nieznanysprawiciel, line 235: From api you get these values already split. Why do you first build this string to dispatch it later?
You could avoid all these duplicated checks here

core/payment/src/api/allocations.rs
```diff
@@ -38,6 +39,71 @@ pub fn register_endpoints(scope: Scope) -> Scope {
...
+    match network {
+        "mainnet" => Ok(()),
+        "rinkeby" => Err("Rinkeby is no longer supported".to_string()),
+        "goerli" => Ok(()),
+        "holesky" => Ok(()),
+        "polygon" => Ok(()),
+        "mumbai" => Ok(()),
+        _ => Err(format!("Invalid network name: {network}")),
+    }
+}
```
nieznanysprawiciel, line 52: i think you are duplicating a lot of code here. We have enum for network already, which could be just deserialized.

core/payment/src/api/allocations.rs
```diff
@@ -38,6 +39,71 @@ pub fn register_endpoints(scope: Scope) -> Scope {
...
+
+fn validate_driver(_network: &str, driver: &str) -> Result<(), String> {
```
nieznanysprawiciel, line 54: The same here. We have enum for driver

core/payment/src/api/allocations.rs
```diff
@@ -46,44 +112,147 @@ async fn create_allocation(
...
+                            log::debug!("Network not specified, using default polygon, because token set to glm");
+                            "polygon"
+                        } else {
+                            log::debug!("Network not specified, using default holesky");
+                            "holesky"
+                        }
+                    } else {
+                        log::debug!("Network not specified and token not specified, using default holesky");
+                        "holesky"
+                    }
```
nieznanysprawiciel, line 155: Driver has `get_default_network` method. Probably you should use it, otherwise defaults will diverge.

core/payment/src/api/allocations.rs
```diff
@@ -46,44 +112,147 @@ async fn create_allocation(
...
+                    );
+                    format!("{}-{}-{}", driver, network, default_token)
+                }
+            }
+        }
+        None => {
+            log::debug!("No paymentPlatform entry found, using {DEFAULT_PAYMENT_PLATFORM}");
+            DEFAULT_PAYMENT_PLATFORM.to_string()
+        }
     };
```
nieznanysprawiciel, line 199: This implementation is so complicated, that you should hide it in separate function

nieznanysprawiciel (changes requested):

PR #3042, opened 2024-02-05
Scx1332/tokenrefaktor

scx1332:


core/payment/src/api/allocations.rs
```diff
@@ -32,187 +33,252 @@ const DEFAULT_TESTNET_NETWORK: NetworkName = NetworkName::Holesky;
...
+                        token: default_token,
+                    }
                 }
-            } else {
-                log::debug!(
-                    "Network not specified and token not specified, using default {}",
-                    DEFAULT_TESTNET_NETWORK
-                );
-                DEFAULT_TESTNET_NETWORK.into()
-            }
```
nieznanysprawiciel, line 155: This could be match statement probably:
```
match (p.driver, p.network, p.token) {
    (None, None, None) => { ... },
    (None, None, Some(token)) => { ... }
}
```

core/payment/src/api/allocations.rs
```diff
@@ -32,187 +33,252 @@ const DEFAULT_TESTNET_NETWORK: NetworkName = NetworkName::Holesky;
...
+                log::debug!("Empty paymentPlatform object, using {default_platform}");
+                default_platform
+            } else if p.token.is_some() && p.network.is_none() && p.driver.is_none() {
+                let token = p.token.as_ref().unwrap();
+                if token == "GLM" || token == "tGLM" {
+                    return Err(bad_req_and_log(format!(
+                        "Uppercase token names are not supported. Use lowercase glm or tglm instead of {}",
+                        token
+                    )));
+                } else if token == "glm" {
```
nieznanysprawiciel, line 134: It looks similar to `from_token_string` function implementation

core/payment/src/api/allocations.rs
```diff
@@ -32,187 +33,252 @@ const DEFAULT_TESTNET_NETWORK: NetworkName = NetworkName::Holesky;
...
+    }
+    impl Display for PaymentPlatformTriple {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "{}-{}-{}", self.driver, self.network, self.token)
+        }
+    }
 
-    let network = validate_network(network_str)
-        .map_err(|err| bad_req_and_log(format!("Validate network failed (2): {err}")))?;
+    fn validate_network(network: &str) -> Result<NetworkName, String> {
```
nieznanysprawiciel, line 253: Common practice would be to use `anyhow` if you need to return something generic as string.

nieznanysprawiciel (dismissed):
I believe this code could still be much more simpler than it is, but I'm not going to push further for now. I left a few comments, please look if you would like to apply them

nieznanysprawiciel (approved):

PR #3050, opened 2024-02-12
payments with signed content

staszek-krotki:


core/model/src/driver.rs
```diff
@@ -517,17 +517,39 @@ impl RpcMessage for SignPayment {
     type Error = GenericError;
 }
 
+// ************************ SIGN PAYMENT ************************
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SignPaymentCanonicalized(pub Payment);
```
nieznanysprawiciel, line 523:
```suggestion
/// We sign canonicalized version of `Payment` struct, so although we could make new struct compatible in terms of deserialization, the signature would be incorrect. That's why we need separate endpoint.
#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct SignPaymentCanonicalized(pub Payment);
```

core/payment-driver/base/src/bus.rs
```diff
@@ -74,6 +74,9 @@ pub async fn bind_service<Driver: PaymentDriver + 'static>(
         .bind_with_processor(
             move |_, dr, c, m| async move { dr.sign_payment( c, m).await }
         )
+        .bind_with_processor(
+            move |_, dr, c, m| async move { dr.sign_payment_canonicalized(c, m).await }
+        )
```
nieznanysprawiciel, line 79: I would reverse naming in driver.
`sign_payment` is our basic version so it should do the canonicalization.
And we need second function `sign_legacy` or maybe some better name.

core/model/Cargo.toml
```diff
@@ -50,3 +39,4 @@ structopt = "0.3"
 strum = "0.24"
 strum_macros = "0.24"
 thiserror = "1.0.9"
+serde_json_canonicalizer = "0.2.0"
```
nieznanysprawiciel, line 42: Is it really necessary here? I think it is unused

core/payment/src/service.rs
```diff
@@ -1073,7 +1074,43 @@ mod public {
...
+        let payment = msg.payment;
+        let signature = msg.signature;
+        if sender_id != payment.payer_id.to_string() {
+            return Err(SendError::BadRequest("Invalid payer ID".to_owned()));
+        }
+
+        let platform = payment.payment_platform.clone();
+        let amount = payment.amount.clone();
+        let num_paid_invoices = payment.agreement_payments.len() as u64;
+        match processor
```
nieznanysprawiciel, line 1110: Is the implementation similar to `send_payment`?
I think you should unify implementations and call one function from another

PR #3069, opened 2024-02-22
payments with signed content

staszek-krotki:


core/payment/src/processor.rs
```diff
@@ -434,25 +446,33 @@ impl PaymentProcessor {
...
+                        msg_with_bytes.signature.clone(),
+                        msg_with_bytes.signed_bytes.clone(),
+                    )
+                    .await?;
+                true
+            } else if send_result.is_err_and(|err| err == SendToGsbError::NotSupported) {
+                // if sending SendPaymentWithBytes is not supported then try sending SendPayment
+                match Self::send_to_gsb(payer_id, payee_id, msg).await {
+                    Ok(_) => true,
+                    Err(SendToGsbError::Rejected) => true,
```
nieznanysprawiciel, line 469: I'm not sure what should we do.
What did previous code do?

staszek-krotki: the boolean return value tells if we should mark as sent is db. This is consistent with previous behavior

core/payment/src/models/payment.rs
```diff
@@ -115,6 +131,22 @@ impl ReadObj {
...
+        activity_payments: Vec<ActivityPayment>,
+        agreement_payments: Vec<AgreementPayment>,
+    ) -> api_model::Signed<api_model::Payment> {
+        api_model::Signed {
+            payload: self
+                .clone()
+                .into_api_model(activity_payments, agreement_payments),
+            signature: Some(Signature {
+                signature: self.signature.unwrap(),
+                signed_bytes: self.signed_bytes.unwrap(),
```
nieznanysprawiciel, line 146: You should never unwrap in production code, unless you checked conditions line before.

nieznanysprawiciel (approved):

PR #3097, opened 2024-03-06
Auto cleanup was forced

scx1332:


kamirr (changes requested):
Please add the new env variables to .env-template with an explanation

scx1332:
> Please add the new env variables to .env-template with an explanation

Provider environment variables are not put in .env_template

kamirr (approved):

PR #3103, opened 2024-03-08
Provider - restrict accepted requestors

nieznanysprawiciel:
resolves: https://github.com/golemfactory/ya-runtime-ai/issues/58

todo:
- [x] Extend remove dangling rules
- [x] Keystore list new rules
- [x] Fix removing restrict rule for certificate
- [x] Don't force user to use `cert-id` instead of `import-cert` when certificate already in keystore
- ~[ ] Commands naming and UX~ Moved to task: https://github.com/golemfactory/yagna/issues/3137
- [x] Add CLI tests for new rules
- [x] [Optional] Add certificate subject to `rule list` description table column

Additional:
- Fixes problem with flattening empty objects in Offers and Proposals


staszek-krotki:
As much as I was able to get to know what it is doing - looks good. But I'd appreciate @pwalski taking a look as well

agent/provider/src/market/negotiator/builtin/allow_only.rs
```diff
@@ -0,0 +1,49 @@
...
+                log::debug!(
+                    "[AllowOnly] Rejecting Proposal from Requestor {}, reason: {reason}",
+                    demand.issuer
+                );
+                Ok(NegotiationResult::Reject {
+                    message: reason,
+                    is_final: true,
+                })
+            }
+        }
```
pwalski, line 47: `allow-only` does not check `blacklist` (and `blacklist` does not check `allow-only`).
Composite negotiator stores negotiators in HashMap and these are iterated in some non arbitrary order.
I created following `allow-only` and `blacklist`:
```
"blacklist": {
  "enabled": true,
  "identity": [
    "0x9acce8db6bdbcae218be2ecf15175ca3c5f227fa"
  ],
  "certified": [
    "cefd51ab9b18986ad11870f2f0e54418d8bbf425b36f30420027bcf3f3d2e11e8533b939272308ede90f57df9242a0d5463f07fdb2199417a4b2c56c9ac00114"
  ]
},
"allow-only": {
  "enabled": true,
  "identity": [
    "0x9acce8db6bdbcae218be2ecf15175ca3c5f227fa"
  ],
  "certified": [
    "cefd51ab9b18986ad11870f2f0e54418d8bbf425b36f30420027bcf3f3d2e11e8533b939272308ede90f57df9242a0d5463f07fdb2199417a4b2c56c9ac00114"
  ]
}
```
Will the negotiation result depend only on negotiators order?

agent/provider/src/cli/rule.rs
```diff
@@ -1,30 +1,64 @@
...
+/// Left for compatibility only. Should be replaced by AddRule and RemoveRule.
 #[derive(StructOpt, Clone, Debug)]
 pub enum SetRule {
     Outbound(SetOutboundRule),
 }
 
+#[derive(StructOpt, Clone, Debug)]
+pub enum RuleCategory {
+    Outbound,
+    Blacklist,
```
pwalski, line 44: Naming it `Blocklist` would not help avoiding confusion with `whitelist` command?

agent/provider/src/cli/rule.rs
```diff
@@ -1,30 +1,64 @@
...
+    Outbound(SetOutboundRule),
+    Blacklist(RestrictRuleDesc),
+    AllowOnly(RestrictRuleDesc),
+}
+
+#[derive(StructOpt, Clone, Debug)]
+pub enum RemoveRule {
+    Outbound(SetOutboundRule),
+    Blacklist(RestrictRuleDesc),
+    AllowOnly(RestrictRuleDesc),
```
pwalski, line 59: Maybe there should be some comment/help message indicating a difference between `Outbound` and `Blacklist` `AllowOnly`. Something small indicating that `Outbound` is a more specific rule about nodes in context of Outbound network feature, and `Blacklist` `AllowOnly` are more general rules about allowing/blocking certain nodes.
To be fair I have no idea how to describe it.

agent/provider/src/cli/rule.rs
```diff
@@ -37,6 +71,15 @@ pub enum SetOutboundRule {
...
 }
 
+#[derive(StructOpt, Clone, Debug)]
+pub enum RestrictRuleDesc {
+    ByNodeId {
+        #[structopt(short, long)]
+        address: NodeId,
+    },
+    Certified(RestrictRuleWithCert),
+}
```
pwalski, line 81: Maybe we could want to unify it, because now it looks like this:
`rule add allow-only certified cert-id  cefd51ab`
`rule add allow-only certified import-cert ../golem-certificate/cert.new_key.signed.json`
but for `by-node-id`
`rule add allow-only by-node-id --address  0x9acce8db6bdbcae218be2ecf15175ca3c5f227fa`

Maybe it would be better to have
`rule add allow-only certified --cert-id  cefd51ab`
`rule add allow-only certified --cert-file ../golem-certificate/cert.new_key.signed.json`
?

pwalski:
In previous comments I have been adding `allow-only` and `blacklist` rules using this sample cert [cert.new_key.signed.json](https://github.com/golemfactory/yagna/files/14849853/cert.new_key.signed.json)
I tried to `rule set outbound audited-payload import-cert cert.new_key.signed.json --mode all` and now `rule list` shows different `certificate` column values for `allow-only`/`blacklist` and `coutbound.audited-payload`:
```
{
  "outbound": {
    "enabled": true,
    "everyone": "whitelist",
    "audited-payload": {
      "9a5bf07573dc7862620910f47349f84a41456da0b08b06ac349e0b901c0f6fb399856aea3c94cda8fa6bf56e7fb025e55c579c5cb13861d467ee8dd0f3c0af7c": {
        "mode": "all",
        "description": ""
      }
    },
    "partner": {
      "d40b0fdcc0df70efc5233f95422b7a9c1d7af87acad4bd5dc4ea83f5c9c9ab1e74103738109471e649a5d3d55faabdf35200f84760aea576105fd43823ec0edf": {
        "mode": "all",
        "description": ""
      }
    }
  },
  "blacklist": {
    "enabled": true,
    "identity": [
      "0x9acce8db6bdbcae218be2ecf15175ca3c5f227fa"
    ],
    "certified": [
      "cefd51ab9b18986ad11870f2f0e54418d8bbf425b36f30420027bcf3f3d2e11e8533b939272308ede90f57df9242a0d5463f07fdb2199417a4b2c56c9ac00114"
    ]
  },
  "allow-only": {
    "enabled": true,
    "identity": [
      "0x9acce8db6bdbcae218be2ecf15175ca3c5f227fa"
    ],
    "certified": [
      "cefd51ab9b18986ad11870f2f0e54418d8bbf425b36f30420027bcf3f3d2e11e8533b939272308ede90f57df9242a0d5463f07fdb2199417a4b2c56c9ac00114"
    ]
  }
}
```
In non json format I get `cerificate` column value `cefd51ab` for `allow-list`/`blacklist` and `9a5bf075` for `coutbound.audited-payload`.

Value  `9a5bf075` is a same value as `ID` column from `keystore list`.

nieznanysprawiciel:
In `rule list` merge node and certificate column into `id` columnd

agent/provider/src/cli/rule.rs
nieznanysprawiciel: I will try to get rid of `--address`

agent/provider/src/cli/rule.rs
nieznanysprawiciel: Documentation will clear this up. Issue: https://github.com/golemfactory/yagna/issues/3137

PR #3107, opened 2024-03-14
/activity api for list of jobs

staszek-krotki:
Resolves: https://github.com/golemfactory/ya-runtime-ai/issues/2

### DIFF
--- a/core/activity/src/api.rs
+++ b/core/activity/src/api.rs
@@ -22,7 +22,7 @@ mod common {
     use actix_web::{web, HttpResponse, Responder};
     use futures::prelude::*;
 
-    use ya_client_model::market::{Agreement, Role};
+    use ya_client_model::market::Role;
     use ya_core_model::{activity, NodeId};
     use ya_persistence::executor::DbExecutor;
     use ya_service_api_web::middleware::Identity;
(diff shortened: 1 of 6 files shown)


staszek-krotki:
Resolves https://github.com/golemfactory/ya-runtime-ai/issues/2

core/activity/src/api.rs
```diff
@@ -37,17 +37,32 @@ mod common {
...
-    // }
+    #[actix_web::get("/activity")]
+    async fn get_activities_web(db: web::Data<DbExecutor>, id: Identity) -> impl Responder {
+        log::debug!("get_activities_web");
+
+        let mut a = vec![];
+        if let Ok(activities) = get_activities(&db).await {
+            for activity in activities {
+                if let Ok(agreement_id) = get_agreement_id(&db, activity.as_str()).await {
+                    if authorize_agreement_executor(id.identity, &agreement_id, Role::Provider)
```
nieznanysprawiciel, line 55: It's problematic, because this endpoint is for both Provider and Requestor, but you check only Provider here.

core/payment/src/api/invoices.rs
```diff
@@ -82,8 +82,41 @@ async fn get_invoice(
...
+
+    let dao: InvoiceDao = db.as_dao();
+    match dao.get(invoice_id, node_id).await {
+        Ok(Some(invoice)) => {
+            let invoice_payments: Vec<Signed<Payment>> = payments
+                .iter()
+                .filter(|p| {
+                    let l1 = p.payload.activity_payments.len();
+                    let l2 = p.payload.agreement_payments.len();
+                    l1 != l2
```
nieznanysprawiciel, line 110: I don't understand this condition.
And I don't see any filter based on Invoice

core/activity/src/common.rs
```diff
@@ -73,8 +73,14 @@ pub(crate) fn generate_id() -> String {
     Uuid::new_v4().to_simple().to_string()
 }
 
-pub(crate) async fn _get_activities(db: &DbExecutor) -> Result<Vec<String>, Error> {
-    Ok(db.as_dao::<ActivityDao>()._get_activity_ids().await?)
+pub(crate) async fn get_activities(
+    db: &DbExecutor,
+    max_items: Option<u32>,
```
nieznanysprawiciel, line 78: It's not good idea to have `max_items` without `after_timestamp`. In practice it is unusable, because have no way to get older items when you exceeded `max_items`.

core/activity/src/dao/activity.rs
```diff
@@ -114,13 +114,23 @@ impl<'c> ActivityDao<'c> {
...
+                    .limit(items.into())
+                    .select(dsl::natural_id)
+                    .get_results(conn)
+                    .map_err(|e| e.into())
+            } else {
+                dsl::activity
+                    .select(dsl::natural_id)
+                    .get_results(conn)
+                    .map_err(|e| e.into())
+            }
```
nieznanysprawiciel, line 133: Here is example how to add additional filters to query dynamically to get rid of duplication and avoid compilation problems: https://github.com/golemfactory/yagna/blob/dc16197f8037b404cea9d5ad97e0782ff21bff17/core/payment/src/dao/payment.rs#L227-L243
In short: you must use `.into_boxed()` on the first query

core/activity/src/api.rs
```diff
@@ -22,32 +22,67 @@ mod common {
...
+        id: Identity,
+    ) -> impl Responder {
+        let mut result = vec![];
+        let activities = get_activities(&db, query.max_items).await?;
+
+        let is_valid = |agreement: Agreement| {
+            if authorize_caller(&id.identity.to_string().parse()?, agreement.provider_id()).is_ok()
+                && (query.after_timestamp.is_none()
+                    || query.after_timestamp.is_some()
+                        && agreement.timestamp >= query.after_timestamp.unwrap())
```
nieznanysprawiciel, line 61: As I commented below it is better to move timestamp filtering to database in this place

core/activity/src/api.rs
```diff
@@ -22,32 +22,67 @@ mod common {
...
+    async fn get_activities_web(
+        db: web::Data<DbExecutor>,
+        query: web::Query<FilterParams>,
+        id: Identity,
+    ) -> impl Responder {
+        let mut result = vec![];
+        let activities = get_activities(&db, query.max_items).await?;
+
+        let is_valid = |agreement: Agreement| {
+            if authorize_caller(&id.identity.to_string().parse()?, agreement.provider_id()).is_ok()
```
nieznanysprawiciel, line 58: If you take Agreement as Provider you want to authorize caller with provider if you take as requestor you should authorize requestor id. Probably it should better to use `authorize_agreement_initiator` and `authorize_agreement_executor`.

nieznanysprawiciel (changes requested):

core/activity/src/api.rs
```diff
@@ -22,32 +22,67 @@ mod common {
...
+                if is_valid(agreement)? {
+                    result.push(activity);
+                }
+            } else if let Ok(agreement) =
+                get_agreement(agreement_id.as_str(), Role::Requestor).await
+            {
+                if is_valid(agreement)? {
+                    result.push(activity);
+                }
+            }
```
nieznanysprawiciel, line 82: This code looks confusing, because you first query agreement in 2 ways and later do the same with this agreement independently in what case you are in. This suggest that you should rather first query agreement and save it in variable and later call the code for processing it.

nieznanysprawiciel (approved):
We should document this new endpoint later in `ya-client`

PR #3108, opened 2024-03-15
gsb http proxy monitor

pwalski:
Resolves: https://github.com/golemfactory/ya-runtime-ai/issues/56


exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
```diff
@@ -51,11 +54,12 @@ impl GsbToHttpProxy {
             builder = headers::add(builder, message.headers);
 
             log::debug!("Calling {}", &url);
+            let response_monitor = requests_monitor.on_request().await;
             let response = builder.send().await;
             let response =
                 response.map_err(|e| GsbToHttpProxyError::ErrorInResponse(e.to_string()))?;
-
             let bytes = response.bytes().await.unwrap();
+            response_monitor.on_response();
```
pwalski, line 62: I did not want to drive request logic through some passing request function into `response_monitor`, and instead I just mark when to start billing and when to stop. If `on_response` is not called directly it will be on `Drop` (to make it work like lets say `Mutex`)

utils/cli/src/table.rs
```diff
@@ -4,8 +4,8 @@ use std::collections::HashMap;
 
 pub fn print_table(
     columns: &[String],
-    values: &Vec<serde_json::Value>,
-    summary: &Vec<serde_json::Value>,
+    values: &[serde_json::Value],
+    summary: &[serde_json::Value],
```
pwalski, line 8: Initially I wanted to update Rust to `1.76` but Clippy produced too many issues.
I missed reverting this one, but it does no harm.

exe-unit/components/gsb-http-proxy/src/monitor.rs
```diff
@@ -0,0 +1,11 @@
...
+    /// Called once on every HTTP request
+    #[allow(async_fn_in_trait)]
+    async fn on_request(&mut self) -> impl ResponseMonitor;
+}
+
+pub trait ResponseMonitor: Sync + Send {
+    /// Called once on HTTP response (for any response code) or on Drop.
+    #[allow(async_fn_in_trait)]
+    fn on_response(self);
+}
```
nieznanysprawiciel, line 11: As I commented in https://github.com/golemfactory/ya-runtime-ai/pull/75 implementations in this library should include more than just unimplemented abstractions

exe-unit/components/gsb-http-proxy/Cargo.toml
```diff
@@ -21,4 +21,4 @@ actix-web = "4"
 actix-http = "3"
 mockito = "1.2.0"
 futures-core = "0.3.30"
-serde_derive = "1.0.195"
\ No newline at end of file
+serde_derive = "1.0.195"
```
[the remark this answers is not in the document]
pwalski: There were incremental versions everywhere. Removed them all.

exe-unit/components/gsb-http-proxy/src/monitor.rs
pwalski: This "more" could only mean an implementation of some "Counter" trait, which does not exist, and there is no obvious place to put it in. `ya-agreement-utils`? `ya-runtime-api`?
I just noticed that _Counters_ are  sometimes called _Metrics_. AI runtime initially had `duration_sec` implemented directly in main loop, without any dedicated struct or module, so I created one, but apparently it already existed as a [`TimeMETRIC`](https://github.com/golemfactory/yagna/blob/9fc18e4a10d4e4b69b8d4c4138a768cc0283b68d/exe-unit/src/metrics/mod.rs#L89).
If we want to have Counters as an API of a library then we should start by making public exe-unit's `Metric` trait together with some reusable implementations (like `duration_sec`).

PR #3127, opened 2024-03-26
counters crate

pwalski:
Resolves: https://github.com/golemfactory/ya-runtime-ai/issues/56
Resolves: https://github.com/golemfactory/ya-runtime-ai/issues/57


exe-unit/src/counters/os/unix.rs
```diff
@@ -101,3 +104,9 @@ impl Metrics {
             })
     }
 }
+
+impl From<SystemError> for MetricError {
+    fn from(err: SystemError) -> Self {
+        Self::Other(err.to_string())
```
pwalski, line 110: 1. `unix.rs` is still inside `exe-unit`. I will move it to `ya-counters` later - I wanted to have _custom_ implementation of `Metrics` outside of `ya-counters` before implementing `Metrics` for `ya-runtime-ai`.
2. I did not wanted to have generics in `MetricError` so I just translate it to `Other`.

exe-unit/src/service/counters.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#[allow(unused_imports)]
+use std::time::Duration;
+
+pub fn build(
+    ctx: &ExeUnitContext,
+    backlog_limit: Option<usize>,
+    supervise_caps: bool,
+) -> MetricsService {
+    let mut builder = MetricsServiceBuilder::new(ctx.agreement.usage_vector.clone(), backlog_limit);
```
pwalski, line 18: `MetricsServiceBuilder` comes from `ya-counters`

exe-unit/components/gsb-http-proxy/src/monitor.rs
```diff
@@ -0,0 +1,11 @@
+pub trait RequestsMonitor: Sync + Send + Clone {
```
pwalski, line 1: It will gone (or at least it will not be public), because Gsb Http Proxy service will hold handle to `MestricsService` (I will rename it to `Counters`), and update it by itself.

exe-unit/components/counters/src/counters.rs
```diff
@@ -0,0 +1,49 @@
+use std::fmt::Debug;
+use std::time::SystemTime;
+
+use crate::error::{self, MetricError};
+
+pub type Result<T> = std::result::Result<T, error::MetricError>;
+pub type MetricData = f64;
+
+#[derive(Clone, Debug)]
+pub enum MetricReport {
```
pwalski, line 10: It will be all renamed to `Counters`. For now I just use old name.

exe-unit/components/gsb-http-proxy/src/counters.rs
```diff
@@ -0,0 +1,183 @@
...
+use chrono::{DateTime, Duration, Utc};
+
+use ya_counters::error::CounterError;
+use ya_counters::Result as CountersResult;
+use ya_counters::{Counter, CounterData};
+
+#[derive(Default, Clone, Debug)]
+pub(super) struct Counters {
+    requests: Option<RequestCounter>,
+    requests_duration: Option<SharedRequestsDurationCounter>,
```
pwalski, line 15: I did not tried to handle them as some `Vec`, because for now there are only two counters, and there is a difference between them - only `requests_duration` counter cares about response

exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
```diff
@@ -76,35 +90,64 @@ impl GsbToHttpProxy {
...
         Box::pin(stream)
     }
+
+    pub fn requests_counter(&mut self) -> impl Counter {
+        self.counters.requests_counter()
+    }
+
+    pub fn requests_duration_counter(&mut self) -> impl Counter {
+        self.counters.requests_duration_counter()
+    }
```
pwalski, line 100: Counters are exposed only as some `impl Counter` Instead of `Monitor` trait implementations.

exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
```diff
@@ -114,5 +157,60 @@ mod tests {
         }
 
         assert_eq!(vec!["response".as_bytes()], v);
+        assert_eq!(1.0, requests_counter.frame().unwrap());
+        assert!(requests_duration_counter.frame().unwrap() > 0.0);
+    }
+
+    #[actix_web::test]
+    async fn multiple_parallel_requests() {
```
pwalski, line 165: technically `concurrent`, not `parallel`

exe-unit/components/gsb-http-proxy/src/counters.rs
```diff
@@ -0,0 +1,183 @@
...
+        self.requests_duration
+            .get_or_insert_with(Default::default)
+            .clone()
+    }
+
+    pub(super) fn on_request(&mut self) -> ResponseHandler {
+        if let Some(counter) = &mut self.requests {
+            counter.on_request();
+        }
+        if let Some(counter) = &mut self.requests_duration {
```
nieznanysprawiciel, line 33: You could use `requests_counter` and `requests_duration_counter` to avoid optional checks later

exe-unit/components/gsb-http-proxy/src/counters.rs
```diff
@@ -0,0 +1,183 @@
...
+            counter.on_request();
+            let counter = Some(counter.clone());
+            return ResponseHandler {
+                counter,
+                ..Default::default()
+            };
```
nieznanysprawiciel, line 39:
```suggestion
            return ResponseHandler {
                counter: Some(counter.clone()),
                ..Default::default()
            };
```

nieznanysprawiciel (commented):
Check this `expect` usage. The rest is ok

exe-unit/components/gsb-http-proxy/src/counters.rs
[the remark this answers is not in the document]
pwalski: gsb_to_http proxy does not force user to use counters, so they are both optional (because only one can be used).  I can remove this `Option`s.

exe-unit/components/gsb-http-proxy/src/counters.rs
pwalski: `requests_counter` and `requests_duration_counter` are public functions to initialize and clone internal counters only to return `impl Count` from `gsb_http_proxy`. I need `RequestCounter` and `RequestDurationCounter` types.

exe-unit/components/gsb-http-proxy/src/counters.rs
pwalski: ok (for some reason I just do not like to explicitly type struct field names)

PR #3132, opened 2024-03-27
Kek/lock rework

kamirr:


core/payment/src/processor.rs
```diff
@@ -290,58 +295,91 @@ impl DriverRegistry {
...
+    db_executor: Mutex<DbExecutor>,
+    registry: RwLock<DriverRegistry>,
+    in_shutdown: AtomicBool,
 }
 
 impl PaymentProcessor {
     pub fn new(db_executor: DbExecutor) -> Self {
         Self {
-            db_executor,
+            db_executor: Mutex::new(db_executor),
```
nieznanysprawiciel, line 310: You assume that `DbExecutor` wasn't cloned, before this function was called.
This mutex protects db from being used in multiple places at the same time by processor, but not in general case

core/payment/src/processor.rs
```diff
@@ -743,8 +820,17 @@ impl PaymentProcessor {
...
     /// When `bool` is `true` all existing allocations are released immediately.
     /// For `false` each allocation timestamp is respected.
-    pub async fn release_allocations(&mut self, force: bool) {
-        let db = Data::new(self.db_executor.clone());
+    pub async fn release_allocations(&self, force: bool) {
+        // keep this lock alive for the entirety of this function for now
+        let db_executor = match self.db_executor.timeout_lock(DB_LOCK_TIMEOUT).await {
+            Ok(db) => db,
+            Err(_) => {
+                log::error!("Timed out waiting for db lock");
```
nieznanysprawiciel, line 828: maybe `inspect_err`?

nieznanysprawiciel (approved):

PR #3134, opened 2024-03-27
Deposits

scx1332:


utils/networking/src/resolver.rs
```diff
@@ -71,12 +71,10 @@ pub async fn try_resolve_dns_record(request_url_or_host: &str) -> String {
...
-    match {
-        if SCHEME_RE.is_match(request_url_or_host) {
-            resolve_dns_record(request_url_or_host).await
-        } else {
-            resolve_dns_record_host(request_url_or_host).await
-        }
+    match if SCHEME_RE.is_match(request_url_or_host) {
+        resolve_dns_record(request_url_or_host).await
+    } else {
+        resolve_dns_record_host(request_url_or_host).await
```
kamirr, line 77: can you extract the `if { .. } else { .. }` into a variable? It's not reallly readable without the previously present brackets

exe-unit/examples/transfer.rs
```diff
@@ -28,6 +28,7 @@ fn create_file(path: &Path, name: &str, chunk_size: usize, chunk_count: usize) -
     let mut file_src = OpenOptions::new()
         .write(true)
         .create(true)
+        .truncate(false)
```
kamirr, line 31: why? new clippy lint?

scx1332: yes

PR #3136, opened 2024-03-27
eventtype => eventType

staszek-krotki:


pwalski (approved):
lgtm, `ya-client` commit from `Cargo.toml` has `Timeout` with nested `timeout` field
https://github.com/golemfactory/ya-client/blob/87188a3f274ebc963debcccc6ae399e98e471c72/model/src/payment/params.rs#L41

scx1332:
Reverted change in Timeout in ya-client. Already bumped version in other pull request and closing this one. If you want to change Timeout struct in ya-client please make sure to not break API without good cause.

PR #3155, opened 2024-04-05
Grafana metrics grouping

nieznanysprawiciel:
resolves: https://github.com/golemfactory/gamerhash-facade/issues/101


core/metrics/src/pusher.rs
```diff
@@ -111,57 +118,127 @@ async fn try_get_default_id() -> anyhow::Result<IdentityInfo> {
...
+
+    #[test]
+    fn test_get_push_url_with_single_label() {
+        let ctx = MetricsCtx {
+            job: "community.1".into(),
+            labels: labels(&[("group", "random")]),
+            ..MetricsCtx::default()
+        };
+        let url = get_push_url("http://a", &default_id_info(), &ctx).unwrap();
+        assert_eq!("http://a/metrics/job/community.1/instance/0x0000000000000000000000000000000000000000/hostname/node1/group/random", url);
```
pwalski, line 216: I assume this `community.1` is a reason you remove some chars here 
https://github.com/golemfactory/yagna/pull/3155/files#diff-28888f6916f83521f834cc9d693aa6412458f751628bcb02dda30600ef5e38c5R12

core/metrics/src/pusher.rs
```diff
@@ -1,12 +1,17 @@
...
+use percent_encoding::{utf8_percent_encode, AsciiSet, NON_ALPHANUMERIC};
 use tokio::time::{self, Duration, Instant};
 
-use percent_encoding::{utf8_percent_encode, NON_ALPHANUMERIC};
 use ya_core_model::identity::{self, IdentityInfo};
 use ya_service_api::MetricsCtx;
 use ya_service_bus::typed as bus;
 
+lazy_static! {
+    static ref SAFE_CHAR_SET: AsciiSet = NON_ALPHANUMERIC.remove(b'.').remove(b'-').remove(b'_');
```
pwalski, line 12: This characters are rather an `UNSAFE_CHAR_SET`, because you want to encode them

core/metrics/src/pusher.rs
nieznanysprawiciel: Community was working correctly earlier and wasn't checked at all. But I wanted to have ability to use labels with special characters and they were looking really ugly in Grafana

core/metrics/src/pusher.rs
nieznanysprawiciel: Changed

PR #3164, opened 2024-04-08
gsb http proxy non streaming + self binding

staszek-krotki:
- `http_proxy_request` can now accept both streaming and non-streaming requests (controlled by accept header: text/event-stream)
- `gsb-http-proxy` component test for happy path
- `gsb-http-proxy` component binds itself to gsb


core/activity/src/http_proxy.rs
```diff
@@ -74,8 +80,32 @@ async fn proxy_http_request(
...
     Ok(HttpResponse::Ok()
         .keep_alive()
         .content_type(mime::TEXT_EVENT_STREAM.essence_str())
         .streaming(stream))
 }
+
+async fn await_results(
+    mut stream: impl Stream<Item = Result<Bytes, Error>> + Unpin,
+) -> crate::Result<impl Responder> {
+    let response = stream.next().await;
```
pwalski, line 103: Shouldn't we at least check if consecutive `stream.next()` will return `None` and handle `Some` somehow? (error) ?

core/activity/src/http_proxy.rs
```diff
@@ -74,8 +80,32 @@ async fn proxy_http_request(
             .call_streaming(msg)
     });
 
+    if let Some(value) = request.headers().get(header::ACCEPT) {
+        if value.eq(mime::TEXT_EVENT_STREAM.essence_str()) {
```
pwalski, line 84: Could it be an `APPLICATION_OCTET_STREAM`?

core/activity/src/http_proxy.rs
```diff
@@ -74,8 +80,32 @@ async fn proxy_http_request(
...
         .streaming(stream))
 }
+
+async fn await_results(
+    mut stream: impl Stream<Item = Result<Bytes, Error>> + Unpin,
+) -> crate::Result<impl Responder> {
+    let response = stream.next().await;
+
+    if let Some(Ok(bytes)) = response {
+        let response_body = String::from_utf8(bytes.to_vec())
```
nieznanysprawiciel, line 106: Why do you assume that response will be encoded as utf8?

exe-unit/components/gsb-http-proxy/tests/test_proxy.rs
```diff
@@ -0,0 +1,98 @@
...
+use ya_framework_basic::log::enable_logs;
+use ya_gsb_http_proxy::http_to_gsb::HttpToGsbProxy;
+use ya_gsb_http_proxy::message::GsbHttpCallMessage;
+use ya_gsb_http_proxy::response::GsbHttpCallResponseEvent;
+use ya_service_bus::typed as bus;
+
+#[test_context(DroppableTestContext)]
+#[serial_test::serial]
+pub async fn test_gsb_http_proxy(ctx: &mut DroppableTestContext) {
+    enable_logs(true);
```
nieznanysprawiciel, line 17: Should be disabled after development is finished
```suggestion
    enable_logs(false);
```

exe-unit/components/gsb-http-proxy/tests/test_proxy.rs
```diff
@@ -0,0 +1,98 @@
...
+            headers: HeaderMap::default(),
+        };
+
+        let mut stream = http_to_gsb
+            .pass(move |msg| bus::service(ya_gsb_http_proxy::BUS_ID).call_streaming(msg));
+        if let Ok(_r) = stream.next().await.unwrap() {
+            if let Ok(s) = String::from_utf8(_r.to_vec()) {
+                return s;
+            }
+        }
```
nieznanysprawiciel, line 46: I know that it was already implemented like this earlier, but I spent a lot of time trying to understand this test and I think the interface of this proxy is opposite of what user would expect:
I would expect that the parameters that I give to `HttpToGsbProxy` struct would be actually passed to `pass` function.

So from my perspective the interface could look like this:
- I create proxy object and pass target parameters like destination NodeId, address on destination and so on
    - Some of them might be optional some necessary, this depends
- I call `proxy.pass(...)` method and provide parameters of http which should be proxied.

Consumer of this proxy object shouldn't be forced to pass function that implements gsb call.

This can be for later

exe-unit/components/gsb-http-proxy/tests/test_proxy.rs
```diff
@@ -0,0 +1,98 @@
...
+use test_context::test_context;
+use ya_framework_basic::async_drop::DroppableTestContext;
+use ya_framework_basic::log::enable_logs;
+use ya_gsb_http_proxy::http_to_gsb::HttpToGsbProxy;
+use ya_gsb_http_proxy::message::GsbHttpCallMessage;
+use ya_gsb_http_proxy::response::GsbHttpCallResponseEvent;
+use ya_service_bus::typed as bus;
+
+#[test_context(DroppableTestContext)]
+#[serial_test::serial]
```
nieznanysprawiciel, line 15: It should be marked as framework-test. We had agreement with Rekuc that tests touching system shouldn't be ran as basic `cargo test`
```suggestion
#[cfg_attr(not(feature = "framework-test"), ignore)]
#[test_context(DroppableTestContext)]
#[serial_test::serial]
```

exe-unit/components/gsb-http-proxy/tests/test_proxy.rs
```diff
@@ -0,0 +1,98 @@
...
+                .expect("internal call should succeed");
+            let response_text = response.text().await.expect("text expected");
+            let response = GsbHttpCallResponseEvent {
+                index: 0,
+                timestamp: Utc::now().naive_local().to_string(),
+                msg_bytes: response_text.into_bytes(),
+            };
+            Ok(response)
+        },
+    );
```
nieznanysprawiciel, line 97: As discussed earlier, this could be part of library functionality, to bind to gsb.

Can be done later

core/activity/src/http_proxy.rs
staszek-krotki: I wasn't sure if it is better to have an error in this case or just assume the first streamed value is OK

core/activity/src/http_proxy.rs
staszek-krotki: I still heaven't created a full-featured headers support. I guess I would need to parse the Content-Type from the response and see if it is any different than Utf-8. This will come later

exe-unit/components/gsb-http-proxy/tests/test_proxy.rs
staszek-krotki: I agree - once the component will take care of gsb binding this will make more sense.

core/activity/src/http_proxy.rs
staszek-krotki: added

core/activity/src/http_proxy.rs
pwalski: No idea. Lets leave it as is.

nieznanysprawiciel (approved):

pwalski (approved):

Issue #3167, opened 2024-04-11
Create a `JobObject` when spawning a runtime on Windows


pwalski:
> `ya-provider` does not support any process groups / jobs on Windows
> 
> https://github.com/golemfactory/yagna/blob/f189daee32819d0b44d3a6809e2ef4f617106042/utils/process/src/lib.rs#L66
> 
> 
> 
> Maybe we should move Windows process Job related code from `ya-runtime-ai` to some lib and use it in `ya-provider`?
> 
> https://github.com/golemfactory/ya-runtime-ai/blob/d5683703359165543c116e7c55e679a71a7e2c74/src/main.rs#L290
> 
> 
> 
> It could be moved into dedicated crate in lets say `exe-unit/components/proc` together with this module https://github.com/golemfactory/yagna/tree/staszek/gamerhash-combined/exe-unit/components/counters/src/os/process because it always looked odd there (and `ya-counters` will use `ya-proc` when `os` feature id enabled). 

 _Originally posted by @pwalski in [#108](https://github.com/golemfactory/gamerhash-facade/issues/108#issuecomment-2049252544)_

Implementing it will allow to avoid having a remaining `ya-runtime-ai` process when `ya-provider` gets killed but GH facade stays alive.

PR #3168, opened 2024-04-11
Shared exe-unit process module

pwalski:
Resolves: https://github.com/golemfactory/yagna/issues/3167


exe-unit/components/counters/Cargo.toml
```diff
@@ -6,6 +6,7 @@ description = "Runtime counters generating data used to collect activity metrics
 
 [dependencies]
 ya-agreement-utils = { workspace = true }
+ya-utils-process = { path = "../../../utils/process", optional = true }
```
pwalski, line 9: I am not sure how to resolve this dependency without publishing it to crates.io

utils/process/src/lib.rs
```diff
@@ -80,12 +100,21 @@ pub enum ExeUnitExitStatus {
...
+    #[allow(unused)]
+    #[cfg(windows)]
+    job_object: JobObject,
 }
 
 impl ProcessHandle {
     pub fn new(command: &mut Command) -> Result<ProcessHandle> {
+        let process = Arc::new(SharedChild::spawn(command)?);
+        #[cfg(windows)]
+        let job_object = JobObject::try_new_current()?;
```
pwalski, line 113: I left `new_process_group` unimplemented for Windows. `process_group` is roughly equivalent of Windows JobObject, but `new_process_group` is implemented for `Command` and I needed to store `job_object` somewhere, I could not extend `Command`, so instead I just crate `job_object` while creating `ProcessHandle`.

utils/process/src/lib.rs
```diff
@@ -80,12 +100,21 @@ pub enum ExeUnitExitStatus {
 #[derive(Clone)]
 pub struct ProcessHandle {
     process: Arc<SharedChild>,
+    // Windows process will be killed once `job_object` is dropped
+    #[allow(unused)]
+    #[cfg(windows)]
+    job_object: JobObject,
```
nieznanysprawiciel, line 106: Having conditionally existing field is hard to maintain. I'm wondering if we could have empty job object implementation for unix?
In general all these conditional compilation in process handling is hard to maintain, because we need to compile with different features to be sure it will compile... But it was already done like this before, so we can leave it as is if it would require to much work.

nieznanysprawiciel (approved):

PR #3179, opened 2024-04-16
Scx1332/goth

scx1332:


goth_tests/helpers/payment.py
```diff
@@ -120,16 +120,46 @@ class AllocationCtx:
...
+                        make_deposit=True,
+                        timestamp=datetime.now(timezone.utc),
+                        timeout=datetime.now(timezone.utc) + timedelta(minutes=30),
+                        payment_platform=self.requestor.payment_config.platform_string,
+                        deposit=global_deposit,
+                    )
+                    allocation = await self.requestor.api.payment.create_allocation(allocation_arg)
+                except Exception as ex:
+                    logger.warning("Failed to create allocation for deposit {} - {}".format(global_deposit["id"], ex))
+                    continue
```
kamirr, line 143: Why iterate over all deposits if you only care about the allocation for the last one?

core/payment-driver/erc20/src/service.rs
```diff
@@ -181,6 +184,21 @@ impl Erc20Service {
...
+                            );
+                            chain.lock_contract = Some(LockContractSettings { address: parsed })
+                        }
+                        Err(e) => {
+                            log::warn!(
+                                "Value {lock_payment_addr} for {lock_payment_addr_env} is not valid H160 address: {e}"
+                            );
+                        }
+                    };
+                }
```
kamirr, line 201: We should remove a requirement for hard-coding the lock-contract address at all at some point.

kamirr (changes requested):

PR #3194, opened 2024-04-24
payment: differentiate insufficient account / deposit funds

kamirr:


core/model/src/driver.rs
```diff
@@ -403,7 +403,8 @@ impl ValidateAllocation {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum ValidateAllocationResult {
-    InsufficientFunds,
+    InsufficientAccountFunds,
```
scx1332, line 406: Where is this enum posted?
InsufficientAccountFunds

scx1332 (changes requested):

scx1332 (approved):

PR #3195, opened 2024-04-24
Agent monitoring metrics

nieznanysprawiciel:


core/market/src/negotiation/provider.rs
```diff
@@ -197,6 +198,9 @@ impl ProviderBroker {
         timeout: f32,
         max_events: Option<i32>,
     ) -> Result<Vec<ProviderEvent>, QueryEventsError> {
+        // Thanks to this counter we can monitor agent activity.
+        counter!("market.events.provider.query", 1);
```
pwalski, line 202: It gets incremented by `1`, while `market.events.provider.queried` increments by `events.len()`.
I assume it is ok (?).

pwalski (approved):

core/market/src/negotiation/provider.rs
nieznanysprawiciel: Yes, this counter counts number of calls

PR #3211, opened 2024-05-07
Activities checkup on startup

pwalski:
Resolves: https://github.com/golemfactory/gamerhash-facade/issues/108

- Spawning activity monitor on startup for activities with not `Terminated` and not `Unresponsive` status.
- `monitor_activity` sets `Unresponsive` status when both _inactivity_ and _unresponsiveness_ limit has been reached
- tested in https://github.com/golemfactory/gamerhash-facade/pull/137


core/activity/src/provider/service.rs
```diff
@@ -74,7 +78,62 @@ pub fn bind_gsb(db: &DbExecutor, tracker: TrackerRef) {
...
+    let activity_state_dao = db.as_dao::<ActivityStateDao>();
+    for activity_id in activities_ids {
+        let activity_state = activity_state_dao.get(&activity_id).await?;
+        if activity_state.alive() {
+            log::info!("Initial activty {activity_id} monitoring");
+            tokio::task::spawn_local(monitor_activity(
+                db.clone(),
+                tracker.clone(),
+                activity_id,
+                agreement.provider_id,
```
pwalski, line 131: It is the reason I added `provider_id` (and `requestor_id`) to `AgreementListEntry`.
https://github.com/golemfactory/ya-client/pull/171/files
Without doing it I would need to do `ListAgreements`, and then for every agreement do `GetAgreement` call only to get provider_id.
@nieznanysprawiciel is it ok to change `ya-client`?

pwalski: No.

core/activity/src/provider/service.rs
```diff
@@ -74,7 +77,60 @@ pub fn bind_gsb(db: &DbExecutor, tracker: TrackerRef) {
...
+        if is_responsive(activity_state) {
+            log::info!("Initial activty {activity_id} monitoring");
+            tokio::task::spawn_local(monitor_activity(
+                db.clone(),
+                tracker.clone(),
+                activity_id,
+                *agreement.provider_id(),
+                None,
+            ));
+        }
```
pwalski, line 131: Initial spawn of `monitor_activity` changes not `Terminated` activities into `Unresponsive` activities, so it makes sense to spawn it only for activities which are not `Terminated` and not `Unresponsive` (`ya-client` provides `alive()` function, but for this usecase only I created more specific `is_responsive()` function. I did not wanted to update `ya-client` only to create this function)

core/activity/src/provider/service.rs
```diff
@@ -350,31 +406,41 @@ async fn monitor_activity(
...
+                        provider_id,
+                        app_session_id.clone(),
+                    )
                     .await;
-                if let Err(e) = set_persisted_state(&db, &activity_id, new_state).await {
-                    log::error!("cannot update activity {} state: {}", activity_id, e);
+
+                    counter!("activity.provider.destroyed.unresponsive", 1);
+                    break;
                 }
```
pwalski, line 442: The only real change is that the `Unresponsive` state will be set after reaching both `unresponsive_limit` or `inactivity_limit`.

core/activity/src/provider/service.rs
```diff
@@ -350,31 +406,41 @@ async fn monitor_activity(
...
-                    db,
-                    tracker.clone(),
-                    &activity_id,
-                    provider_id,
-                    app_session_id.clone(),
-                )
-                .await;
+            let duration = (Utc::now().timestamp() - usage.timestamp) as f64;
+
+            if duration > unresponsive_limit || duration > inactivity_limit {
```
nieznanysprawiciel, line 411: It would be more clear if we had only `duration > unresponsive_limit` check.
The second check should be always true since we should have `unresponsive_limit < inactivity_limit`.
If the relation is opposite that the whole code makes no sense, because we make activity responsive again, after we already added ActivityDestroy event

core/activity/src/provider/service.rs
```diff
@@ -350,31 +406,41 @@ async fn monitor_activity(
...
+                    )
                     .await;
-                if let Err(e) = set_persisted_state(&db, &activity_id, new_state).await {
-                    log::error!("cannot update activity {} state: {}", activity_id, e);
+
+                    counter!("activity.provider.destroyed.unresponsive", 1);
+                    break;
                 }
-            } else if state.state.0 == State::Unresponsive && dt < unresp_s {
+            } else if state.state.0 == State::Unresponsive && duration <= unresponsive_limit {
```
nieznanysprawiciel, line 443: I would make sure that panic below will never get triggered. Maybe it should be error and end of loop?

core/activity/src/provider/service.rs
```diff
@@ -350,31 +406,41 @@ async fn monitor_activity(
...
+                    log::warn!("activity {} unresponsive after {}s", activity_id, duration);
+                    let new_state =
+                        ActivityState::from(StatePair(State::Unresponsive, state.state.1));
+                    prev_state = Some(state);
+                    let _ = tracker
+                        .update_state(activity_id.clone(), State::Unresponsive)
+                        .await;
+                    if let Err(e) = set_persisted_state(&db, &activity_id, new_state).await {
+                        log::error!("cannot update activity {} state: {}", activity_id, e);
+                    }
```
nieznanysprawiciel, line 422: It would be good to track it. Note that it is different than counter below for purpose.
And please add counter initialization to 0.
```suggestion
                    }
                    counter!("activity.provider.unresponsive", 1);
```

nieznanysprawiciel: And it would be good to add 
`counter!("activity.provider.responsive-again", 1);`
in the if clause below

core/activity/src/provider/service.rs
```diff
@@ -350,31 +406,41 @@ async fn monitor_activity(
...
-                log::warn!("activity {} inactive for {}s, destroying", activity_id, dt);
-                enqueue_destroy_evt(
-                    db,
-                    tracker.clone(),
-                    &activity_id,
-                    provider_id,
-                    app_session_id.clone(),
-                )
-                .await;
+            let duration = (Utc::now().timestamp() - usage.timestamp) as f64;
```
nieznanysprawiciel, line 409: Maybe?
```suggestion
            let since_update = (Utc::now().timestamp() - usage.timestamp) as f64;
```

core/activity/src/provider/service.rs
pwalski: `unresponsive_limit` and `inactivity_limit` come from env variables, and I have found no code making sure that `unresponsive_limit < inactivity_limit`, which is why I wrote it this way.

> If the relation is opposite that the whole code makes no sense, because we make activity responsive again, after we already added ActivityDestroy event

Where it would happen?

core/activity/src/provider/service.rs
```diff
@@ -74,7 +77,60 @@ pub fn bind_gsb(db: &DbExecutor, tracker: TrackerRef) {
...
     counter!("activity.provider.destroyed.unresponsive", 0);
 
-    local::bind_gsb(db, tracker);
+    local::bind_gsb(&db.clone(), tracker.clone());
+
+    tokio::task::spawn_local(initial_checkup(db.clone(), tracker.clone()));
+}
+
+async fn initial_checkup(db: DbExecutor, tracker: TrackerRef) {
+    log::info!("Initial activities checkup");
```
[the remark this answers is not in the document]
pwalski: Changed.

core/activity/src/provider/service.rs
```diff
@@ -74,7 +77,60 @@ pub fn bind_gsb(db: &DbExecutor, tracker: TrackerRef) {
...
+        return;
+    };
+
+    for agreement in agreements {
+        let agreement_res = bus::service(market::BUS_ID)
+            .send(market::GetAgreement::as_provider(agreement.id))
+            .await;
+        let Ok(Ok(agreement)) = agreement_res else {
+            log::error!(
+                "Acitivties checkup error. Failed to get Agreement. Err: {agreement_res:?}"
```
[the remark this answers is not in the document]
pwalski: Changed.

core/activity/src/provider/service.rs
pwalski: Terrible naming. I assumed it is the Activity duration.

core/activity/src/provider/service.rs
```diff
@@ -74,7 +77,60 @@ pub fn bind_gsb(db: &DbExecutor, tracker: TrackerRef) {
...
+    local::bind_gsb(&db.clone(), tracker.clone());
+
+    tokio::task::spawn_local(initial_checkup(db.clone(), tracker.clone()));
+}
+
+async fn initial_checkup(db: DbExecutor, tracker: TrackerRef) {
+    log::info!("Initial activities checkup");
+    let agreements_res = get_agreements_by_state(AgreementState::Approved).await;
+    let Ok(agreements) = agreements_res else {
+        log::error!("Acitivties checkup error. Failed to list Agreements. Err: {agreements_res:?}");
```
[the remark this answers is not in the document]
pwalski: Changed.

core/activity/src/provider/service.rs
```diff
@@ -74,7 +77,60 @@ pub fn bind_gsb(db: &DbExecutor, tracker: TrackerRef) {
...
+    log::info!(
+        "Initial activities checkup for agreement {}",
+        agreement.agreement_id
+    );
+    let activities_ids = get_activities_for_agreement(&db, &agreement.agreement_id).await?;
+    let activity_state_dao = db.as_dao::<ActivityStateDao>();
+    for activity_id in activities_ids {
+        let activity_state = activity_state_dao.get(&activity_id).await?;
+        if is_responsive(activity_state) {
+            log::info!("Initial activty {activity_id} monitoring");
```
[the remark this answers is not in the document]
pwalski: Changed.

core/activity/src/provider/service.rs
pwalski: Changed.

core/activity/src/provider/service.rs
pwalski: Counters added.

core/activity/src/provider/service.rs
pwalski: Changed.

core/activity/src/provider/service.rs
pwalski: Changed.

nieznanysprawiciel (approved):

PR #3220, opened 2024-05-16
Deposit validation on contract

kamirr:


scx1332 (approved):

kamirr:
Will be merged after @pociej verifies that it works for his use-case

kamirr:
verified

PR #3239, opened 2024-06-10
no payment init in provider

kamirr:


agent/provider/src/provider_agent.rs
```diff
@@ -320,56 +317,32 @@ impl ProviderAgent {
...
+        if accounts.is_empty() {
+            anyhow::bail!(
+                "Payment account {} not initialized. Please run\n\
+                    \t`yagna payment init --receiver --network {} --account {}`\n\
+                    for all drivers you want to use.",
+                account,
+                networks[0].network,
+                account,
+            )
         }
```
nieznanysprawiciel, line 343: Is it still valid message?

agent/provider/src/startup_config.rs
```diff
@@ -153,8 +208,8 @@ pub struct ReceiverAccount {
     #[structopt(long, env = "YA_ACCOUNT")]
     pub account: Option<NodeId>,
     /// Payment network.
-    #[structopt(long = "payment-network", env = "YA_PAYMENT_NETWORK", possible_values = NetworkName::VARIANTS, default_value = NetworkName::Mainnet.into())]
-    pub networks: Vec<NetworkName>,
+    #[structopt(long = "payment-network", env = "YA_PAYMENT_NETWORK", default_value = NetworkName::Mainnet.into())]
```
nieznanysprawiciel, line 211: Is it possible to provide these `possible_values`?
They were very helpful.
If it isn't, than maybe at least description saying what can be passed here?

agent/provider/src/startup_config.rs
```diff
@@ -142,7 +143,61 @@ impl ProviderConfig {
     }
 }
 
-#[derive(StructOpt, Clone, Debug, Serialize, Deserialize, derive_more::Display)]
+#[derive(Clone, Debug)]
+pub struct NetworkOrPlatform {
+    pub driver: String,
```
nieznanysprawiciel, line 148: I'm wondering if we have typed structure somewhere?

agent/provider/src/provider_agent.rs
kamirr: No, it isn't. removed

agent/provider/src/startup_config.rs
kamirr: Unfortunately, no, it can't be generates as before, mostly due to API constraints of structopt. I did add a help message

agent/provider/src/startup_config.rs
kamirr: No*, but I'll replace this with a newtype around the `PaymentPlatform` when I'm done with the refactor.

*actually yes, but it's driver-specific implementation detail, and I intend to remove it as well

nieznanysprawiciel (approved):

PR #3270, opened 2024-07-12
core/identity/cli/identity.rs cleanup

dommyrock:
Minor code readability / cleanup .

### DIFF
--- a/core/identity/src/cli/identity.rs
+++ b/core/identity/src/cli/identity.rs
@@ -322,41 +322,29 @@ impl IdentityCommand {
                     log::warn!("Using private key directly is not recommended. Use keystore instead. Your key could leak in command history, check and clean logs.")
                 }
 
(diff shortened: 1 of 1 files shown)


prekucki (dismissed):

core/identity/src/cli/identity.rs
```diff
@@ -338,41 +338,29 @@ impl IdentityCommand {
...
-                } else {
-                    None
-                };
-
-                let key_file = if let Some(keystore) = from_keystore {
-                    std::fs::read_to_string(keystore)?
-                } else {
-                    let password = if *no_password {
-                        Protected::from("")
-                    } else if let Some(password) = password {
```
prekucki, line 362: You lost this case:

```
$ yagna id create --from-private-key ".." --password ".." alias
```

without `--password`  means that you will be asked for a password unless the `--no-password` flag is passed.

dommyrock: Yep, my mistake. Thanks for the head up!

PR #3280, opened 2024-07-26
Payments testing environment

nieznanysprawiciel:
Environment for testing standalone payment module with other parts of yagna mocked.
It doesn't require interactions with external yagna processes, because it can be instantiated fully in test.

Some payment examples were rewritten to tests and removed afterwards.

Additional fixes:
- Fix DebitNotes unique constraint fail, when Provider and Requestor is on the same node
- Fix deadlock in payment confirmation, when Provider and Requestor is on the same Node
- Fix LogErr trait to display logs


nieznanysprawiciel:
Force merging because it was already approved as: https://github.com/golemfactory/yagna/pull/3290

PR #3297, opened 2024-08-19
Do not lock when performing remote GSB call

mwalkiewicz:


core/payment/src/processor.rs
```diff
@@ -655,25 +655,24 @@ impl PaymentProcessor {
...
-                            payment_id.clone(),
-                            msg_with_bytes.signature.clone(),
-                            msg_with_bytes.signed_bytes.clone(),
-                        )
-                        .await
-                        .is_ok()
+                // Mutablility increases code readability - otherwise we'd have to take the lock
+                // and create payment_dao in every branch of the if-else below because we have to
+                // ensure that it is taken after the remote call performed only in some of the
+                // branches.
```
prawilny, line 664: Note there is a nice trick with setting multiple values at once conditionally (as blocks of code return values in Rust) - see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=76aeaf7adb0ca4f079fbd2e789c9d0b3 for an example

prawilny (approved):

prawilny (approved):

PR #3302, opened 2024-09-05
Initialize more termination reason counters

prawilny:
Interestingly enough, we encountered both `RequestorUnreachable` reported both by provider and requestor.
Even more interestingly, we encountered `ProviderUnreachable` reporterd only by provider.


nieznanysprawiciel (approved):
Approving, but this PR requires yagna to know something about termination reasons, what is against design assumptions, so we need to consider if this PR should ever be merged to master.

Other thing: uknown reasons will still be reported starting from 1, so we can encounter unexpected issues later.

PR #3304, opened 2024-09-06
Add logs for GSB http proxy

nieznanysprawiciel:
resolves: https://github.com/golemfactory/ya-runtime-ai/issues/105

### DIFF
--- a/core/activity/src/http_proxy.rs
+++ b/core/activity/src/http_proxy.rs
@@ -44,9 +44,12 @@ async fn proxy_http_request(
     let activity_id = path_activity_url.activity_id;
     let path = path_activity_url.url;
 
-    let result = authorize_activity_executor(&db, id.identity, &activity_id, Role::Requestor).await;
+    let result =
+        authorize_activity_initiator(&db, id.identity, &activity_id, Role::Requestor).await;
     if let Err(e) = result {
-        log::error!("Authorize error {}", e);
+        log::info!(
+            "Proxy authorize error (currently not authorized requests are not rejected): {e}"
+        );
     }
 
     let agreement = get_activity_agreement(&db, &activity_id, Role::Requestor).await?;
--- a/exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
+++ b/exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
@@ -53,9 +53,9 @@ impl GsbToHttpProxy {
 
     pub fn bind(&mut self, gsb_path: &str) -> Handle {
         let this = self.clone();
-        bus::bind(gsb_path, move |message: GsbHttpCallMessage| {
+        bus::bind_with_caller(gsb_path, move |caller, message: GsbHttpCallMessage| {
             let mut this = this.clone();
-            async move { Ok(this.pass(message).await) }
+            async move { Ok(this.pass(caller, message).await) }
         })
     }
 
@@ -67,10 +67,13 @@ impl GsbToHttpProxy {
         })
     }
 
-    pub async fn pass(&mut self, message: GsbHttpCallMessage) -> GsbHttpCallResponse {
+    pub async fn pass(
+        &mut self,
+        caller: String,
+        message: GsbHttpCallMessage,
+    ) -> GsbHttpCallResponse {
         let url = format!("{}{}", self.base_url, message.path);
-        log::info!("Gsb to http call - Url: {url}");
-
+        let path = message.path.clone();
         let mut counters = self.counters.clone();
 
         let method = match Method::from_bytes(message.method.to_uppercase().as_bytes()) {
@@ -82,9 +85,10 @@ impl GsbToHttpProxy {
                 )
             }
         };
-        let builder = Self::create_request_builder(method, &url, message.headers, message.body);
+        let builder =
+            Self::create_request_builder(method.clone(), &url, message.headers, message.body);
 
-        log::debug!("Calling {}", &url);
+        log::info!("Gsb proxy http call {method} to {url} from {caller}");
         let response_handler = counters.on_request();
         let response = builder
             .send()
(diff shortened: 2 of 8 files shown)


core/activity/src/http_proxy.rs
```diff
@@ -44,9 +44,12 @@ async fn proxy_http_request(
     let activity_id = path_activity_url.activity_id;
     let path = path_activity_url.url;
 
-    let result = authorize_activity_executor(&db, id.identity, &activity_id, Role::Requestor).await;
+    let result =
+        authorize_activity_initiator(&db, id.identity, &activity_id, Role::Requestor).await;
```
prawilny, line 48: Does this change alter semantics anyhow?

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
```diff
@@ -74,44 +83,52 @@ impl HttpToGsbProxy {
...
-                    .to(binding.to)
-                    .service(&self.bus_addr)
-                    .call(msg)
-                    .await
-            }
-        };
+        let endpoint = self.endpoint();
 
-        let result = response.unwrap_or_else(|e| Err(HttpProxyStatusError::from(e)));
+        log::info!("Proxy http {method} `{path}` call to [{}]", endpoint.addr());
```
prawilny, line 94: micronit: wouldn't it be cleaner to implement `fmt::Display` for `GsbHttpCallMessage`?

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
```diff
@@ -74,44 +83,52 @@ impl HttpToGsbProxy {
...
-                body: actix_web::web::Bytes::from(format!("Error: {}", err))
-                    .try_into_bytes()
-                    .map_err(|_| {
-                        Error::GsbFailure("Failed to invoke GsbHttpProxy call".to_string())
-                    }),
-                status_code: StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
-                response_headers: HashMap::new(),
-            },
+            Ok(r) => {
+                log::debug!(
```
prawilny, line 102: Is `debug` log level enabled by default?
Do we want to collect these logs by default? On requestor side? On provider side?

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
```diff
@@ -74,44 +83,52 @@ impl HttpToGsbProxy {
...
+                    endpoint.addr()
+                );
+                HttpToGsbProxyResponse {
+                    body: actix_web::web::Bytes::from(format!("Error: {}", err))
+                        .try_into_bytes()
+                        .map_err(|_| {
+                            Error::GsbFailure("Failed to invoke GsbHttpProxy call".to_string())
+                        }),
+                    status_code: StatusCode::INTERNAL_SERVER_ERROR.as_u16(),
+                    response_headers: HashMap::new(),
```
prawilny, line 129: Recently Marcin mentioned his worries about not knowing what part of the data path 500 came from.
Maybe we could consider adding some `x-...` header pointing to the yagna proxy's fault?

nieznanysprawiciel: It was considered, but I didn't treat it as part of this task: https://github.com/golemfactory/ya-runtime-ai/issues/109

core/activity/src/http_proxy.rs
nieznanysprawiciel: This is a bug, because wrong identity is checked: https://github.com/golemfactory/ya-runtime-ai/issues/105

prawilny: Then IMO we should use merge without squash so that the commit message for the fix stays separate.

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
prawilny: I think you made the correct choice and it was me who was overeager.

core/activity/src/http_proxy.rs
nieznanysprawiciel: I always use merge for my PRs

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
nieznanysprawiciel: I added this in last commit. It is not so elegant, because this Display doesn't print all fields but it's ok

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
nieznanysprawiciel: They are not. I changed to info to simplify, but probably this log should be on debug when we will merge this later to master

exe-unit/components/gsb-http-proxy/src/http_to_gsb.rs
prawilny: I always thought that `Debug` is for all fields :smiley:

nieznanysprawiciel: I'm not using this Display impl everywhere, since I don't want to clone body

exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
```diff
@@ -53,9 +53,9 @@ impl GsbToHttpProxy {
 
     pub fn bind(&mut self, gsb_path: &str) -> Handle {
         let this = self.clone();
-        bus::bind(gsb_path, move |message: GsbHttpCallMessage| {
+        bus::bind_with_caller(gsb_path, move |caller, message: GsbHttpCallMessage| {
```
prawilny, line 56: Do I understand correctly that it aims to resolve https://github.com/golemfactory/ya-runtime-ai/issues/105?

If so, how does it solve the problem if we don't make use of that parameter in `pass()` apart from logging? Or do we only want to log the messages for now and fix it after collecting more data?

prawilny: Or do we solve the problem by switching `authorize_activity_executor()` to `authorize_activity_initiator()` and log to ensure that there are no problems with it after the change?

prawilny (approved):

exe-unit/components/gsb-http-proxy/src/gsb_to_http.rs
nieznanysprawiciel: Yes it's solves the problem with spam that reports incorrectly that something is wrong when it isn't.
I don't want to block possibility of using other yagna as proxy, so the call is not rejected even if it is from incorrect identity

Issue #3402, opened 2025-06-25
Workaround GolemBase maximal query limits


nieznanysprawiciel:
There is limit of 5Mb or query.
On the market startup it queries all existing Offers.
This situation must be handled when the marketplace grows.

PR #3494, opened 2025-10-16
Fix GolemDb query limit on yagna start

nieznanysprawiciel:
resolves: https://github.com/golemfactory/yagna/issues/3402


nieznanysprawiciel:
Tests are failing due to `Faucet donation not received`. We need to switch tests to hoodie.
Merge with master will solve this. It will be done in next PR.

